PR #77, opened 2026-03-09
Kimi K2.5 NVFP4 SGLang MCR Sweep

Slonegg:


Slonegg:
/run-experiment experiments/Kimi-K2.5-NVFP4/optimal_mcr_b200_sglang

PR #90, opened 2026-04-08
Reuse server for benchmark

trifonova-n:
Added an option to pass on docker flags
Added an option to reuse existing server for benchmarks.
Model and docker container loaded on previous experiments can be reused


Slonegg (commented):
I suggest to ask AI to fix merge conflicts by dropping reuse changes and keeping other changes.

PR #167, opened 2026-05-29
Restore Tile-IR / Kernel-IR layering for warp specialization

Slonegg:
## Context

PR #166 (warp specialization for TMA matmul) shipped working WS kernels but punched through the Tile-IR / Kernel-IR layering. `085_warp_specialize.py` imported `MbarrierArrive / MbarrierInit / MbarrierWait / SetMaxNReg / Smem / Sync` from `ir.kernel.ir` and spliced them into the Tile body itself; `AsyncWait` gained `barrier_id` / `barrier_count` fields that only made sense as named-`bar.sync` hardware identifiers.

The companion pass `080_pipeline_stages` (clean) declares pipelining intent with one Tile-IR Stmt (`AsyncWait`) and lets `100_materialize_tile`'s `emit_async_wait` translate to `CpAsyncWait+Sync` or `MbarrierWait+Sync` based on `StagePolicy`. WS should follow the same pattern. This PR makes it do so.

After this PR `lowering/tile/*.py` does not import from `ir.kernel.ir` — locked in by a CI guard.

## Approach

Two new artifacts, mirroring the 080 model:

1. **`WarpSpecialize` Tile-IR Stmt** (`ir/tile/ir.py`) — block Stmt with `producer_body` / `consumer_body` / `ring_depth` / `n_producer_threads`. Predicate, consumer thread count, slot/phase exprs, and mbar name are derivable at materialize time from the enclosing `ThreadTile.axes`. The K_o axis is discovered structurally (the single `SerialTile(serial_outer)` in each branch) rather than carried by name — survives `normalize_body`'s canonical Axis rename. `__post_init__` rejects nested `WarpSpecialize`.

2. **`emit_warp_specialize`** closure in `100_materialize_tile.py` — expands the marker into the empty-mbarrier ring (Smem + per-slot MbarrierInit gated on tid==0 + Sync), per-K_o handshake (MbarrierWait gated by `K_o >= bc-1` in the producer; named `bar.sync 1, n_cons` + single-thread MbarrierArrive appended in the consumer), and the producer/consumer `Cond` wrapper framed by `SetMaxNReg(24,\"dec\")` / `SetMaxNReg(240,\"inc\")`. `emit_async_wait` gains an optional `ws_consumer` kwarg; the consumer subtree walk threads it via a per-call `ews` override on `_process_stmt` so AsyncWaits inside route their trailing fence to the named bar.sync.

Pass + cleanup:

3. **`085_warp_specialize`** now packages the σ-shifted role split into a `WarpSpecialize` and stops. The hand-rolled `_wire_producer_wait` / `_wire_consumer_arrive` / `_stamp_async_wait_barrier` helpers and the inline Kernel-IR primitives are gone (pass shrinks ~430 → ~290 lines).

4. **`AsyncWait.barrier_id` / `barrier_count`** deleted — the materializer derives named-barrier params from the enclosing `WarpSpecialize` context instead.

5. **`tests/architecture/test_layering.py`** — static-import guard that scans `lowering/tile/*.py` for `from deplodock.compiler.ir.kernel` and fails CI if any reappear.

## Commit-by-commit

Each commit ends with `make test` green.

- **C1** `f48a9eb6` — Add `WarpSpecialize` Stmt + 16 protocol tests. Pure addition; nothing emits it yet.
- **C2** `62b253a2` — Add `emit_warp_specialize` + 6 materializer integration tests (build `WarpSpecialize` directly, materialize, check the lowered shape).
- **C3** `9ce8a083` — **Boundary restored.** `085_warp_specialize` emits `WarpSpecialize`; drops the Kernel-IR import and the wiring/stamping helpers. Adds 2 boundary checks to the existing pass tests.
- **C4** `f394e3c6` — Delete `AsyncWait.barrier_id` / `barrier_count` (the legacy C3-compat fields).
- **C5** `f268539e` — Add the layering CI guard.

## Verification

- `make test` — 1364 passed (+7 new), 27 skipped, 0 failed
- `make lint` — clean
- `grep -rn \"from deplodock.compiler.ir.kernel\" deplodock/compiler/pipeline/passes/lowering/tile/` — empty
- Unit-level coverage of the materializer's emitted shape: empty-mbar prologue, role-split Cond with SetMaxNReg framing, per-K_o handshake, named-bar trailing fence vs default `__syncthreads()` outside WS.

### What I didn't verify locally

End-to-end CUDA bit-equivalence vs `main` for a TMA-eligible matmul on sm_90 — I don't have a Hopper box in this session. Suggested check before merge:

\`\`\`
git checkout main
DEPLODOCK_WS=1 DEPLODOCK_DUMP_DIR=/tmp/ws_before deplodock compile --code \\
    \"import torch; torch.matmul(torch.randn(512, 512, dtype=torch.float16, device='cuda'), \\
     torch.randn(512, 512, dtype=torch.float16, device='cuda'))\" --target sm_90
git checkout feature/warp-specialize-tile-stmt
DEPLODOCK_WS=1 DEPLODOCK_DUMP_DIR=/tmp/ws_after deplodock compile --code \"...\" --target sm_90
diff -ur /tmp/ws_before /tmp/ws_after
\`\`\`

Expected: empty diff (modulo identifier shuffling if any). Then `DEPLODOCK_WS=1 deplodock run --code \"...\" --bench` for accuracy + latency parity.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,4 @@ recipes/*/[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]_*/
 # Claude Code per-user local settings (skills/agents under .claude/ are tracked)
 .claude/settings.local.json
 *-dump/
+.claude/scheduled_tasks.lock
--- a/deplodock/compiler/ir/tile/ir.py
+++ b/deplodock/compiler/ir/tile/ir.py
@@ -158,6 +158,100 @@ class AsyncWait(Stmt):
 
 
 # ---------------------------------------------------------------------------
+# WarpSpecialize — producer/consumer split for TMA-pipelined kernels
+# ---------------------------------------------------------------------------
+#
+# Tile-IR marker that the materializer (``100_materialize_tile``) lowers
+# into the full mbarrier handshake: empty-mbarrier ring (``Smem`` +
+# per-slot ``MbarrierInit``), per-K_o ``MbarrierWait`` / ``MbarrierArrive``
+# pairs, named ``bar.sync`` consumer fences, ``SetMaxNReg`` register
+# budget redistribution, and the producer/consumer ``Cond`` wrapper.
+#
+# The pass that emits this (``085_warp_specialize``) keeps all Tile-IR
+# vocabulary — no ``from deplodock.compiler.ir.kernel.ir import …``.
+# Companion to 080's ``AsyncWait``: where ``AsyncWait`` lets the pass
+# declare "wait for an async chunk, materializer picks the primitive",
+# ``WarpSpecialize`` lets the pass declare "split this ThreadTile body
+# into producer and consumer roles, materializer wires the rest".
+
+
+@dataclass(frozen=True)
+class WarpSpecialize(Stmt):
+    """Producer/consumer warp split inside a TMA-pipelined ThreadTile.
+
+    Fields:
+
+    - ``producer_body`` — stmts run by producer threads (TMA-issue
+      ``StageBundle`` scaffolding inside ``SerialTile(serial_outer)``).
+    - ``consumer_body`` — stmts run by consumer threads (``AsyncWait`` +
+      reduce loop + output ``Write``). Already σ-shifted by the pass so
+      every reference to the extended thread axis sees the original
+      ``[0, n_consumer_threads)`` range.
+    - ``ring_depth`` — empty-mbarrier slot count (== TMA buffer_count).
+    - ``n_producer_threads`` — for SetMaxNReg accounting (producers get
+      a small budget so consumers can claim the rest) and for deriving
+      the producer/consumer ``Cond`` predicate from the enclosing
+      ThreadTile.
+
+    The K_o axis the WS pass aligned scheduling against is identified by
+    the materializer structurally — the (single) ``SerialTile(serial_outer)``
+    in each branch. Carrying the axis *name* would break under
+    ``normalize_body``'s canonical rename pass (which renames Axis but
+    can't see a plain string field).
+
+    Other quantities (``predicate``, ``n_consumer_threads``, slot/phase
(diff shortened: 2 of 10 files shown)


Slonegg:
End-to-end validated on RTX 5090 (sm_120, locally).

**Found and fixed one bug** (commit 690cbe72): `partition_tma_groups` in `_tma_groups.py` didn't descend into `WarpSpecialize.producer_body` / `consumer_body`, so the TMA stage's smem source never got registered → `KeyError: 'x0_smem'` in `emit_tma_stage`. Added a `WarpSpecialize` case mirroring the existing `Cond` case. The unit tests didn't catch it because they exercised `emit_warp_specialize` directly; the bug only surfaced when going through the full TileOp → KernelOp pipeline.

**Bit-equivalent vs main on a 128×128 fp16 matmul with `DEPLODOCK_WS=1`:**

| | main | feature/warp-specialize-tile-stmt |
|---|---|---|
| Kernel name | `k_matmul_6c2220` | `k_matmul_6c2220` |
| Accuracy vs eager | PASS (max_diff=0.031) | PASS (max_diff=0.031) |
| Latency | 8.2 us | 8.2 us |
| Grid / block | 256 / 128 | 256 / 128 |
| Smem / regs / occ | 8.1K / 42 / 100% | 8.1K / 42 / 100% |

**CUDA source diff** (512×512 fp16 matmul, `--target sm_120` via live device) — semantically equivalent:
- `if (0)` (main, constant-folded) vs `if (a6 >= 1)` (feature) for the producer-wait predicate — same behavior on a single-iter main loop
- Accumulator `acc0..acc15` declared at function scope vs inside the role-split `Cond` branches — both correct, different placement
- No other differences in `__shared__` allocation, mbarrier init / wait / arrive, `setmaxnreg`, or `bar.sync` emission

`make test` — 1364 passed (+8 over baseline), 27 skipped.

PR #205, opened 2026-06-05
Learned online prior for the inner tune MCTS

Slonegg:
## Summary

Replaces the hand-written `TileOp.score` tiebreaker with a **learned, in-memory Bayesian-ridge prior** fit from each
`tune` run's own benches, and nukes the static-prior + DB-greedy selection logic so the learned prior is tested in
isolation. Design + findings: `plans/learned-prior-puct.md`.

### M1 — nuke static + DB-greedy selection
- `mcts._ucb_key` no longer calls `score_of`; the unvisited-sibling tiebreak is the learned prior, or `0` (pure UCB
  emission order) when none is attached.
- `greedy.push` keeps option-0; `_best_fork` / `_knob_eq` / the `best=` push arg are removed.
- The planner score **compute** (`Fork.score` / `lazy_score` / `score_of`) stays as latent infra (still exercised by the
  partition-planner tests) — only its use in *selection* was removed.
- Deleted `test_greedy_db_lookup.py` (tested removed behavior); `xfail`'d the matmul golden-picker test that relied on
  greedy's prior pick (restore when compile/run reconnect to the prior).

### M2 — `OnlinePrior` (`search/prior.py`)
- **Value-of-position labels** — real benches exist only at leaves, but the prior ranks partial-knob siblings at every
  fork level, so each node is labeled `log(best_reward)` (max over its benched descendants, already maintained by the
  tree). Leaves *and* branches are training rows.
- **Refit from a tree snapshot** (labels are non-stationary), **Thompson-sampled** per descent. `knob.knob_features` is
  the featurizer.
- `tune --prior {off,online,shadow}` — `shadow` fits + reports the same stats but scores 0 (pure-UCB baseline for the
  A/B). End-of-run sanity block: best-found, silly-pick warmup/post, self-calibration. 11 new unit tests.

### Incidental fix
Benching was fully broken on the branch (`ModuleNotFoundError: '010_partition_loops'`): the plan-memo `_Plan` rode the
pickled `Op.source` chain into the bench worker. `Op.__getstate__` now drops `meta` (documented "never serialized").

## Probe result
Large-K linear reduce (M=8, K=N=4096), same seed/patience:

| arm | best found |
|---|---|
| shadow (pure UCB) | 205.9 µs |
| online (learned prior) | **41.0 µs** |

The learned prior found a **5× better** kernel; calibration Spearman **+0.6–0.9** (it learns reduction-knob quality
within one run). Not a fixed-budget comparison yet — a `max_visits`-capped A/B is the honest next measurement.

## Heads-up
The `_best_fork` nuke **severs the tune→compile handoff**: `compile`/`run` now emit rule-default forks regardless of the
DB (CLAUDE.md / ARCHITECTURE.md / docstrings updated to say so). Reconnecting greedy to the learned prior is the planned
follow-up.

## Test plan
- `make test`: 1645 passed, 1 xfail (intended), 29 skipped. (One `test_inner_reward_is_separable` failure under parallel
  GPU contention — passes standalone.)
- `make lint`: clean.


deplodock/compiler/ir/features.py
```diff
@@ -0,0 +1,119 @@
+"""Structural feature extraction — describe a kernel's loop-level computation
```
Slonegg: I think these features are only relevant for LoopIr, so let's move them directly to the pass.

deplodock/compiler/pipeline/knob.py
```diff
@@ -292,3 +297,86 @@ def format_tuning_knobs(knobs: dict) -> str:
...
+            pop = float(s.count("1"))
+            feats[f"{name}_popcount"] = pop
+            feats[f"{name}_width"] = float(len(s))
+            feats[f"{name}_frac"] = pop / len(s) if s else 0.0
+        # STR knobs other than MMA: no generic numeric encoding.
+    feats.setdefault("MMA_tier", 0.0)  # scalar tier = no atom selected
+    return feats
+
+
+def _mma_features(mma: str) -> dict[str, float]:
```
Slonegg: Let's declare the feature function at knob declaration, so we don't need to do a custom lookup.

deplodock/compiler/pipeline/search/policy/mcts.py
```diff
@@ -41,12 +48,9 @@ class SearchNode:
     visits: int = 0
     best_reward: float = 0.0  # max reward over this subtree's measured leaves
     live: int = 0  # count of un-popped frontier leaves in this subtree
-    # Memoized ``Search.score_of`` read (filled by ``_ucb_key``) — UCB
-    # descent reads the prior on every unvisited sibling at every level
-    # per pop, and a branch fork's prior is a max over its param
-    # subgroup, so the per-node memo keeps repeated descents O(1). Sound
-    # because the fork's score is frozen once the SearchNode is attached.
-    prior: float | None = field(default=None, repr=False)
+    # No prior memo: the learned ``OnlinePrior`` is Thompson-sampled (a fresh
```
Slonegg: Drop this comment

deplodock/compiler/pipeline/search/prior.py
```diff
@@ -0,0 +1,216 @@
...
+from __future__ import annotations
+
+import math
+
+import numpy as np
+
+from deplodock.compiler.pipeline import knob
+
+
+class OnlinePrior:
```
Slonegg: Let's create a prior/base.py module with a prior base class, and a module with this specific prior - give it a more descriptive name.

deplodock/compiler/pipeline/search/two_level.py
```diff
@@ -246,6 +270,8 @@ def run_two_level_tune(
     ucb_c: float = TuningSearch.DEFAULT_UCB_C,
     dump=None,
     progress=None,
+    prior: str = "off",
```
Slonegg: Can these be incorporated into the prior constructor? Can we pass the prior object here?

deplodock/compiler/pipeline/passes/lowering/tile/010_partition_loops.py
```diff
@@ -614,11 +614,11 @@ def _plan_kernel(loop_op: LoopOp, ctx: Context, *, kernel_name: str = "", graph:
     only written after a successful non-empty enumeration, so skipped /
     rejected shapes never poison it.
     """
-    if "SID" not in loop_op.knobs:
+    if not any(k.startswith(STRUCT_PREFIX) for k in loop_op.knobs):
         # Normally stamped by ``loop/fusion/040_stamp_structural_id``;
         # direct callers (tests, ad-hoc planning) get the same identity
         # computed here.
-        loop_op.knobs["SID"] = loop_structural_id(loop_op, graph)
+        loop_op.knobs.update(structure_features(loop_op.body, graph))
```
Slonegg: We should get rid of this hack and only stamp them in 040_stamp_structural_id

deplodock/compiler/pipeline/passes/loop/fusion/040_stamp_structural_id.py
```diff
@@ -1,41 +1,42 @@
-"""Stamp ``LoopOp.knobs["SID"]`` with the kernel's structural identity.
+"""Stamp ``LoopOp.knobs`` with the kernel's structural features.
```
Slonegg: Let's rename the file to 992_stamp_structural_features and rename 030_stamp_loop_names.py -> 991_stamp_loop_names.py

deplodock/compiler/ir/base.py
```diff
@@ -73,6 +73,22 @@ class Op:
...
+    def __getstate__(self) -> dict:
+        """Pickle without ``meta`` — it is per-process scratch (the partition
+        planner stashes its ``_Plan`` there, whose class is loaded under a bare
+        rule-module stem and is unimportable in a subprocess). The bench worker
+        receives the ``CudaOp`` graph by pickle and its ``Op.source`` provenance
+        chain reaches a ``LoopOp`` carrying that ``_Plan``, so without this the
+        unpickle dies with ``No module named '010_partition_loops'``. ``meta``
+        is documented "never serialized"; this enforces it."""
+        state = self.__dict__.copy()
+        state["meta"] = {}
```
Slonegg: I think we can drop meta - I believe it is no longer used. Everything is in knobs.

tests/compiler/test_matmul_mma_tma.py
```diff
@@ -150,6 +150,11 @@ def _compile_and_render(*, M: int, N: int, K: int, out_dtype: DataType):
 
 
 @pytest.mark.skipif(not _supports_tma(), reason="TMA needs sm_90+ (Hopper / Blackwell)")
+@pytest.mark.xfail(
+    reason="greedy's prior-based pick was nuked (learned-prior work, plans/learned-prior-puct.md): greedy now emits "
+    "option-0, not the score-max golden. Restore when compile/run reconnect to the learned prior.",
+    strict=False,
+)
 def test_default_picker_lands_on_tma_golden_at_2048_fp16(monkeypatch):
```
Slonegg: Drop this test

tests/compiler/passes/test_structural_features.py
```diff
@@ -0,0 +1,159 @@
...
+from deplodock.compiler.ir.elementwise import ElementwiseImpl
+from deplodock.compiler.ir.expr import Var
+from deplodock.compiler.ir.features import STRUCT_PREFIX, structure_features
+from deplodock.compiler.ir.stmt.blocks import Loop
+from deplodock.compiler.ir.stmt.body import Body
+from deplodock.compiler.ir.stmt.leaves import Accum, Assign, Load, Write
+from deplodock.compiler.tensor import Tensor
+
+
+def _rms_body(ext_i: int = 8, ext_k: int = 64) -> Body:
```
Slonegg, line 33: Add some more complex tests, maybe from our perf set, like triple-matmul and other less-common shapes.

Slonegg:
Addressed the review in `098a9e48` (+ depth-2 PUCT in `9f24d86d`). Mapping:

- **`features.py` → move to the pass / stamp only in 040** — `structure_features` now lives in
  `992_stamp_structural_features` (renamed from `040_stamp_structural_id`; `030_stamp_loop_names` → `991`). Deleted
  `ir/features.py`; `STRUCT_PREFIX` moved to `pipeline/knob.py` (importable home). Removed the `_plan_kernel` fallback —
  the pass is the sole producer; ad-hoc callers stamp explicitly.
- **Feature fn at knob declaration** — added `Knob.features`; `knob_features` dispatches through it (no MMA special-case).
  `_mma_features` sits next to the `MMA` knob in `_enumeration.py`.
- **Drop the mcts comment** — done.
- **prior/base.py + descriptive name** — `prior.py` → `prior/` package: `Prior` base (interface + end-of-run stats) in
  `base.py`, `BayesianRidgePrior` (was `OnlinePrior`) in `bayesian_ridge.py`.
- **Pass the prior object** — `inner_reward` now takes a `prior_factory` (no-arg callable → fresh prior per kernel);
  mode flags (`active`/`acquisition`) ride the prior object, so `TuningSearch` reads `acquisition` off it.
- **Drop `meta`** — dropped `Op.meta` + my `__getstate__` hack. ⚠️ It *was* used — by the partition-planner plan memo
  (`meta["plan"]` / `_plan_cache_key`), so I removed that too (and its 3 tests). This root-fixes the bench-worker pickle
  crash, but it removes the "structural memo" compile-time optimization from the compile-perf work — correctness is
  unaffected; flagging in case the re-enumeration cost matters.
- **Drop the golden-picker test** — deleted.
- **More feature tests** — moved to `tests/compiler/passes/test_structural_features.py` + added triple-matmul, uncommon
  (non-pow2) shapes, and matmul-vs-reduction discrimination.

One nuance from the knob change: MMA expansion now requires the `MMA` knob to be *registered* (its module loaded),
which is always true in the pipeline but not in a bare unit test — updated `test_knob_features_mma_expansion` to load it.

`make lint` clean; 823 passed / 2 skipped across `tests/compiler/{passes,pipeline,ir}` (incl. the real-CUDA tune test).

Also FYI: depth-2 `--prior puct` (prior in the UCB arithmetic) landed too — on RMSNorm(8192) it robustly finds the true
best at bench #35–41 with ~33% silly picks vs pure UCB's #72 / 66%, across 3 seeds.

deplodock/commands/tune.py
```diff
@@ -181,7 +181,15 @@ def handle_tune(args):
     t0 = time.monotonic()
     try:
         result = run_two_level_tune(
```
Slonegg, line 220: Maybe we should construct prior and tuning search and pass constructed objects to run_two_level_tune

scripts/train_prior.py
```diff
@@ -0,0 +1,43 @@
+"""Offline-fit the global learned prior from accumulated tune datasets.
```
Slonegg: Delete this script, now we have deplodock tune for it

scripts/prior_data.py
```diff
@@ -0,0 +1,55 @@
+"""Shared dataset extraction for offline learned-prior training / evaluation.
```
Slonegg: Delete this script as we're deleting train_prior.py

scripts/prior_bakeoff.py
```diff
@@ -0,0 +1,218 @@
+"""Offline model bake-off for the learned tuning prior.
```
Slonegg: Delete this script

scripts/gen_prior_data.sh
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
```
Slonegg: Delete this script

deplodock/compiler/pipeline/search/golden.py
```diff
@@ -33,6 +33,8 @@
```
Slonegg, line 39: Let's rename to just golden.py

deplodock/compiler/pipeline/search/policy/greedy.py
```diff
@@ -7,30 +22,73 @@
...
-    sibling via :meth:`Search.score_of` (ties keep emission order, so
-    rules whose options carry no scores still get their option-0
-    default)."""
+    """Keep one pending candidate; pick it by the learned prior (argmax) or, with
+    no trained prior, by emission order (option-0)."""
 
     def __init__(self) -> None:
         super().__init__()
         self._pending: LazyCandidate | None = None
+        self._prior = None  # lazily loaded on first push (the regime's checkpoint)
```
Slonegg, line 62: Maybe we should just let the caller configure it? Either pass prior or use a hardcoded heuristic? Maybe derive a hardcoded heuristic from Prior.

deplodock/compiler/pipeline/search/policy/mcts.py
```diff
@@ -17,21 +21,33 @@
...
 
 from deplodock.compiler.pipeline.search.candidate import LazyCandidate
 from deplodock.compiler.pipeline.search.db import PerfStats
 from deplodock.compiler.pipeline.search.policy.base import Search
 
+if TYPE_CHECKING:
+    from deplodock.compiler.pipeline.search.prior import Prior
+
+
+def _softmax(xs: list[float]) -> list[float]:
```
Slonegg: Is CatBoost not doing this by default?

PR #222, opened 2026-06-11
tune/compile: first-class structural forks in the two-level search (plan steps 1-3)

Slonegg:
Executes milestones M1–M3 of `plans/structural-forks-in-two-level.md` — kernel-set-changing lowering decisions
(the `SPLIT_CONE` demoted-matmul split) become first-class in the two-level search, and greedy `compile` / `run`
learn to deploy them unpinned. M4 (the `SPLITK` hoist for 017 / Stream-K) remains; its open design questions are
recorded in the plan.

## M1 — classify forks by effect at the spawn site
Any `Graph`-splicing option marks the fork `structural=True` at `Run.drive`'s spawn site (where the raw option
list is concrete, no thunk fired); the flag rides `Search.push(structural=)`. `Op` rebinds and the partition
planner's branch Forks read op-variant. Inert by itself.

## M2 — boundary redrawn on effect, not pass index
The outer search drives through the pre-partition head of `lowering/tile` (`two_level.outer_pipeline()`), so
005's keep-vs-split offer branches the **outer** tree: one terminal per kernel set, compared by Σ-per-op cost,
with split producers/consumers tuned in first-class per-op slices (own patience, own progress leaf,
`op_cache_key` dedup) instead of cross-producting inside the fused kernel's slice. Structurally identical offer
sites within a trajectory replay the same side via a per-`(rule, op_cache_key)` decision memo
(`Candidate.structural_memo`), keeping the outer tree linear in unique kernels instead of `2^sites`. 017's
sub-partition splice stays on the tolerated slice-sum path until M4.

**Verified on GPU** (RTX 5090, norm→linear 32×1024×3072, isolated DB/prior): the tune reports 2 outer terminals —
fused Σ 238.8 µs vs split Σ 20.3 µs — with the `_xn` producer as its own tuned op leaf.

## M3 — greedy pricing for structural options (the deploy path)
With the *trained* prior loaded, `GreedySearch._pick_structural` prices both sides — per kernel, a nested greedy
descent over the single-node slice (`lowering/tile` only, CPU, no backend) records the prior's predicted µs for
the chosen tile at the partition fork, memoized per `op_cache_key`; the structural side is the Σ over its
fragment's kernels — and the cheaper kernel set wins. Cold (analytic / no prior) or unpriceable options keep the
old filter, so a cold compile never changes kernel sets. A structural pick whose fragment kernel fails to lower
retires structural picks on `Pipeline.run`'s retry and re-drives down the keep-fused branch. This also re-aligns
`run_two_level_tune`'s final assembly with its reported Σ.

**Verified on GPU**: after the tune above, unpinned `run --code` deploys the two-kernel split.

## Also
- `tests/compiler/pipeline/search/test_two_level.py` now isolates its prior checkpoint to `tmp_path` — the
  fake-backend tune tests were training and checkpointing into the host's real `~/.cache/deplodock/prior.json`.
- Plan + `pipeline/ARCHITECTURE.md` + CLAUDE.md updated.

## Notes for reviewers
- `test_inner_reward_is_separable_not_a_product` is flaky on at least one host **independent of this PR**
  (reproduced at main-equivalent code): the same tune trajectory benches the same variant label at different
  fake latencies across processes even under a fixed `PYTHONHASHSEED`, i.e. some per-process nondeterminism in
  tune-mode exploration. Worth a separate investigation — it undermines tune reproducibility generally.
- `make test` (1823 passed) + `make lint` green at every milestone commit.


Slonegg:
Pushed a follow-up commit (M3b) from the design discussion: composed kernel-set Σ training rows + an informed outer PUCT.

- Each outer terminal emits one value-of-position row per structural decision — features `{ctx, pre-decision op knobs, decision delta}`, label = that side's per-kernel-best Σ. Attribution reuses the `Op.source` chain (stamped across loop-dialect lowering splices by `Candidate.apply`; 005 sets it explicitly on the keep rebind; `_rename_buf_in_op` now preserves `source`) rather than new search-engine state.
- The outer `TuningSearch` now carries the global prior (its PUCT was uniform before — `prior_model` was never passed), and `LazyCandidate.resolved_knobs` fixes a latent `_node_knobs` blind spot: a resolved ancestor's fork delta used to vanish from descendants' feature vectors, so the structural branch's continuation was scored as a knob-less row against the fully-knobed unresolved sibling.
- Decision hops (loop→loop source hops) are excluded from the `lowering` table — its one-best-child upsert would let a multi-kernel decomposition's parent key resolve through a single fragment kernel's median.
- Deploy decisions unchanged: composed rows order tune-time exploration; greedy keeps the compositional probe.

GPU-verified from a cold start: run 1 explores fused-first (emission order), run 2 descends the split kernel set first.

PR #250, opened 2026-06-17
serving: on-device torch I/O (dlpack zero-copy) + opt-in static batched mode

Slonegg:
Two related serving-perf changes (stacked commits on one branch). Both are correct + tested; the headline finding is that the **per-request latency bottleneck is kernel compute, not host I/O or batching**.

---

## 1. dlpack zero-copy I/O (`ebad93f8`)

The plugin bridged vLLM's torch CUDA tensors to the compiled program's `cupy` buffers **through host numpy** (`.cpu().numpy()` in, `.get()` + `from_numpy().to(device)` out, plus a host-built causal mask per request) — two full GPU→host→GPU round-trips every request.

Now bridged **on-device**: `runner.forward_hidden_states` takes/returns torch CUDA tensors, bridging via `cp.from_dlpack` / `torch.from_dlpack` on torch's stream (`cp.cuda.Stream.from_external`); new `CompiledProgram.upload_prefix_device` / `output_prefix_device` are device-to-device twins of the host methods; the causal mask is built once per S on-device. (Serving ARCHITECTURE follow-ups #2 + the host half of #3.)

**Perf — concurrency 1:** no change (64.0 vs 63.3 req/s; 15.45 vs 15.67 ms mean). Host round-trips cost *overlap under concurrency*, not single-request latency — at c1 the GPU forward is on the critical path regardless. Lands because it's correct, removes real CPU↔GPU syncs, and is the prerequisite for batching.

## 2. Opt-in static batched mode (`d57d95c4`)

`DEPLODOCK_SERVING_STATIC=1` builds ONE fully-static `(max_num_seqs, max_seq_len)` program (static extents for **both** batch and seq) and runs each scheduler step as a padded batched forward. Batch size = vLLM's own `--max-num-seqs` (read at init), not a new knob. Off by default.

Static-only because the dynamic-seq (masked-tile) kernels miscompute batch>1 (verified empirically); a static trace bakes correct batch strides. Padding is safe under causal masking.

**Perf — concurrency 32, uniform 512 tokens (RTX 5090, Qwen3-Embedding-0.6B):**

| config | conc. | req/s | per-forward |
|---|---|---:|---|
| deplodock batch-1 (dlpack) | 1 | 64 | 1 seq / ~15 ms |
| **deplodock static batch-32** | 32 | **26.6** | 32 seq / ~1.1 s |
| stock vLLM (warm) | 32 | ~232 | 32 seq / ~0.14 s |

**This is a throughput *regression*, not a win** — batching does B× the existing non-flash `O(B·H·S²)`-materialized attention, and the static-shape kernels are cold greedy picks (the batch-1 symbolic kernels carried prior tuning). Committed off-by-default as correct, tested scaffolding + the recorded negative result. The real lever is efficient attention (flash/varlen, follow-up #1) and/or an autotune of the static batched shape.

---

## Validation

- `tests/serving/test_vllm_plugin_gpu.py` (real in-process vLLM, cosine vs HF eager) passes through the new torch path.
- `tests/compiler/ir/test_dynamic_shapes.py::test_capture_replay_device_io_matches_eager` — the dlpack device-I/O primitives vs eager.
- `tests/serving/test_runner_batched_gpu.py` (perf) — batched forward of several different-length sequences matches eager per row (causal-independence-under-padding).
- Compiler regression suite green (1587 passed, 4 skipped); lint clean.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -26,7 +26,7 @@ When the user asks about a CLI flag, recipe field, or matrix combinator, read th
 
 All `DEPLODOCK_*` config env vars (the two above plus `DEPLODOCK_NVCC_FLAGS`, `DEPLODOCK_DEBUG`, `DEPLODOCK_KNOBS`,
 `DEPLODOCK_TUNE_PATIENCE`, `DEPLODOCK_TUNE_EPS`, `DEPLODOCK_O3_TOL`, `DEPLODOCK_ANALYTIC_TILT`, `DEPLODOCK_BENCH_BACKENDS`,
-`DEPLODOCK_CUBIN_CACHE`, `DEPLODOCK_NO_NVCC`, `DEPLODOCK_GPU_LOCK`,
+`DEPLODOCK_CUBIN_CACHE`, `DEPLODOCK_NO_NVCC`, `DEPLODOCK_GPU_LOCK`, `DEPLODOCK_SERVING_STATIC` (static batched serving),
 …) are read and written through a single module, `deplodock/config.py` — the sole owner of `os.environ` for these vars.
 CLI `--flag` overrides (e.g. `--nvcc-flags`) resolve via `config.set_nvcc_flags` inside the library, not in the command
 layer, so programmatic callers and tests get the same precedence. The dynamic `DEPLODOCK_<KNOB>` namespace is owned by
--- a/deplodock/compiler/backend/cuda/program.py
+++ b/deplodock/compiler/backend/cuda/program.py
@@ -939,6 +939,26 @@ class CompiledProgram:
                 raise ValueError(f"upload_prefix: {name!r} has {flat.size} elems > capacity {arr.size}")
             arr.ravel()[: flat.size].set(flat)
 
+    def upload_prefix_device(self, input_data: dict[str, cp.ndarray]) -> None:
+        """Device-to-device twin of :meth:`upload_prefix`: copy each cupy source
+        into the contiguous PREFIX of its capacity-sized device buffer with NO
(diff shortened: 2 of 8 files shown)


Slonegg:
## Follow-up: opt-in static batched mode + batched benchmark (commit d57d95c4)

Added `DEPLODOCK_SERVING_STATIC=1` — a fully-static `(max_num_seqs, max_seq_len)` program (static extents for **both** batch and seq) that runs each scheduler step as one padded batched forward. Batch size comes from vLLM's own `--max-num-seqs` (read at init), not a new knob. Correct (new perf test passes; causal masking makes right-padding safe), **off by default**.

### Benchmark — RTX 5090, Qwen3-Embedding-0.6B, uniform 512 tokens, concurrency 32

| config | concurrency | req/s | per-forward |
|---|---|---:|---|
| deplodock batch-1 (dlpack) | 1 | 64 | 1 seq / ~15 ms |
| **deplodock static batch-32** | 32 | **26.6** | 32 seq / ~1.1 s |
| stock vLLM (warm) | 32 | ~232 | 32 seq / ~0.14 s |

**Honest result: batching the current kernels is a throughput *regression*, not a win.** The static `(32,512)` forward takes ~1.1 s (~8× stock). Making the extents static is not the lever — the batched program does B× the existing non-flash `O(B·H·S²)`-materialized attention, and the static-shape kernels are cold greedy picks (the batch-1 symbolic kernels carried prior tuning). The real throughput fix needs **efficient attention (flash/varlen)** and/or an **autotune of the static batched shape**.

Committed off-by-default as correct, tested scaffolding plus the recorded negative result, to set up that follow-up work.

PR #259, opened 2026-06-19
W4A16 (AWQ-INT4) quantization: in-kernel int4 dequant

6erun:
## Summary

Adds **W4A16 / AWQ-INT4** support: compressed-tensors quantized models now trace, compile, and run on the CUDA backend
with **in-kernel weight dequant**. Previously `compile`/`run`/`tune` on any such model hard-failed — the checkpoint's
decompress pre-hook fires mid-trace and injects a data-dependent slice `torch.export` can't specialize
(`GuardOnDataDependentSymNode`).

Verified end-to-end on `cyankiwi/Llama-3.1-8B-Instruct-AWQ-INT4 --layer 0` (224 linears substituted; accuracy vs eager
`max_diff=0.0020`, `mean_diff=8e-6` ✅) plus a hermetic `tests/compiler/test_dequant.py` suite.

## Approach

A pre-trace **submodule substitution** swaps each quantized linear for a `DequantLinear` whose forward is an opaque
`deplodock::dequant_linear` custom op — keeping the packed int32 / fp16-scale / int32-zp tensors as graph constants and
stopping the decompress hook. The tracer maps it to a new frontend `DequantLinearOp`; the `045_dequant_linear`
decomposition expands it into an 8-lane int4 unpack → `(nibble - zp)` → int→fp16 cast → group-scale → transpose →
matmul cone the rest of the pipeline lowers. **Default fusion already pulls the dequant producer into the matmul
kernel**, so gmem holds only packed int32 + scale + zp — much of the Phase 3 VRAM goal, for free.

All format-specific logic is isolated in `trace/quantized.py` + the `045` rule; everything below `DequantLinearOp` is
format-agnostic.

## New IR capability: first mixed-int codegen path

- Bitwise `>>` / `&` ops (expr/elementwise/stmt), dtype-aware `Select`, integer scalar literals, a tensor-IR `CastOp`
  (+ `015_lift_cast`), and i32 `type_name` / `convert` (`__int2half_rn` etc.) / native-op support in the CUDA render target.
- `right_shift` / `bitwise_and` force an i32 result in `dtype_promote` (integer-only; no float form) so the unpack stays
  correct when a staged-smem `Load`'s dtype is unresolved at stamp time.
- dtype-preserving constant binding so packed int32 weights aren't corrupted via a float32 round-trip; `LoopOp.forward`
  raises on mixed-int (float-only runner).

## Verification status

- **Phase 0** (in-kernel dequant) — ✅ real 8B AWQ model + hermetic suite.
- **Phase 2** (granularity) — `group_size` read from the `QuantScheme` (never hardcoded); per-group GPU e2e passes at
  **G=32 and G=128** (sym + asym). Single-group (`group_size >= in_features`, per-channel/per-tensor-along-K — which no
  compressed-tensors AWQ model uses) raises a clear `NotImplementedError`; fixing the splicer is the open Phase-2 follow-up.
- **Phase 3** (VRAM) — confirmed on RTX 4080: at a realistic 4096×4096 q_proj-sized linear, deplodock gmem **9.96 MB vs
  33.82 MB** dense fp16 = **3.39× total** (4.0× on the weight itself), `rel_err 0.00045`. Guarded by
  `test_dequant_fuses_no_materialized_weight_vram`.

## Test plan

- `make test` — 2070 passed (incl. the new `tests/compiler/test_dequant.py`: numerics oracle, compressed_tensors
  parity, integer/cast codegen, substitution walker, CUDA e2e over the G=32/128 × sym/asym matrix, VRAM guard,
  per-channel guard).
- Docs updated: `plans/w4a16-quantization-support.md`, `compiler/ARCHITECTURE.md`,
  `compiler/trace/ARCHITECTURE.md`, `compiler/backend/cuda/ARCHITECTURE.md`, `compiler/ir/ARCHITECTURE.md`, `CLAUDE.md`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -12,6 +12,11 @@ The `README.md` is intentionally short — example-driven, no narrative. For det
 - **Serving** (vLLM out-of-tree embedding plugin — deplodock-compiled kernels behind vLLM's `/v1/embeddings`; `serving` extra) → [`deplodock/serving/ARCHITECTURE.md`](deplodock/serving/ARCHITECTURE.md)
 - **Recipe format** (matrices/cross/zip combinators, variant filtering, deep merge, named fields, extra_args validation, command recipes, aggregate, docker_options, driver/cuda pinning, SGLang) → [`deplodock/recipe/ARCHITECTURE.md`](deplodock/recipe/ARCHITECTURE.md)
 - **Compiler** (Graph IR dialects, passes, backends) → [`deplodock/compiler/ARCHITECTURE.md`](deplodock/compiler/ARCHITECTURE.md) and child docs
+  - **W4A16 quantization** (compressed-tensors AWQ-INT4): `compile` / `run` / `tune` handle quantized models via a
+    pre-trace substitution to an opaque `deplodock::dequant_linear` custom op, decomposed to an in-kernel int4 unpack +
+    dequant + matmul (the compiler's one integer codegen path). Entry seam:
+    [`deplodock/compiler/trace/quantized.py`](deplodock/compiler/trace/quantized.py); design + status in
+    [`plans/w4a16-quantization-support.md`](plans/w4a16-quantization-support.md).
 
 When the user asks about a CLI flag, recipe field, or matrix combinator, read the relevant ARCHITECTURE.md before answering — they hold details that are no longer in the README.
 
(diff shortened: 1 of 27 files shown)


Slonegg (changes requested):
Proper quantization will take a lot of work. Let's not complicate the compiler for now and focus on nailing the FP16 path.

PR #280, opened 2026-06-26
Generative decode: device-resident seam (~2× served decode) + tune/profile findings

6erun:
## Summary

Follow-up to #270 (generative decode-bucket). Takes **generative (chat) serving** from the post-bucket ~4.5 tok/s to
**178.7 tok/s** single-stream decode on TinyLlama-1.1B (RTX 4080, fp16) — closing the gap to native vLLM from ~6× to
**~1.5×** — via two levers found by measure-first profiling, plus the findings/scoping trail behind them.

1. **Kernel tuning** of the static decode-bucket kernels (built the learned prior from scratch): 46.5 → 85.5 tok/s.
2. **Phase A: device-resident decode seam** (this PR's main code change): 85.5 → 178.7 tok/s — the decode hot path keeps
   activations on-device (no per-layer numpy↔torch round-trip), reusing the embedding runner's captured-replay path.

## Results (TinyLlama-1.1B, fp16, RTX 4080)

| stage | single-stream decode | system (16-way) | gap to vLLM |
|---|---|---|---|
| decode-bucket only (cold prior) | 46.5 tok/s | 446 | ~5.9× |
| + kernel tune | 85.5 | 1007 | ~3.2× |
| **+ Phase A device-resident** | **178.7** | **2367** | **~1.5×** |
| stock vLLM (native, reference) | 273.1 | 3862 | 1× |

> ⚠️ **Post-merge update (this branch now contains `main`).** The 178.7 was measured **pre-merge**. After merging
> `main`'s tile-lowering refactor (#272–#279), a re-bench shows **two effects**: (1) a **stale prior** — the refactor
> renamed/restructured the decode kernels so the tuned prior's keys stopped matching → cold picks (9.3× slower),
> **recovered by re-tuning** (25.5 → 91.8 tok/s); and (2) a **residual ~2.4× lowering regression** — even after a clean
> re-tune, the decode kernels are 9.86 vs 4.06 ms (same shapes/device-path, only the compiler changed). **Phase A's
> device-residency contribution is intact** (host/dispatch unchanged); the absolute drop is a regression in the kernels
> `main` now emits on M=16 skinny-decode matmuls (known issue, owner working on it). So on **current main**: ~91.8 tok/s
> re-tuned (~25.5 if not re-tuned). Full split in `plans/generative-device-resident-decode.md` → *Post-merge re-bench*.

## How the results were measured

All numbers are a real **served** A/B — a live HTTP server + a chat-completions load client — not a micro-bench. Every
comparison point (Phase A, the tuned baseline, stock vLLM) used the **same client, flags, model, GPU and prompt**, just
pointed at a different server, so the ratios are apples-to-apples.

**Server** (the deplodock plugin via the in-process vLLM engine):

```
deplodock serve TinyLlama/TinyLlama-1.1B-Chat-v1.0 --generate \
  --gpu-memory-utilization=0.80 --max-num-batched-tokens 1024 --max-model-len 2048 --port 8001
```

The stock-vLLM reference is the same with `--stock` (native, CUDA graphs on). The low util + batched-tokens cap fit
TinyLlama's weights **plus the deplodock cupy workspaces** on the 16 GB card (the decode-bucket roughly doubled that
footprint; default `--gpu-memory-utilization=0.9` OOMs).

**Client** — `scripts/bench_gen_serve.py` (committed in this PR), against the OpenAI `/v1/chat/completions` endpoint:

```
python scripts/bench_gen_serve.py --port 8001 --max-tokens 128 --concurrency 16 --requests 64
```

- **Single-stream decode** (the headline tok/s): 3 streaming requests, `max_tokens=128`, `temperature=0`,
  `stream_options.include_usage`. Reports **steady-state decode rate, excluding TTFT** —
  `(completion_tokens − 1) / (t_last − t_first)` from the SSE token-chunk timestamps, with the exact token count from the
  final usage chunk — as the **median of 3**. TTFT is reported separately.
- **System throughput**: 64 non-streaming requests at concurrency 16 (`ThreadPoolExecutor`),
  `tok/s = Σ output_tokens / wall_time`.

**Independent cross-checks** (so the gain is real work removed, not a timing fluke):

- `scripts/profile_gen_decode.py` decomposes one runner-level decode step into wall vs pure-GPU kernel time
  (CUDA-graph-captured) vs host+dispatch — showing host/dispatch was **40%** of the tuned step, the lever Phase A
  removes. (It also showed the gap was overhead-bound, not kernel-bound — which redirected the work here rather than to
  more tuning.)
- The device path is **bit-identical** to the host path (`max|Δ|=0` on embed/pre/post/final_norm) and **token-for-token**
  vs HF eager through the real vLLM plugin (`tests/serving/test_vllm_plugin_gen_gpu.py`).

**Caveats**: localhost single-box (no network); one run per config (single-stream is median-of-3); streaming token timing
assumes ~1 SSE chunk/token (true for vLLM; the *count* is exact from usage). The ~2× is well outside run-to-run drift.

## Code change (Phase A — the only behavior change)

`deplodock/serving/gen_runner.py` + `vllm_model_gen.py`: the generative decode hot path (`num_tokens ≤ decode_bucket`)
is now **device-resident** — q/k/v and attn_out stay CUDA tensors through RoPE + vLLM paged attention, no per-layer
host hop. Reuses the embedding runner's pattern over the static M=bucket programs (`run_device`:
`upload_prefix_device` → `capture_program_graph` → `replay_program_graph` → `output_prefix_device`, torch↔cupy DLPack
zero-copy under the external torch stream). `embed_device` gathers on-device; `final_norm_device` runs a **deep-copied**
CUDA norm (the shared host norm stays on CPU). Prefill / `T > bucket` keeps the existing host path. Still under
`--enforce-eager` — the captured-replay alone also removed the per-program launch dispatch, which is why it beat the
~100–120 tok/s projection.

## Docs / tooling

- `plans/generative-device-resident-decode.md` — scoping for the device-resident + whole-step-capture work
  (Codex-reviewed; Phase A applied, Phase B + the post-merge regression still open). The one plan doc kept — the
  decode-bucket/tune/Phase-A *findings* doc was removed as fully-executed (matching main's "delete executed plans"),
  its trail preserved in the script docstrings + this PR + git history.
- `scripts/bench_gen_serve.py`, `scripts/profile_gen_decode.py` — the served-throughput and decode-step-profile
  reproducers (+ CLAUDE.md entries). `serving/ARCHITECTURE.md` updated.

## Testing

- New `tests/serving/test_gen_runner_device_path_matches_host` — device path bit-identical to host (also guards the
  in-place-norm bug a review caught).
- `tests/serving/test_vllm_plugin_gen_gpu.py` (token-for-token) still green; `make lint` clean.

## Follow-up (not in this PR)

Phase B — drop `--enforce-eager` and let vLLM capture the whole step — remains, but with **smaller headroom than
expected**, since Phase A's per-program captured-replay already removed most of the dispatch I'd attributed to it.
Scoped in the plan doc; gated on a replay-inside-vLLM-capture prototype.


Slonegg (approved):
Generally, using bucketization just bypasses the poorly optimized dynamically shaped kernels. Better to optimize dynamically shaped kernels rather than rely on a less-broken statically shaped path.

PR #296, opened 2026-07-02
Sync compute-fill transport: async B copies, vectorized fill, warp-cooperative stats

Slonegg:
## What

Squeezes the fused-edge `sync` transport (the computed-A `Contraction` operand pipeline from #295) before any
un-fuse work, so the fused form's real ceiling stays visible. The diagnosis: a ~30:1 fill-to-mma instruction ratio
— tensor cores idling behind a scalar per-element software copy with per-cell `div`/`mod` addressing, an
uncoalesced serial stat prologue, and zero overlap.

1. **Async B copies.** A canonical `[K,N]` B weight is a plain copy, so it rides vectorized `cp.async` issued
   before the compute fill — the hardware copies fly underneath it. When a two-slot ring also fits the 48 KiB smem
   budget the stage resolves at `depth=2` (`d2/sync`): the prefetched chunk's B copies stay in flight across the
   current chunk's drain. Transposed-B pipelines keep the per-thread fill (slab cells aren't gmem-contiguous).
2. **Vectorized compute fill.** Each thread owns a 16-byte run of contiguous slab cells: one `div`/`mod` per run
   instead of per cell, per-thread gmem reads / smem stores merge into wide accesses, cone stmts replicate per cell
   with a `__c<j>` suffix on run-local SSA only.
3. **Warp-cooperative stat prologue.** `sync_stat_fill` stripes one row per warp — 32 lanes stride the row's stat
   reduce coalesced and close the fold with the carrier's shuffle butterfly (`emit_combine` at warp width) —
   replacing the serial one-row-per-thread fold whose lanes strode 2 KB apart.

## Numbers (RTX 5090, fp16 `rmsnorm → gate/up → SwiGLU`, seq 512 / H 1024 / I 3072, canonical weights)

| stage | latency | vs eager |
|---|---|---|
| before (#295) | 287 µs | 0.20× |
| + vectorized fill | 179 µs | 0.32× |
| + cp.async B | 145 µs | 0.39× |
| + warp-cooperative stats | **89 µs** | **0.65×** |
| eager (two cuBLAS + pointwise) | 57 µs | 1.00× |

Tile sweeps (K-chunk 16–64, tile_n 64–128, ring on/off) plateau at 89 µs — the remaining gap is the A fill's
per-n-block redundancy (each n-block re-reads `x` and re-runs the cone) and its serialization against the drain:
warp-specialization / cross-CTA-stat-sharing territory, beyond this PR.

Full suite green (2091 passed); accuracy verified across static/masked/symbolic M, single- and multi-fold, both B
layouts (the `b_trans` sync-fallback path is exercised by the `LinearOp`-built e2e tests, the cp.async path by the
canonical traced-weight layout).

### DIFF
--- a/emmy/compiler/pipeline/passes/lowering/kernel/ARCHITECTURE.md
+++ b/emmy/compiler/pipeline/passes/lowering/kernel/ARCHITECTURE.md
@@ -137,19 +137,25 @@ flash operand) silently falls back to gmem-direct, and a staged kernel is
 **The fused edge — the mma tier's `sync` transport.** A demoted-cone matmul (`f(x, …) @ w`) takes the warp tier
 under a warp `TILE` pin: `_schedule._demoted_warp_option` nodifies the PLANAR ⊗-fold to a computed-A `Contraction`
 (the same `a_operand = Body` flash P@V rides) and stamps a `sync` `Stage`; `_staged` then builds a `SyncTransport`
-whose A fill is the producer CONE evaluated per slab cell (compute-fill) and whose B fill is a plain copy — the same
-`fill`/`commit`/`wait` seam, single-buffer, one CTA barrier, feeding the unchanged `ldmatrix` drain. A
+whose A fill is the producer CONE evaluated per slab cell (compute-fill) — the same `fill`/`commit`/`wait` seam,
+feeding the unchanged `ldmatrix` drain. The compute fill assigns each thread a 16-byte run of CONTIGUOUS slab
+cells (the row/col derivation hoists out of the per-cell code; per-thread gmem reads and smem stores merge into
+wide accesses; the cone replicates per cell with a `__c<j>` SSA suffix). A canonical (non-transposed) **B** is a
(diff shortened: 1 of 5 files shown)


Slonegg:
Second commit: the stage resolvers now size against `Context.max_dynamic_smem` (99 KiB on sm_120) instead of the
hardcoded 48 KiB `STATIC_SMEM_CAP` — the backend already handled `extern __shared__` + `cudaFuncSetAttribute` past
the static floor, so this just threads `ctx` from the recognize rewrite into `schedule()` and the resolvers.

Empirical note: with the real budget, the sync ring (`depth=2`) started auto-engaging at the best tile and
regressed it (89 → 140 µs — occupancy cost at 2× slot bytes, and the B copies were already hidden under the
compute fill). The unpinned default therefore stays single-buffer; `EMMY_STAGE=d2/sync` opts the ring in, now
actually reachable at big K-chunks for the tuner to explore.

PR #302, opened 2026-07-02
Warp-flash K/V operand staging: the STAGE@kv cp.async stream (article Moves 4-5)

Slonegg:
## Summary

The warp (mma) flash tier's K/V operands now stage through per-block shared-memory slabs on the same `staged_kloop` skeleton the matmul tier runs — closing the "K/V operand staging on the warp-flash stream" follow-up from #300 and giving the Part-2 article its Move-4/Move-5 example kernel.

- **Scheduler** (`_schedule`): `_resolve_twisted_stage` sizes the K `(bn × head_dim)` + V `(bn × d_v)` slot pair against the smem budget net of the resident `flash_pv_smem` handoff slab. cp.async only (TMA's 2-D descriptor can't encode the batched K/V operands — this is the article's sm_120 rung), static block-divisible kv only. Each warp geometry row crosses with its resolver-gated stage candidates, stamped `STAGE@<kv>` (resolved spelling / decided-empty `""`); the chain/coop/serial escapes stamp the decided-empty key, keeping every flash leaf on one key set.
- **Materializer** (`_twist`): the streaming step is now a ring-slot-parameterized builder; a resolved `ctx.stage` re-parents it as the `staged_kloop` drain (`d1` single-buffer, `d2+/ring` prefetch). Slabs keep each operand's own layout — verbatim row copies, so **staged output is bit-identical to gmem-direct** (test-enforced).
- **Renderer**: the staged transposed-B drain is implemented — plain `ldmatrix.x2` (no `.trans`): the N-major K slab's 8×8 rows ARE the mma's col-major B columns. Replaces the `NotImplementedError`.

## Measured (RTX 5090, (1,8,4096,64) f16)

Scalar chain 24943µs → mma gmem-direct 612µs → **staged `d1/cp` 511µs** (winner: `w2x1/f1x16/k4`); eager FA-2 206µs. The `d2+` ring regresses at this occupancy — NCU attributes the remaining gap to smem bank conflicts on the unswizzled slabs + the C→A handoff (follow-ups listed in `plans/flash-article-kernel-findings.md`).

## Test plan

- `make test`: 2135 passed, 39 skipped (includes new staged accuracy {d1,d2,d3} × {causal, GQA}, bit-identity staged-vs-gmem, symbolic-decline, and the stage-crossed fork-grid assertions)
- `make lint`: clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #303, opened 2026-07-03
RTX 4080 (sm_89) GpuSpec + golden matmul configs

6erun:
## Summary

Add RTX 4080 (sm_89) support: a `GpuSpec` for the card (pci 2704, 76 SMs, cc 8.9) and a 23-shape golden
matmul config set (`rtx4080_sm89.yaml`), so emmy recognizes and deploys on the 4080 — an sm_89 sibling of
the 4090.

## Notes

- Schema matches the existing per-GPU golden yamls (4090 / 5090 / pro6000); the 23 configs load and parse
  cleanly, and `tests/compiler/test_golden_configs.py` passes (30 tests).
- The golden knobs are a **pre-ring baseline** — recorded before main's scalar cp.async ring landed — so
  they can be re-tuned on the 4080 to pick up the ring.

## History

This branch previously also carried a scalar cp.async prefetch ring, but main independently shipped an
equivalent (more complete) ring via #293 / #296 / #302, so that work was dropped as redundant. Two
correctness deltas main still lacks — a bk-degrade fix and a masked-N staging gate — are being re-verified
against main for a separate focused PR.

### DIFF
--- a/emmy/compiler/pipeline/search/goldens/rtx4080_sm89.yaml
+++ b/emmy/compiler/pipeline/search/goldens/rtx4080_sm89.yaml
@@ -1,0 +1,191 @@
+gpu_name: NVIDIA GeForce RTX 4080
+compute_cap: [8, 9]
+configs:
+  - kernel: matmul
+    name: square.512
+    M: 512
+    N: 512
+    K: 512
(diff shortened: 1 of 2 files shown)


Slonegg (approved):
Seems legit 👍

PR #310, opened 2026-07-05
Flash on model-graph V layouts: materialize the folding V indexmap, per-operand GQA groups

6erun:
## Summary
Qwen3-Embedding layer-0's attention ran the un-fused 3-kernel scalar SDPA path at ~9.4 ms (the softmax→P@V slice alone 6.7 ms, ~65× behind eager — `plans/qwen3-embedding-06b-layer0-tune-findings.md` finding 1). Root cause was **not** the report's RoPE'd-Q/K hypothesis: the tensor-core flash tier already ships and the fusion boundary already keeps RoPE out of the score producer. Recognition died on the **V operand**: model graphs feed SDPA a V reshaped from the flat `[B, S, Hkv·D]` projection, and the folded `kv_head·D + d` load slot de-certified `_extract_v_layout`, silently degrading the fuse to cut.

Three general (shape-agnostic) rules fix it:
- **`frontend/optimization/010_compose_indexmaps`**: an axis-folding indexmap (a coord entry reading ≥2 output placeholders) composes through rank-preserving maps only — the composite stops before the matmul's unsqueeze/broadcast scaffolding, so the copy that later materializes is the operand's canonical `(b, h, s, d)` form (what the warp fragment loaders assume).
- **`loop/fusion`**: a folding indexmap loses the pure-indexmap exemption at the flash guards (`010_merge_loop_ops` + the multi-consumer `005_split_shared_indexmap`) and stays materialized — a ~µs layout-copy kernel instead of a de-certified offer site. The predictive exp scan resolves through pure-indexmap wrappers, and the guards judge the compute consumer at the end of a scaffolding chain (`through_indexmap_users`). Shared predicates moved to `_helpers`.
- **`lowering/tile/_flash`**: per-operand GQA groups `(group_k, group_v)` — K's slot-plain repeat fuses into the score producer while V arrives via its materialized copy, so the operands legitimately reach recognition at different head counts.

Also updates the stale "flash lowers only on the scalar tier" docstrings in `_flash.py` / `_atomize.py` / the test header (pre-#300 — the warp tier exists).

## Verification (RTX 4080)
- New tests: folded-V projection fuses to one warp-flash kernel (+1 copy kernel) and matches torch; explicit-KV-repeat variant exercises `gk=2, gv=1`; a no-GPU structural test asserts the flash fork rows are enumerated for the folded-V layout.
- Real qwen layer-0 SDPA reproducer: recognition certifies; cold prior deploys warp mma flash (was: scalar cut).
- Clean layer-0 re-tune (`--layer 0 --dynamic seq_len@x:1 --clean --bench`, isolated caches): attention deploys as **one `k_scaled_dot_product_attention_reduce` mma kernel at 87 µs in-layer** (was 9,444 µs across three kernels). Layer e2e 8,880 → 7,438 µs; the residual is the projection-linear scalar pick-miss (q_proj main 1,461 µs `n16x8/f2x2`, o_proj's in-layer deploy diverging from its 85.5 µs MMA DB pick) — the separate ranking-lever follow-up from `plans/next-steps-flash-folded-v-and-mma-pick.md`, not an attention issue.
- `make test`: 2070 passed; the 4 failures (`test_contraction_emits_matmul`, 3× TMA staging) are pre-existing on origin/main on this sm_89 box (reproduced with the diff stashed). `make lint` clean.


Slonegg (changes requested):
Let's not mess with the fusion for now

PR #311, opened 2026-07-06
Projection linears onto tensor cores: analytic tier gates + a deployable-evidence lane for mma

6erun:
## Summary
Independent of #310 (restacked onto main). The qwen3-emb / gemma-4-e2b layer-0 reports found every projection linear (q/k/v/o_proj, PLE) deploying scalar split-K tiles at 5–20× the -O3 cost of their enumerated mma siblings. Two fixes, complementary to #317's golden refit:

**1. Analytic tier gates — the structural invariant behind #317's data-side fix.** The dynamic weight set ranked scalar rows first on warp-eligible f16 contractions because no feature told the prior the alternative existed. #317's `_W_A_DYN` refit fixes the ranking for the current golden distribution; these gates encode the interaction a linear weight can't express — "a scalar tile where tensor cores were on offer must clear the roofline bar" — so the property holds independent of future refits and off-distribution shapes, and the new stamp feeds the learned prior the same signal.
- `_schedule._tile_rows` stamps `S_warp_eligible` onto every enumerated row when the kernel offers any tensor-core row (an `S_` structural fact; row identity / `tile_signature` untouched).
- `features._geom_feats` adds `D_scalar_on_warp_eligible` and `D_splitk_roundtrip` (the `g<w>k` deferred finalize's workspace round-trip — the un-priced `D_cut_roundtrip` analogue).
- `AnalyticPrior` gains two hard-coded gates (the `atomic_free_weight` precedent).
- Verified vs the refit weights on main: identical golden ranks (no over-rotation), identical dynamic-regime top picks; `tests/compiler/pipeline/search/test_analytic_tier_preference.py` pins mma-above-every-scalar in both weight regimes.

**2. A deployable-evidence lane for mma rows (load-bearing; #317 hit this exact failure — "deployed greedy kept picking scalar until retrained").** The -O3 rebench sampled only configs within 15% of the best **-O1**, and -O1 overprices register-tile mma kernels 1.5–3× (the reports' finding 4) — so mma configs never earned `H_opt=3` evidence rows and the evidence-first deploy hierarchy kept deploying the scalar rows that had them. Per review, the common `O3_REBENCH_TOL` is raised to 2.0 (within 3× of best -O1, bounded by the per-config dedup; `EMMY_O3_TOL` overrides), so a fresh tune produces mma -O3 evidence instead of requiring a golden-dataset retrain to displace stale scalar rows.

## Measured effect (RTX 4080, qwen3-emb layer-0, dynamic seq_len, 512 hint — with #310 in the tree)
| Deploy path | before | after |
|---|---|---|
| Cold analytic prior | 6.6–7.4 ms (scalar splits, one 5.4 ms partial = 82%) | **1.46 ms** — all projections `a:mma` `g2k` partials |
| Freshly tuned (learned prior) | 7.4 ms | 5.9–6.1 ms (healthy per-kernel picks: q/v_proj 51/29 µs at 0.5× eager, down_proj 80 µs — minus one anomalous launch, below) |

Without #310, attention stays on the un-fused ~9.4 ms path and dominates the layer number, but the projection-pick improvements stand alone.

## Known issues documented (follow-ups, not addressed here)
- **A constant ~5.4–5.5 ms cost attaches to ONE kernel launch in layer benches on this box** — position-stable per run, moves across kernels/configs between runs, appears in standalone reproducer benches — while the same variant benched via `emmy run --code` at the same shape/symbolic-M runs 26 µs. Likely related to the gemma k_proj ">1 s hang" class and possibly #317's "attention golden re-bench pathology"; needs its own investigation.
- **gemma-4-E2B layer-0 fails accuracy (`output mul_ contains NaN`) pre-existing** — reproduced on a baseline without these changes; blocks re-validating that layer end-to-end. The historical g4a hang pick is dodged by the mma routing.
- The tune's 4 s per-variant compile budget and reproducer coverage flags (reports' findings 3/5) remain open tune-infra items.

## Test plan
- `tests/compiler/pipeline/search/` + move-catalog + goldens permanence: 184 passed on this base; full `make test` green minus the 4 pre-existing sm_89 failures (`test_contraction_emits_matmul`, 3× TMA staging).
- `make lint` clean.

### DIFF
--- a/emmy/compiler/pipeline/search/policy/mcts.py
+++ b/emmy/compiler/pipeline/search/policy/mcts.py
@@ -44,22 +44,16 @@ if TYPE_CHECKING:
     from emmy.compiler.pipeline.search.prior import Prior
 
 # Re-bench at -O3 any config whose -O1 latency is within this fraction of the best
-# -O1 so far (not just a strict new best). -O1 is the fast ranking compile but
-# ties configs that diverge at -O3 (deployable) — e.g. a warp-tier WARPSPEC /
-# occupancy split where the -O1 latencies are within ~1% but -O3 differs ~15%.
-# Widening the deployable-sample net to this band feeds the prior -O3 truth for
-# every near-best contender, so it can rank by -O3 cost. Env-overridable via
-# ``EMMY_O3_TOL`` (a fraction, e.g. ``0.15`` for 15%).
-O3_REBENCH_TOL = 0.15
-
-# Warp (mma) rows get a wider -O3-worthiness band: the -O1 ranking lane systematically
-# overprices register-tile mma kernels (1.5–3× vs -O3 — the cicc unroll blowup the -O1
-# ranking compile dodges), so an mma config that loses the 15% -O1 band can still be the
-# -O3 deploy winner. Without a deployable sample the evidence-first deploy hierarchy only
-# ever sees scalar -O3 rows and keeps deploying them (the qwen3-emb layer-0 projections:
-# scalar g2a evidence at 394 µs deployed over the 70 µs-class mma picks). Cost is
-# bounded: ``_o3_done`` dedups per config, one -O3 compile + short bench each.
-O3_WARP_REBENCH_TOL = 2.0
+# -O1 so far (not just a strict new best). -O1 is the fast ranking compile but its
+# ranking diverges at -O3 (deployable): register-tile mma kernels run 1.5–3× slower at
+# -O1 (the cicc unroll blowup the ranking compile dodges), so an mma config outside a
+# narrow -O1 band can still be the -O3 deploy winner — and without a deployable sample
+# the evidence-first deploy hierarchy only ever sees scalar -O3 rows and keeps deploying
+# them (the qwen3-emb layer-0 projections: scalar g2a evidence at 394 µs deployed over
+# the 70 µs-class mma picks). The band covers that skew (within 3× of best -O1); cost is
+# bounded by ``_o3_done``'s per-config dedup — one -O3 compile + short bench each.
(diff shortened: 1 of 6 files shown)


emmy/compiler/pipeline/search/policy/mcts.py
```diff
@@ -52,6 +52,15 @@
...
 O3_REBENCH_TOL = 0.15
 
+# Warp (mma) rows get a wider -O3-worthiness band: the -O1 ranking lane systematically
+# overprices register-tile mma kernels (1.5–3× vs -O3 — the cicc unroll blowup the -O1
+# ranking compile dodges), so an mma config that loses the 15% -O1 band can still be the
+# -O3 deploy winner. Without a deployable sample the evidence-first deploy hierarchy only
+# ever sees scalar -O3 rows and keeps deploying them (the qwen3-emb layer-0 projections:
+# scalar g2a evidence at 394 µs deployed over the 70 µs-class mma picks). Cost is
+# bounded: ``_o3_done`` dedups per config, one -O3 compile + short bench each.
+O3_WARP_REBENCH_TOL = 2.0
```
Slonegg: Just increase common rebench tol

Slonegg (approved):

emmy/compiler/pipeline/search/policy/mcts.py
6erun: But it will increase number of rebenches. Are you ok with it?

Slonegg: We will see and correct if needed. There are too many special hacks to account for in the codebase.

PR #313, opened 2026-07-06
rtx4090 golden sweep: #269 regression findings, re-baselined goldens, native-keyed golden support

gnull:
Deliverables from the 2026-07 rtx4090 perf-tuning session (golden sweep + whole-model tune on a rented 4090 VM).

## What's here

1. **Golden sweep findings** (`plans/golden-sweep-rtx4090-findings.md`) — headline: the recorded golden knob sets
   re-bench at **median 1.66× their recorded latency** (worst 4.07×) while cuBLAS reproduces exactly; bisected to
   the single commit `8c7d2d54` (#269 Block-DAG Tile IR). Identical-knob CUDA diff shows the new assembly drops
   A-operand smem staging (gmem-direct fallback), the RING/STAGE cp.async ring pipeline, and the GROUP_M swizzle.
   Per the team's direction this is a **known, accepted** consequence of the rewrite (optimizations to be
   reintroduced gradually); the report's per-shape drift column is the reintroduction scoreboard.

2. **Golden YAML re-baselined** to live post-#269 -O3 numbers (same knobs; pre-#269 targets stay recoverable via
   `git show 272153d3:<yaml>` — noted in the YAML header). One knob win recorded: `square.512` + `SPLITK: 4`
   (19.3 → 17.4 µs, reproduced 3×). **Review decision to confirm:** re-baselining in-file vs keeping pre-#269
   targets in the YAML — this PR takes the "YAML = current reality, git history = the target" position.

3. **Native-keyed golden support** (code):
   - `config.knob_var` maps `@`→`_` so a native `MOVE@element` knobs dict (`SPLIT@a0`) pins as the documented
     `EMMY_SPLIT_A0` env var — previously a silent no-op, making native-keyed golden replay impossible.
   - `eval golden` / `eval prior --dataset golden` now diff **native-vs-native** (both sides compiled to the
     kernel dialect; a legacy-keyed golden's pins translate through the `_knob_legacy` ingest ramp). Fixes the
     all-dash "found" columns every post-#269 run produced.
   - Shared `pipeline.knob.pinned_knobs` (was `run.py`-private); 2 new unit tests.
   - Six goldens where greedy beat the legacy-recorded knobs now carry greedy's config natively, each verified
     against the benched sweep row + a live replay A/B (11.2 µs vs the stale entry's 24.0).

4. **Whole-model tune findings** (`plans/qwen3-embedding-0.6b-tune-findings.md`) — Qwen3-Embedding-0.6B, dynamic,
   2 h 21 m clean tune. Two urgent bugs documented with repros, to be fixed in a **follow-up PR**:
   - bench worker can't feed split-producer intermediates (`KeyError('linear_196__xn')`) — silently disables
     whole structural branches, crashes the whole-model `--bench` table, wastes the search budget;
   - `emmy serve` dies at startup with a trained prior — greedy deploys an un-lowerable config
     (`StageBundle: requires at least one Source`); the search prunes that state, the deploy path doesn't.
   Stock-vLLM serving baseline recorded (103.7 req/s / 53.1k tok/s) for the eventual A/B.

## Validation

`make test` on the rtx4090 VM: **2250 passed, 124 skipped**. `make lint` clean. Golden schema test 28/28.
Sweep/tune logs + dump artifacts on the VM under `_tune/` (paths in the reports).

### DIFF
--- a/emmy/commands/eval.py
+++ b/emmy/commands/eval.py
@@ -627,34 +627,42 @@ def _perf_cell(perf: dict | None, name: str) -> tuple[str, str] | None:
 
 
 def _emit_prior_golden_check(configs: list, *, title: bool = True, perf: dict | None = None) -> None:
-    """Greedy fork pick through the tile pipeline vs recorded golden. The pick reads
-    the learned-prior JSON (``config.prior_path()``: ``EMMY_PRIOR_FILE`` /
-    ``--prior``); option-0 with no fitted prior. Stops at the tile dialect (every
-    knob fork resolves there: no codegen / nvcc). One row per shape (configs sharing a
-    name share a snippet → one greedy pick): the pick is scored against the shape's
-    *closest* recorded golden (most knobs reproduced), so multiple goldens for a shape
-    don't duplicate rows. A trailing ``TOTAL`` row carries per-knob match counts over the
-    deduped rows + the exactly-reproduced row count. Rows print with column-aligned
-    ``found/golden`` knobs (canonical order). ``title`` prints the
+    """Greedy fork pick through the lowering pipeline vs recorded golden, diffed in
+    the **native** ``MOVE@element`` knob vocabulary. The pick reads the learned-prior
+    JSON (``config.prior_path()``: ``EMMY_PRIOR_FILE`` / ``--prior``); option-0 with no
+    fitted prior. Both sides are compiled to the kernel dialect (knob stamping
+    completes there — still no codegen / nvcc): the greedy side unpinned, the golden
+    side with the recorded ``knobs`` pinned (``pinned_knobs``), so a legacy-keyed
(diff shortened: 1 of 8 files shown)


gnull:
Superseded by events on main: the Tile IR was rebuilt (#293) with a new knob codec, goldens re-tuned (#294/#298/#304/#308), and the #269-window regression this PR documented has been recovered (verified live on the 4090: square.512 golden row 15.2µs vs recorded 13.5; fp16 6.7µs). The branch stays as the record of the regression-window analysis + two still-relevant leads (cold-greedy STAGE mispricing, pooling-head hang, emmy-serve cold-start 1.95 req/s vs stock 103.7) — next step is a fresh golden tune on current main, then we cherry-pick what still applies.

PR #325, opened 2026-07-08
Mixed-dtype (f32-A × 16-bit-B) contractions ride the mma tier via the demoting sync compute-fill

6erun:
## Summary
**Reworked per review** — the original tracer + fusion-boundary approach is fully replaced by a **downstream-only** fix in the tile scheduler. No fusion files are touched; fusion's fuse-everything invariant stays intact. (Previous approach preserved on `feature/materialize-dtype-casts` history for reference.)

An f32-A × 16-bit-B contraction can only enter a traced graph through an erased dtype cast: torch cannot execute a mixed matmul, so the model itself cast one side (gemma's `self._norm(x.float()).type_as(x)` rounds A to f16 before every projection; the tracer maps `to`/`type_as` to identity pass-throughs). The warp tier's 16-bit-A gate (`_warp_atoms`) locked those contractions into scalar tiles — gemma-4-12B's q_proj at 1573µs (nsys) where cuBLAS does 185µs; the down-projection at ~98ms (the latencies #324's attribution fix untangled).

**The fix** (`lowering/tile/_schedule.py` only):
- `_demoted_atoms`: the 16-bit atoms an f32-A contraction may ride — every fold's B `Load` shares one 16-bit dtype. (The converse erased-upcast graph — `w.float()` on a 16-bit weight — shows B=f32 and never triggers.)
- `_demote_mixed_a`: re-expresses the plain f32-A `Load` as a computed-A cone (`a_operand=Body((load,))`), so it rides the **existing** mandatory sync compute-fill, whose slab store demotes to the atom dtype. Copy transports (gmem-direct ldmatrix / cp.async / TMA) move raw bytes and can never convert — that's why the demotion routes through the cone form.
- The demoted mma rows are fork **siblings** of the scalar rows (searchable, pinnable, never forced); B's values genuinely carry 16 bits and the accumulate stays f32, so the demotion costs ~2⁻¹¹ relative noise on A — the rounding the model performed anyway in the erased-downcast case.
- Bonus: fixes a pre-existing `StopIteration` in `_demoted_warp_option` on square folds (`multiply(x,x)` — mean-of-squares), which crashed any warp-`TILE`-pinned compile of a graph containing a norm.

## Verification (gemma-4-12B q_proj reproducer — unchanged f32×f16 graph, RTX 4080)
- Search measures **140 variants, all top ranks `a:mma_…/d1/sync`**, pick rank 1/140.
- Pinned winner deploys at **303µs -O3 vs 1573µs scalar before (5.2×)**; whole-program 307µs vs 1580µs; accuracy passes. (cuBLAS 185µs — the remaining gap is sm_89 mma schedule quality, the known projection-class gap.)
- New e2e test (`test_fused_edge.py`): the mixed matmul engages mma under a warp pin and matches the f16-rounded-A numpy reference; scalar cell covers the unpinned fallback.
- `make lint` clean; `make test` 2072 passed, same 9 pre-existing failures on this box as main (2× `--golden` CLI tests needing h4096 seeds on the live card, 7× staged-flash TMA post-#321).

## Notes
- Cold greedy still needs prior/tune data to *pick* the mma rows over scalar — the standard two-level-autotune story, same as every fork (verified: a tiny isolated prior misranks; the search itself ranks the mma rows first).
- Split-K rows are not offered on the demoted form this cut (same as other computed-A cones).

### DIFF
--- a/emmy/compiler/pipeline/passes/lowering/tile/_schedule.py
+++ b/emmy/compiler/pipeline/passes/lowering/tile/_schedule.py
@@ -319,7 +319,10 @@ def _warp_atoms(kernel, probe) -> tuple[str, ...]:
     """The dtype-eligible tensor-core atom names for this contraction, ``()`` when the warp tier
     doesn't apply (unbindable node, a non-16-bit operand dtype, or a fragment-unrealizable gather
     epilogue). A computed-A (fused-cone) node reads its operand dtype off the cone's K-indexed
-    ``Load`` — the value the sync compute-fill stores to the A slab."""
+    ``Load`` — the value the sync compute-fill stores to the A slab. A cone whose leaf is **f32**
+    may still ride the folds' 16-bit atom (:func:`_demoted_atoms`) — the fill demotes on the slab
+    store; the plain-``Load`` form stays 16-bit-only here (copy transports move raw bytes and can
+    never convert) and reaches the warp tier through :func:`_demote_mixed_a`'s cone wrap instead."""
     if probe is None or not kernel.inputs:
         return ()
     if not _fragment_epilogue_ok(probe.epilogue):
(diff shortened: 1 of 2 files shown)


emmy/compiler/pipeline/passes/loop/fusion/010_merge_loop_ops.py
```diff
@@ -252,6 +253,13 @@ def rewrite(match: Match, producer: Node, consumer: Node) -> Graph | None:
     if producer.id not in consumer.inputs:
         raise RuleSkipped(f"producer {producer.id!r} is not an input of consumer {consumer.id!r}")
 
+    # Dtype-boundary protection: a materialized narrowing cast must not dissolve forward —
+    # load-rewiring would hand the consumer the PRE-cast buffer, dropping the rounding and
+    # (e.g.) locking a matmul consumer out of the f16 mma tier on an f32 A operand. The
+    # cast's own producer side still fuses INTO it (this guard only checks the producer).
+    if _is_dtype_boundary(graph, producer):
```
Slonegg: Let's not mess with the fusion! Check if scheduling and codegen can be fixed first.

Slonegg (changes requested):

emmy/compiler/pipeline/passes/loop/fusion/010_merge_loop_ops.py
6erun: It's just simple check. I think it's required to proceed with optimizations.

Slonegg:
The invariant is fuse everything, then cut if needed. Changes in fusion can
break a lot of optimizations. So, if you want to make this change you need
to prove that it is not possible to introduce the fix downstream.

6erun:
Reworked as requested — you're right that this didn't need fusion changes, and the new version proves it: the fix now lives entirely in the tile scheduler (`_schedule.py`), fusion untouched.

The key realization: the erased cast's *intent* survives fusion as a dtype **signature** — a plain f32-A `Load` against uniformly 16-bit fold B operands can only mean the model cast one side (torch can't run mixed matmuls), and since B genuinely carries 16 bits, demoting A on the slab store is faithful in the dominant (erased-downcast) case. So instead of preserving a boundary through fusion, the scheduler re-expresses the mixed plain-Load contraction as a computed-A cone and rides the existing mandatory sync compute-fill — the one transport that evaluates values (and thus converts) rather than copying bytes. The demoted mma rows are fork siblings of the scalar rows, so nothing is forced.

Same outcome as before on the gemma reproducer (303µs -O3 vs 1573µs scalar, 5.2×), plus it now also fixes *old* traced graphs and reproducers, not just fresh traces — the tracer change became unnecessary and is dropped.

One pre-existing bug fixed in passing: `_demoted_warp_option` raised `StopIteration` on square folds (`multiply(x,x)` — any mean-of-squares norm) under a warp `TILE` pin; it now degrades gracefully.

Slonegg (approved):

Issue #328, opened 2026-07-08
Tune needs a numerics gate before a measurement becomes deploy evidence


gnull:
## Summary

The tune's benchmark measures latency only — it never validates the kernel's output against a reference. A broken kernel that happens to run fast is therefore recorded as a legitimate `ok` measurement in the tune DB and the prior's training reservoir. Since #326, measured evidence *preempts* the learned model at `compile`/`run` (reservoir -O3 rows first, tune-DB rows second, model prediction only where nothing was measured) — so a fast-but-wrong measurement is no longer just metric noise: it is deployable.

## History — why this became load-bearing

- #222/#223 made greedy deploys prior-driven; #304 added the measured-evidence lane so a config the tune proved fastest can't lose the deploy to a model extrapolation.
- #311 unintentionally disabled both evidence tiers (a new `S_*` feature stamp broke the strict signature join), which #326 fixed — and in fixing it, made measurements decisive again. The assumption that a recorded `ok` latency is *true* is now a correctness assumption of the deploy path, and nothing enforces it.

## Concrete incident (2026-07-08, RTX 4090)

An ε-greedy tune of the `matmul.mlp_gate_up.h4096(.dynM)` goldens (512×4096×28672 fp16 GEMM) reached the `TILE=a:mma_m16n8k16_f16/w4x4/f4x8/k8, REDUCE=g2k, STAGE=d1/cp` config, which compiles to a kernel that computes garbage fast (lowering bug, filed as #327). It was measured and recorded as the shape's best at **105.8 µs (static) / 103.5 µs (dynM)** — physically impossible for the shape (~120 GFLOP → >1 PFLOPS-equivalent; cuBLAS ~720 µs; correct emmy configs 900+ µs). The tune summary's own silly-rate stat flagged it indirectly (93–99 % of benches ≥2× "best").

No wrong deploy shipped, but only by accident: the bogus rows happened never to reach the perf table, and their -O3 re-bench never landed in the reservoir — so the `H_opt=3` evidence tier that decides deploys stayed clean and `run --bench --golden` kept shipping the measured-good config. A slightly different failure (a garbage kernel that survives the -O3 recompile) would have deployed.

## Reproduction

```bash
# The run path's accuracy check catches the broken kernel (fails vs torch) — the tune path has no such check:
emmy run --bench -c "torch.matmul(torch.randn(512,4096,dtype=torch.float16), torch.randn(4096,28672,dtype=torch.float16))" \
[...]

PR #330, opened 2026-07-08
Tile lowering: decline over-budget warp stages; bench_fail un-lowered terminals (#327)

### LINKED ISSUE #327: Tile lowering: w4x4/f4x8/k8 warp tile emits broken programs (unlowered TileOp with g4k/d1; fast-but-wrong kernel with g2k)
## Summary

On the 512×4096×28672 fp16 GEMM (the `matmul.mlp_gate_up.h4096` golden shape), the `TILE=a:mma_m16n8k16_f16/w4x4/f4x8/k8` warp tile produces broken programs in two distinct ways, depending on the reduce/staging siblings:

1. **Compiler invariant violation** — with `REDUCE=g4k, STAGE=d1/cp` the pipeline emits a graph whose `matmul__partial` node is still a `TileOp`; the CUDA backend rejects it at bench time:
   ```
   TypeError("CudaBackend: node 'matmul__partial' has non-CudaOp 'TileOp'; lowering must produce Graph[CudaOp].")
   ```
   The tune marks it `bench_fail`, but this is a lowering bug, not an invalid config — an unlowerable combination should be rejected by the rule (not offered), or lower correctly.

2. **Fast-but-wrong kernel** — with `REDUCE=g2k, STAGE=d1/cp` the program compiles, launches, and *measures* at **~104–106 µs**. That is physically impossible: the shape is ~120 GFLOP, so 104 µs ≈ 1.15 PFLOPS — several times an RTX 4090's fp16 peak. cuBLAS runs it at ~720 µs; correct emmy configs run 900+ µs. The kernel presumably covers a fraction of the output or exits early. Because the tune's bench does not validate numerics (see #328), this measurement was recorded as the shape's `ok` "best".

## Environment

RTX 4090 (sm_89), CUDA 12.9, the #326 branch at `6820b381` (the tile-lowering code there is unchanged from `main`). Found by an ε-greedy tune (`--explore-eps 0.25`) of the gate_up goldens — the ε exploration is what reached this warp-tile family; deterministic PUCT avoids it, so ordinary sweeps mask the bug.

## Reproduction

Both modes, one command each (any machine with a CUDA GPU; the shape/knob combination is the trigger — sm_89 is not known to be required):

```bash
# Mode 2 — fast-but-wrong: the ab row benches implausibly fast and the run's accuracy check vs torch fails
emmy run --bench -c "torch.matmul(torch.randn(512,4096,dtype=torch.float16), torch.randn(4096,28672,dtype=torch.float16))" \
  --ab "TILE=a:mma_m16n8k16_f16/w4x4/f4x8/k8,REDUCE=g2k,STAGE=d1/cp"

# Mode 1 — unlowered TileOp: same command with the g4k sibling
emmy run --bench -c "torch.matmul(torch.randn(512,4096,dtype=torch.float16), torch.randn(4096,28672,dtype=torch.float16))" \
  --ab "TILE=a:mma_m16n8k16_f16/w4x4/f4x8/k8,REDUCE=g4k,STAGE=d1/cp"
```

How it was found (tune-side): `EMMY_TUNE_PATIENCE=100 emmy tune --dataset golden --kernel matmul.mlp_gate_up.h4096 --explore-eps 0.25` on a 4090 — repeated `TypeError` bench_fails for the `g4k`/`d1/cp` siblings, and a per-shape "best" of 105.8 µs (static) / 103.5 µs (dynM) from the `g2k`/`d1/cp` sibling. The post-fit silly rates of 93–99 % (nearly every bench ≥2× "best") are the telltale signature in the tune summary.

## Impact

- Tune budget burned on a broken family (each `TypeError` costs a full compile before failing).
- The fast-garbage measurement enters the tune DB / prior reservoir as `ok` data — a mis-training signal always, and a potential deploy hazard now that measured evidence preempts the model at deploy (#326). In the observed incident the deployable evidence tiers stayed clean by accident (the bogus rows never got an -O3 re-bench), so no wrong deploy shipped — but that is not a guarantee. Companion issue: #328.

gnull:
Fixes #327 — the `w4x4/f4x8/k8` warp tile's broken programs (unlowered `TileOp` with `g4k/d1/cp`; the physically-impossible ~104 µs "ok" rows with `g2k/d1/cp`).

## Root cause

Both modes are one bug with two symptoms. That tile is a 256×256 CTA tile with a 128-element K-chunk, so a single operand-staging slot is 131072 B — over the RTX 4090's 101376 B dynamic-smem cap. `_resolve_warp_stage` clamped the ring depth with `max(1, budget // slot_bytes)`: the floor of 1 made an unfittable slot **resolve at d1 instead of declining**, so the fork offered a row that could only die at `010_materialize`'s `validate(ctx)`. Under tune a validation-filtered rewrite leaves the node un-lowered, silently.

- **Mode 1** (`g4k`): the un-lowered `matmul__partial` `TileOp` reached the CUDA backend → the `TypeError` bench_fails, one full compile wasted per row.
- **Mode 2** (`g2k`): `_TerminalBench` collects only `CudaOp` nodes, so the un-lowered partial was invisible to the bench; the graph's only visible kernel was the cheap split **finalize**, whose cached perf row made `prelude()` take the full-cache-hit path and report finalize-only time as an `ok` terminal measurement. The finalize benches at ~92 µs on the 4090 — the issue's ~104 µs "1.15 PFLOPS matmul". Which sibling got the garbage-ok vs. the TypeError was just cache order during the ε-tune.

## Fix

- **`_schedule._resolve_warp_stage`** declines (`None` → gmem-direct) when even a depth-1 slot exceeds the budget. The warp slab is codec-sized (`bk` is spelled, not derived) so unlike the scalar resolver it cannot shrink — declining is the only honest resolution. Unfittable stage rows are no longer offered by the fork, and a `STAGE` pin degrades to gmem-direct, restoring the resolver contract the scalar/sync resolvers already keep: everything offered materializes within budget.
- **`_TerminalBench`** now detects kernel-bearing nodes a rewrite left un-lowered (same membership test as the backend's `_launches` walk) and returns `bench_fail` from `prelude()` before any bench or cache lookup. A partial kernel sum can never be recorded as `ok` again, whatever leaves a node un-lowered in the future — this half is defense in depth beyond #327's specific trigger (numeric validation of benches stays #328).

## Verification

- Both issue repro commands now lower and bench honestly on the 4090: the ab rows run gmem-direct (~29 ms partial + finalize, `STAGE=""` stamped per the honest-stamping rule) — no un-lowered node, no TypeError, no impossible timing.
- New regression tests: every enumerated warp row with a non-empty `STAGE` fits its depth-1 slot in the ctx budget and the over-budget tile stays enumerable gmem-direct with its split-K rows (`test_move_catalog.py`); an un-lowered terminal is `bench_fail` both with no CudaOps at all and with a cached residual kernel — the exact mode-2 scenario (`test_lowering_error_guardrail.py`).
- `make test` on the 4090: 2074 passed; the 8 failures (7 TMA-flash, 1 bench_block dry-run) reproduce identically on unmodified HEAD in the same environment (sm_89 has no TMA) — zero new failures. `make lint` clean.

### DIFF
--- a/emmy/compiler/pipeline/ARCHITECTURE.md
+++ b/emmy/compiler/pipeline/ARCHITECTURE.md
@@ -201,7 +201,10 @@ engine loops (`drive` for exploration, `resolve` for deterministic resolution):
   `TuningSearch(patience=, ucb_c=)`; the async generator yields one terminal `Candidate` per fully-explored rollout, and
   `tune_async` benches each via `await _bench_terminal_async` (writes per-kernel `perf` / `lowering` / inventory rows,
   returns the aggregate `PerfStats`), then calls `search.observe(stats, status)`. With `backend=None` the bench is
-  stubbed to `latency_us=1.0` and nothing is persisted, so a backend-less sweep never overwrites tuned rows.
+  stubbed to `latency_us=1.0` and nothing is persisted, so a backend-less sweep never overwrites tuned rows. A terminal
+  still carrying an un-lowered kernel-bearing node (a validation-filtered rewrite) is a `bench_fail` decided before any
+  bench or cache lookup — the bench sums `CudaOp`s only, so without that guard the un-lowered kernel priced at zero and
+  a cached residual kernel's µs could stand in for the whole graph as an `ok` measurement (issue #327).
 - `Run.drive(graph) -> Iterator[(token, Candidate)]` — the exploration engine loop (`tune`). `Run` is the per-run state
   object (`pipeline` + `ctx` + `search` + `db` + `backend` + `dump` + `rejections`): `Pipeline` stays a frozen,
   shareable pass layout while every run-scoped sink lives on the Run, reached through the candidate (`cand.run.dump`,
--- a/emmy/compiler/pipeline/passes/ARCHITECTURE.md
+++ b/emmy/compiler/pipeline/passes/ARCHITECTURE.md
@@ -57,7 +57,11 @@ bound (e.g. a non-`Load` operand — a computed-cone / demoted matmul) is reject
   instead of `lower()`-ing the contraction and pattern-matching the result. A `STAGE` pin follows the same rule: the
   option builders resolve it against the built node ONCE (`_resolve_warp_stage` / `_resolve_scalar_stage` — transport
   eligibility, the slab K-chunk `bk_elems`, the depth clamps) and stamp the resolved `Stage` (or `None`, gmem-direct)
-  on the `TileOp`, so the materializer's one staged driver applies it verbatim, deciding nothing. One staging fact
+  on the `TileOp`, so the materializer's one staged driver applies it verbatim, deciding nothing. Fitting the smem
+  budget is part of resolving: a tile whose single depth-1 slot already exceeds it declines to gmem-direct (the warp
+  slab is codec-sized and cannot shrink; the scalar resolver steps `bk_elems` / depth down first), so every offered
+  stage row materializes within budget — a resolved-but-unfittable row would only die at `validate(ctx)`, leaving an
+  un-lowered `TileOp` in the tune's terminal (issue #327). One staging fact
   is derived at materialization rather than resolved here because it is layout, not eligibility: a TMA slab feeding an
   mma drain is **swizzled** (`_stage.pick_swizzle_atom` picks B32/B64/B128 per operand from the slab's inner row span;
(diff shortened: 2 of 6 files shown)


gnull:
@Slonegg I don't understand the compiler and this change very well, it's heavily vibe-coded. High degree of skepticism is adviced when reviewing this :)

Slonegg (approved):

PR #339, opened 2026-07-09
f16-accumulate mma atom: bypass the consumer-die f32-accumulate half-rate nerf

Slonegg:
## Summary

On the consumer GeForce dies (sm_86/89/120 — RTX 3090/4090/5090 and their workstation SKUs), f16 HMMA with f32 accumulate runs at **half** the f16-accumulate rate. This PR adds the `mma_m16n8k16_f16_f16acc` atom: the mma chain accumulates into packed f16 C fragments at the full HMMA rate, and a periodic `FragmentPromote` folds them into f32 shadow registers (`cvt.f32.f16` + add, rezeroing the f16 partials), bounding the f16 rounding error to one K chunk.

**Promote cadence** (no new codegen knob — the chunk structure is the cadence):
- staged rows: once per bk slab chunk
- gmem-direct: every `_F16ACC_STEPS` (4) atom-K steps + an unconditional final fold (covers symbolic-K tails)
- flash: per streaming KV block, folded in at the `O·α` rescale point — **P@V only**; the score node always stays f32-accumulate (and the P@V f16 partial is overflow-safe: Σp ≤ 1 per row bounds it by max|V|)

**Gating** — precision-trading, off by default:
- `EMMY_F16_MMA_F32_ACC` (precise pin, authoritative on any target: `1` offers everywhere, `0` never)
- `EMMY_FAST_MATH` umbrella (the `-use_fast_math` analogue; also enables `FAST_EXP` via the new `space.precision_pin` precedence: own pin > umbrella > off) — offers the forks on the consumer-die ccs only
- a `TILE` pin naming the atom (or a flash golden's axis-keyed `TILE@<pv_k>` spelling) bypasses the gate — pins stay authoritative, so recorded goldens replay without the env
- the umbrella is a meta gate, so `Knob.cosmetic` is renamed **`Knob.unfeatured`** (must never enter the prior's feature vector); the realized fork is identified by the `TILE` codec's atom token and priced by the existing `MMA_acc_bits` feature

## Measured (RTX 5090)

4096³ f16 GEMM, `d4/tma/ring`, `w2x4/f2x4` (the recorded golden schedule, which ties cuBLAS at 210 TFLOP/s = the f32-acc hardware ceiling):

| bk | f32-acc | f16-acc | speedup |
|----|---------|---------|---------|
| k1 (golden) | 211 TF | 223 TF | 1.06× |
| k2 | 203 TF | 251 TF | 1.24× |
| k4 | 188 TF | 250 TF | 1.33× |
| **k8** | 192 TF | **259 TF** | **1.35×** |

Best f16-acc config is **1.21× over the recorded golden/cuBLAS** and 24% past the f32-accumulate ceiling. f16acc kernels match the f32-acc sibling within 2.4e-4 (K=128) and f16acc flash matches torch SDPA at 4.9e-4.

## Tests

- `test_matmul_coverage.py`: f16acc section — transports (gmem-direct / cp.async ring / TMA) × (static / symbolic-M), each checked against the f32-acc **sibling** (the loose fp16 eager tolerance can't see accumulate drift); symbolic-K tail; the no-GPU enumeration-gate matrix (umbrella × precise pin × cc)
- `test_attention_coverage.py`: f16acc-PV fork enumeration (rows double under FAST_MATH on sm_120, never on sm_90, `TILE@dd` never f16acc); pinned f16acc-PV flash vs torch over gmem-direct + staged streams
- `test_knob.py`: `precision_pin` precedence
- Full suite: 2264 passed, 39 skipped; lint clean


Slonegg:
Added in the follow-up commit: atom naming now follows the `mma_<shape>_<ab_dtype>_<acc_dtype>` convention (compressed PTX/CUTLASS D.A.B.C order) — the f16-accumulate atom is `mma_m16n8k16_f16_f16`, the existing atoms are `mma_m16n8k16_f16_f32` / `mma_m16n8k16_bf16_f32`, and the old acc-unspecified spellings remain permanent parse aliases (old goldens/pins/DB rows canonicalize on parse; `values_equal` canonicalizes TILE so alias pins don't false-flag in the A/B pin gate). In-repo goldens migrated. Full suite: 2268 passed.

Slonegg:
Third commit closes the fast-math golden gaps: `GoldenConfig.fast_math` (regime derived from the recorded knobs — an f16-acc TILE atom or `FAST_EXP: true`), `analytic.evaluate_golden` self-gates so a fast-math golden ranks inside the `F16_MMA_F32_ACC`-pinned enumeration (scoped `Knob.pinned` context manager), `eval golden` splits each shape into standard and `<name> [fm]` rows picked against their own regime's enumeration, and the tune-golden skill documents the fast-math lane (sweep under `EMMY_FAST_MATH=1` — a superset enumeration, so one sweep measures both regimes; fast-math winners recorded beside the standard entry, never replacing it). Full suite green (2272 passed).

Slonegg:
Regression follow-ups from the fm sweep: (1) **fixed** the gate-off deploy starvation — the -O3 rebench band is now per-regime (`mcts.TuningSearch`), so the best standard row reaches -O3 even when a fast-math row owns the global -O1 best; gate-off qkv.h4096 now deploys at 250.8µs / 0.99x vs cuBLAS (was ~260ms on the scalar tier), and the `[fm]` golden row replays gate-off at 240.1µs via its pin. (2) **resolved** the gate_up "codegen drift" — worktree bisect shows the recorded pin benches ~638µs at its own recording commit, so the old 605.5 was a recording-fidelity miss, not a regression; findings + YAML comments corrected. 2273 tests green.

Slonegg:
Analytic refit landed with this sweep's rows (fm goldens reconstructed under the gate): `_W_A_DYN` median golden rank **1063 → 97**, the `[fm]` qkv golden ranks #1 in its gated pool, o_proj.dynM 663→37. The new `D_reduce_ilp` weight immediately steered the cold pick into a region the old weights censored and exposed **two pre-existing masked-ILP codegen bugs** (symbolic-dim rename in strides → nvcc failure; twisted-carrier tail corruption via raw intermediates in the online-softmax rescale, up to 5.7e-3) — both fixed with off-stride regression sweeps at 1e-5. Fork-regret diagnosis in action: price the censored region honestly and you find the broken codegen hiding in it. 2277 tests green.

PR #342, opened 2026-07-10
Deploy fix: refresh op I/O at apply time — the gemma o_proj 16× scalar misdeploy

6erun:
## The bug

On gemma-4-12B layer 0 (RTX 4090, dynamic), the o_proj matmul deployed a scalar `n32x8/f4x14` tile at **3 625 µs — 48% of the layer** — while the tune DB held measured -O3 MMA rows for the same kernel at **225 µs** (finding 1 of `plans/gemma-4-12b-layer0-rtx4090-tune-findings.md`, on #340's branch).

Root cause (established with instrumented compiles): an apply earlier in the same rule batch (the flash fragment splice) swaps a consumed node's op for a **rebuilt instance whose I/O still carries `_seed_io_placeholders`' `Tensor(name, (), F32)` stubs**. `Match.is_alive` snapshots `Node` identity and cannot see an op swap, so a later match applies against placeholder I/O. The all-f16 o_proj terminal then enumerated with both operands reading `(f32, ())`: `_warp_atoms` refused the warp tier, `_demote_mixed_a` refused the mixed rescue, the fork offered **370 scalar-only leaves**, and the #326 evidence join — correctly — matched nothing against the DB's 34 MMA rows, leaving the model to extrapolate.

## The fix

`Candidate.try_rewrite` refreshes each consumed node's op I/O (`populate_io`) against the live graph before invoking the rule — the same contract as the match-time refresh, idempotent when nothing changed.

**Measured (RTX 4080, tune → deploy on the o_proj reproducer):** the fork offers 1 140 warp rows (was 0), the tune measures them, and greedy deploys the measured MMA config: o_proj kernel **3 630 → 396 µs (9.2×)**, reproducer e2e **3 680 → 466 µs (7.9×)**. Expected on the 4090 against the existing tune DB: ~3.4 ms back on the layer (8.07 → ~4.1 ms, 4.4× → ~2.2× behind eager).

## Guardrail

Greedy's `decide` now **warns when a fork's measured DB evidence is disjoint from every offered candidate** — "the tune measured a schedule tier this compile did not offer". That condition is precisely this bug class's signature and was previously silent.

## Tests

- `test_try_rewrite_refreshes_swapped_op_io` (engine unit) — op swapped between match build and apply; **fails without the fix**, passes with it.
- `test_sdpa_consumer_projection_reaches_mma` (e2e) — the sdpa→transpose→reshape→linear composition (symbolic seq, causal): the previously uncovered o_proj path, asserting mma + numerics.
- `test_disjoint_evidence_warns` — the guardrail (and silence on cold signatures).
- Docs: the graph-true-Tensors contract added to `emmy/compiler/pipeline/ARCHITECTURE.md`.

`make test`: 11 failures on this box, **set identical with the fix stashed** (4 dynamic-golden CLI + 7 TMA attention-coverage, pre-existing on sm_89) — zero new failures. `make lint` clean.

Related: #340 (the tune findings report + hd256 flash goldens). Independent — either merges first.


Slonegg (approved):
We need to drop evidence pick logic from greedy completely I think

PR #354, opened 2026-07-13
Swizzle the fused-edge sync compute-fill slabs + vector fill stores: gemma megakernel −29% std / −41% fm

Slonegg:
## What

Extends the #351 slab swizzle to the **fused-edge `sync` compute-fill** — the transport behind the gemma norm→gate⊗up megakernel (`k_linear_mean_reduce` / `k_rms_norm_matmul_reduce`), which #351 explicitly left row-major. Every fill kind now carries the per-operand mode from `_MmaOps.slab_swizzles`:

- the **compute-filled A** and **transposed-B copy** fills through a new `Write.swizzle` (the identical flattened-index XOR the `CpAsyncCopy` fill applies; `_swizzle_prelude` collects `Write` modes too)
- the **canonical-B cp.async** fill through its existing `Operand.swizzle` (was plumbed, never passed)
- the `ldmatrix` drain reads each slab back through its own mode (already per-operand)

The compute fill also now emits **one vector `Write` per 16-byte run** instead of V scalar 2 B stores — those were 8-way store-bank-conflicted at 16 B thread stride, and the downstream vectorizer can't re-merge them (the `(fe·V) % cols` anchor defeats its affine alignment proof).

## Why / measured

The 4090 gemma tune report measured 206M smem bank conflicts on this kernel class; NCU on the 5090 reproduces it at **294.9M ld conflicts / 82.5M LSU instructions** (>3.5 replays per load, SM 25.5%) while the cuBLAS reference has zero. Post-change, on the gemma fused edge (`RMSNorm → 512×3840 @ 3840×30720`, RTX 5090, unfused eager pair = 582 µs):

| config | before | after |
| --- | --: | --: |
| best std (`w2x4/f2x4/k4 d1/sync`) | 1372 µs | **969 µs (−29%)** |
| best fm (f16acc, same tile) | 1372* | **813 µs (−41%)** |
| ld conflicts / SM% | 294.9M / 25.5% | 29.6M / 46.6% |

Bit-exact numerics (max_diff 0.25 = fp16 rounding); `d2+/sync` ring re-measured post-swizzle and still loses (1268 vs 1013) — stays pin-only. Remaining fused-edge gap vs the plain-matmul golden (562/362 µs) is **B re-streaming** (64.6% DRAM; 8 M-tile rows re-read the full B) — `RASTER` doesn't resolve on fused-cone grids yet, which is the natural follow-up.

## Found in flight

1. **A silent fill/drain mismatch class**: `030_stamp_types` reconstructed `Write` without the new field → unswizzled fill under a swizzled drain → a *scrambled A tile whose matmul output PASSED the runner's accuracy check* (mean_diff 52 under tol ≈340 on this shape). Fixed, plus `vectorize_stores` now merges only uniform-swizzle runs and carries the mode. **The accuracy-tolerance heuristic deserves a separate look** — a tolerance that admits mean error of 52 on values of scale ~60 can't catch layout bugs.
2. New regression test `test_fused_sync_fill_slab_swizzle` pins both derived modes (B64/B128) with a structural `emmy_swizzle` assert + an rtol that catches the scramble.

## Validation

`make test`: **2318 passed, 39 skipped** (incl. the 2 new cells); `make lint` clean; NCU before/after; A/B JSONs under the dev box's `_tune/gemma4-goldens-5090/megakernel.*.json`.


Slonegg:
**Follow-up landed (59a07e3e): RASTER now resolves on fused-cone grids** — `_computed_a_rows` crosses its warp rows with the shared `_raster_candidates` (one-line eligibility change: a computed-A kernel's output is the same static 2-D block-tile grid the plain contraction rasterizes). This is the B-re-streaming fix the PR body called out: the flat N-fastest order re-streamed each B stripe once per M-tile row (the 64.6% DRAM residual).

Measured on the gemma fused edge (5090, 3× reproduced, bit-exact): **fm 813 → 750 µs (−8%), std 969 → 804 µs (−17%)**; every grouping (`gm4/gm8/gn2/gn4/gn8`) lands ≈750 — the L2 reuse saturates once anything breaks the flat sweep. **Cumulative for this PR: the megakernel class went 1372 → 750 µs (−45%)**, against the unfused eager pair at ~580 and the plain-matmul fm golden at 362.

The symbolic-M gate is untouched: a dynM fused edge still decides flat `""` and a `gn8` pin on it fails loudly — the masked-tile grouped decode remains the known follow-up (now demonstrably worth ~60 µs on the masked twin). The B128 regression cell additionally pins `RASTER=gn8` and asserts the `_rsub` grouped decode is emitted. Suite: 2320 passed.

Slonegg:
**Lever 1 landed (see commit): compute-fill LSU diet — fm 750 → 710 µs, std 804 → 787 (3× stable, accuracy exact).** `SyncTransport._run_plans` classifies each producer-cone position once by probing the value closure at synthetic coords: run-invariant stmts (stat-row loads) hoist out of the V-cell replication; a stride-1 k-indexed gmem `Load` (proven by `affine_form` comparison, with a V-aligned-anchor guard) merges into one 16-byte vector `Load` per run — the fill previously issued 8 scalar 2 B loads of `x` plus 8 redundant stat re-loads per run, part of the 3.6× LSU inflation vs cuBLAS. The transposed-B copy's strided gather correctly stays per-cell.

**Branch cumulative: the gemma megakernel is 1372 → 710 µs (−48%), 1.21× behind the unfused eager pair.** Next in the ranked plan: hoisting the row statistic to a grid-level prologue (each CTA still recomputes the full-K mean-of-squares — 240× per row across N-stripes), which is the remaining structural recompute tax. Suite: 2320 passed.

Slonegg:
**Non-splitting exhaustion round (see commit).** Three levers tried, one paid:

1. **Tile re-sweep at the current codegen (knob-space, no code): fm 710 → 658–665 µs** — the swizzle/raster/fill fixes moved the tile optimum to `w2x4/f4x4/k4` (taller M tiles became viable once the fill got cheap). The megakernel is now **1.13× behind the unfused eager pair; branch cumulative 1372 → ~660 µs (−52%)**.
2. **Asymmetric B-only prefetch ring (landed, pin-only): refuted by measurement.** `d≥2/sync` now rings only the cp.async B slabs (new `k0_cur` current-chunk handle through `staged_kloop`; A slab + stat rows stay single-buffer) — a strict improvement of the old full-slab ring, bit-exact, but still 897 vs 665 µs at d2: even one extra B slot crosses the smem occupancy quantization (3 → 2 CTAs/SM). The kernel sits exactly on that cliff; **no prefetch slab of any kind is affordable at this tile**. `d1` remains the deployable form.
3. **16-byte stat-prologue reads (prototyped, reverted): refuted** — 660 → 854 µs; the 8-wide unpack chains raise the register high-water mark past an occupancy boundary.

Both negatives are recorded in the `_resolve_sync_stage` / `SyncTransport` docstrings so nobody re-litigates them. With the non-splitting space now genuinely exhausted (tiles, transport, raster, fill vectorization, ring, stat reads — plus WSPEC nvcc-fails and f2/f8 ILP forms spill), the remaining ~80 µs to eager and ~300 µs to the plain-matmul golden are structural: the per-CTA statistic recompute and A re-materialization, i.e. the split-at-the-stat-seam / `PLACE@cone=cut` conversation. Suite: 2320 passed.

Slonegg:
**`PLACE@cone=cut` implemented (b6c4cc88) — and it beats everything.** The pin was a silent near-no-op: its only consumer was the tile recognizer's nodification gate, which could demote an already-fused kernel to its coop Map form but never split it (the kernel boundary was decided upstream by loop fusion). The pin now has its fusion half: `010_merge_loop_ops` refuses to splice a compute-bearing producer into a `_contraction_shaped` consumer under the pin (a multiply of two *distinct* values under a reduce axis + add-Accum — the distinct-deps test keeps the norm's own `x·x` square from reading as a contraction, so its internal assembly still fuses). Default behavior is byte-identical; the guard is pin-gated.

**Measured on the gemma gate_up fused-edge shape (5090, 3×, accuracy exact):**

| form | e2e µs | vs eager (570) |
| --- | --: | --: |
| fused, session-optimized best | ~660 | 0.86× |
| unfused eager pair | ~570 | 1.00× |
| **cut + golden-family config** | **496–503** | **1.14× — emmy wins** |

The split program is `k_rms_norm_reduce` (3.2 µs) + pointwise (4.2) + the matmul on the staged TMA tier (~490) — the plain-gmem-A matmul reaches the golden-class schedules the compute-fill never could. This empirically settles the fusion-profitability question for wide-N fused edges: **at N=30720 the cone fusion is a pessimization**, and the recompute tax we spent the session shaving was structural after all.

Follow-ups this opens: (1) the auto default is still `fuse` — flipping it per-shape (or enumerating cut as a fork sibling) is a behavior change gated on the validation suite + a layer-0 re-tune; (2) the unkeyed-pin leak bit again during the probe (`REDUCE=` on the matmul serialized the norm's reduce — 345 µs); (3) the real gemma edge (gate⊗up + GeGLU) under cut needs the layer-0 re-tune to validate end-to-end. Suite: 2320 passed (one unrelated OOM flake from two suites racing the GPU; passes in isolation).

Slonegg:
**Cone cut relocated to the tile level per the architecture invariant (464c9e50) — the fusion guard is reverted.** `lowering/tile/020_cut_cone` now realizes `PLACE@cone=cut` as a `030_split`-style graph rewrite on the *recognized* kernel: it re-binds the cone via `bind_prologue_contraction` and splits at the A seam into three plain un-mapped `LoopOp`s — per-row statistic (coop reduce), pointwise scale (m×k grid), and the matmul re-lowered with a gmem `Load` of the `__cone` workspace. All three re-enter `010_recognize` on the pass-scan restart (the per-directory rule fixpoint that `030_split`'s `_mapped` finalize deliberately opts out of), so each half gets its own schedule and fork — which is what makes this codec-shaped: cut can become a fork sibling the tune arbitrates, which the fusion-boundary version never could. Termination is structural (a Load-A consumer has no cone to bind; the producer's sweep has no fold Accums). v1 scope: single-fold, one statistic, no lead axes — a multi-fold (gate⊗up) cut needs the copy-A sync-transport extension first or it loses its mma tier.

**Measured (gemma gate_up shape, 5090, 3×, accuracy exact):** the cut consumer matmul reaches the golden-class staged TMA tier at **395–400 µs** and the scale kernel 4.2 µs. Cold-greedy e2e is 741 µs only because the *fresh stat-kernel identity* misranks its REDUCE to serial (345 µs; `b256` pins it to 3.2 — the deploy warning even reports 1165 signature rows with none matching the offered set, the recurring unseeded-identity gap). Correctly ranked, the pieces sum to **~402 µs vs the fused form's 660 and eager's 570**. Suite: 2321 passed.

Discovered while porting, for the record: a *single* nested `for m: [stat…, for k: …]` producer only lifts m onto the grid (1.8 ms serial sweeps) — the statistic must be its own kernel to give the scale its 2-D grid; that's now documented in the rule.

Slonegg:
**Naming + docs follow-up (see commit): the tile divide rules now carry their dimension** — `020_cut_edge` (dataflow, the PLACE codec's memory edge) and `030_split_reduce` (reduce axis, the REDUCE codec's `g<w>` shard). A new "two divide rules" section in `passes/ARCHITECTURE.md` pins the distinction: what each divides (two computations vs one computation's iteration space), the opposite re-entry contracts (cut's pieces re-enter recognition un-mapped and earn fresh forks; split's pieces opt out — the decided row IS the point), the decision altitude (placement vs schedule), and how the shared per-directory fixpoint composes them (a cut consumer can subsequently take a `g<w>k` row and get sharded — cut-then-split on one original kernel). Suite: 2321 passed.

Slonegg:
**Full 5090 golden reproducibility audit on this branch (see commit): 84/84 entries replayed, 82 within ±3% — the branch's codegen changes drift nothing.** Both files (main + gemma), 51 names, live golden-plumbing replays; the 4 known greedy-hang names ran under pinned deploys. Three findings:

1. **Two same-knobs speedups since recording** (3× at zero spread, `emmy_us` refreshed): `matmul.square.512` fp32 10.24 → **8.6 µs** and `matmul.square.512.dynM` std 6.14 → **4.2 µs**.
2. **The `reduce.*` recorded references were 3× stale** (torch.sum eager now 5.4/6.1 vs the recorded 16.38/16.39) — `cublas_us` corrected on all four entries; they stay comfortably golden.
3. The **dynM FLOP-floor false positive** from the gemma report reproduces on all six reduce-tier `.dynM` shapes (softmax/reduce/rms_norm k2048–k8192 + gemma k3840) — every one with live ≡ recorded values. That integrity-check overcount is now clearly systematic and worth its own small fix.

Audit JSONs under `_tune/golden-audit-5090/` on the dev box. Suite/schema green.

Slonegg:
**dynM FLOP-floor false positive fixed (see commit).** Root cause: the intensity gate reconstructed FLOPs from the ShapeKey with a `×hint` for dynamic shapes — correct for `from_matmul` keys (symbolic axes excluded, mirroring the stamps) but a **512× overcount** for the reduce-tier/attention golden keys, whose `free_prod` includes the symbolic axis. Since the join key legitimately differs by kind, the gate now reads a config-side count instead: per-kind `GoldenConfig.flops()` (never-overestimated by contract, and hint-free — a dynamic golden's `M` *is* its benched hint), threaded through `Sample.flops`. The `db.py` twin is deliberately untouched — its extents come from stamps, which always exclude symbolic axes, so its multiplier is correct.

All 8 previously-flagged `.dynM` shapes (softmax/reduce/rms_norm k2048–k8192 + gemma k3840) now replay **clean** at their recorded values, and tests pin both directions — a correct-value dynM replay must not flag, an impossible bench must still fire. With this, the golden audit's integrity signal is trustworthy again: **84/84 entries reproducible, zero flags**. Suite: 2322 passed.

Slonegg:
**Loose accuracy tolerance fixed (see commit) — the last open harness item from this PR's findings.** The fp16 verdict was `max ≤ peak OR mean ≤ peak`: under that OR, any output with max error below peak passed without the mean ever being consulted, which is how the sync-fill scramble (mean_diff 52 under tol 340) slipped through. New three-clause form, each clause with one job: `(max ≤ tol AND mean ≤ 3%·peak) OR mean ≤ 0.5%·peak` — the max clause stays the loose per-cell outlier ceiling, the tightened mean clause fails permutation-class corruption (~15% of peak) regardless of its max, and the 0.5% escape preserves the documented split-K atomic-outlier semantics. A direct `_check_accuracy` regression test pins all three behaviors. **Full suite: 2323 passed with zero tolerance adjustments** — no legitimate fp16 path (including half-atomic split-K) exceeds the 3% mean gate, so the old 1.0·peak width was pure defensive slack. With this, all three harness gaps found during this PR (fill/drain silent field drop, dynM FLOP-floor false positive, loose accuracy tolerance) are closed with regression tests.

Slonegg:
**Branch rebased onto main @ 68d2e676 (#355 prior renames — one trivial prose conflict, suite 2328 green) and the full 4090 golden audit run on it (see commit): 79/79 entries reproduce, zero integrity flags.** Both sm_89 files, 54 names, on the rented box. Highlights:

- **All 6 apparent >5% drifts were re-verified 3×; four dissolved as single-pass transients** (the box had a noisy stretch — same window showed ~15%-slow eager refs). Lesson reconfirmed: never update a golden off one pass.
- **Two real refreshes, both faster, 3× at zero spread**: `pointwise.n4096` 5.0 → 4.6 (now matching its `.dynM` twin) and `pointwise.n16384` 17.4 → 16.3.
- The **dynM FLOP-floor fix holds on sm_89** — all 8 reduce-tier `.dynM` shapes replay clean where they previously would have flagged.
- `gemma4_12b.attention.hd256` needed the pinned-deploy fallback (the known greedy-hang class) and replays exactly at recorded (44.1/40.0).

With the 5090 audit earlier, that's **163/163 golden entries reproducible across both cards on this branch** — the PR's codegen changes drift nothing, and both files are value-current.

PR #356, opened 2026-07-14
Flash causal tile-skip + split-KV REDUCE (the g<n>k arm at fragment residence)

Slonegg:
Two composing optimizations for the warp-flash stream, one per commit.

## 1. Causal tile-skip (`68a43d72`)

The tensor-core flash tier streamed the **full key extent for every CTA** on causal shapes — ~half the mmas were fully-masked identity work. The realizer now derives a **triangular kv bound** off the causal `Select`'s predicate shape (`kv ≤ m` — structural, never a kernel identity): `kv_end = min(seq, (grid_m + 1) · um·fm·atom_m)`, CTA-uniform, bit-identical, hoisted into the `StridedLoop.end` for-init; the staged prefetch clamp re-pins onto the last needed chunk. Static and dynM alike; non-causal/additive-mask untouched.

| shape (hd256 16h fp16 causal, vs torch SDPA) | before | after |
| --- | --: | --: |
| seq 2048 (grid 512 > 170 SMs) | 441 µs (0.55×) | **265 µs (0.90×)** |
| seq 512 static (nt4 = new primary) | 34.6 µs (0.86×) | **32.9 µs (0.93×)** |
| seq 512 w2x1 forms (re-opened) | 97 µs | **35.1 µs** |

`rtx5090_sm120_gemma4.yaml` attention entries refreshed (3× reproduced; clean golden-plumbing replay). The sm_89 attention goldens are now stale-slow — follow-up refresh on a rented box.

## 2. Split-KV REDUCE (`ba57cd52`)

`030_split` could split a twisted carrier only at the scalar per-cell tier (a >10 s hang). A `REDUCE=g<n>k` pin on the flash fork now selects **warp rows with the split stamped on their `Reduction` node**, realized as:

- a **partial keeping full fragment residence**: the kv stream windowed to the CTA's slice via the new `Reduction.offset` (local `[0,B)` fold; `_twist` re-bases the mask columns, gmem/TMA coords, and the causal tile-skip bound — an above-the-diagonal slice runs zero steps), storing the raw `(m,l,O)` state to an f32 `__partial` workspace (O per cell via the fragment store; m/l once per row by the `_t==0` lanes);
- a **finalize** folding partitions through the carrier's `as_state_merge` (the exp-family LSE combine, generated from the spec) + the original projection/layout-aware store.

Kernel finalize only; static block-divisible kv only. Pin-driven (not offered to cold greedy).

**Where it pays:** grid-starved shapes. The 2-head hd256 seq-512 shape: **33.6 → 11.3 µs under `g8k` (3.0×)** — from 0.36× to **1.00× vs torch SDPA** (parity with torch's internally-split flash). The 16-head golden shape does *not* profit (34.9 vs 32.9 unsplit) — and that's diagnostic: halving the per-CTA critical path moved the partial only 9%, so the seq-512 residual is **latency hiding at 1–2 CTAs/SM** (smem/reg occupancy), not causal imbalance. Goldens unchanged; the remaining hd256 parity lever is occupancy (Q residency / register pressure).

## Validation

- New tests: `test_warp_flash_causal_tile_skip` (bound emitted causal-only), `test_warp_flash_split_kv_matches_torch[plain|causal]` (two-kernel split + accuracy vs torch).
- Accuracy validated vs torch on causal, non-causal, GQA, and the f16acc-PV sibling.
- `make test`: 2321 passed, 39 skipped. `make lint`: clean.

### DIFF
--- a/emmy/compiler/ir/stmt/blocks.py
+++ b/emmy/compiler/ir/stmt/blocks.py
@@ -284,6 +284,11 @@ class StridedLoop(Stmt):
     axis Var directly; the strided iteration shape is encoded by the
     loop construct itself rather than via affine indexing in the body.
 
+    ``end`` overrides the ``<`` bound with an arbitrary ``Expr`` over in-scope names
+    (default: the axis extent) — a data-independent early stop such as the causal
+    flash stream's triangular kv bound. The axis extent stays the analysis-visible
+    range; ``end`` must never exceed it.
+
     Reduction detection mirrors ``Loop``: a ``StridedLoop`` is a
     reduce-loop iff its annotated :attr:`role` folds or its body contains an ``Accum``."""
 
@@ -294,6 +299,7 @@ class StridedLoop(Stmt):
     unroll: bool = False
     role: AxisRole = AxisRole.FREE
     carrier: Carrier | None = None
+    end: Expr | None = None
 
     def __post_init__(self) -> None:
         if not isinstance(self.body, Body):
@@ -308,14 +314,22 @@ class StridedLoop(Stmt):
     def with_bodies(self, bodies: tuple[Body, ...]) -> Stmt:
         (body,) = bodies
         return StridedLoop(
-            axis=self.axis, start=self.start, step=self.step, body=body, unroll=self.unroll, role=self.role, carrier=self.carrier
+            axis=self.axis,
+            start=self.start,
+            step=self.step,
+            body=body,
+            unroll=self.unroll,
+            role=self.role,
+            carrier=self.carrier,
+            end=self.end,
         )
 
     def binds_axes(self) -> frozenset[str]:
         return frozenset({self.axis.name})
 
     def exprs(self) -> tuple[Expr, ...]:
-        return (self.start, self.step) if isinstance(self.step, Expr) else (self.start,)
+        out = (self.start, self.step) if isinstance(self.step, Expr) else (self.start,)
+        return (*out, self.end) if self.end is not None else out
 
     @property
     def is_reduce(self) -> bool:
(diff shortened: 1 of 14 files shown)


Slonegg:
## Occupancy investigation (follow-up to the split-KV finding) — hypothesis refuted

The remaining seq-512 hd256 residual (32.9 µs vs torch SDPA 30.7, 0.93×) was attributed to occupancy (1–2 CTAs/SM at 232–255 regs). Four experiments, all on the nt4 primary config:

| lever | occupancy effect | result |
| --- | --- | --: |
| baseline nt4 `d2/tma/ring` | 1 CTA/SM (8%) | **32.9 µs** |
| nt2 `d2/tma/ring` (smaller slabs) | 2 CTAs/SM (17%) | 35.5 µs |
| nt4 `d1/tma` (single buffer → 2 CTAs) | 2 CTAs/SM | 46.3 µs |
| per-step Q reload from gmem (`Q_REG_RESIDENT=0` prototype) | none realized (ptxas keeps the fragments live across the step's mma chain — 252 regs) | 87.5 µs |
| forced 3 CTAs/SM (`__launch_bounds__(128,3)`: 168 regs + 56 B spills) | 3 CTAs/SM (25%) | 74.6 µs |
| split-KV `g2k` (halves per-CTA critical path) | grid 256 | 34.9 µs |

Every occupancy/balance lever loses to the 1-CTA prefetch-ring form. The 232–255 registers are genuinely needed (128 O-accumulator + 64 resident Q + working set), and Q-via-smem can't raise CTAs/SM either — the +32 KB Q tile makes shared memory the binding constraint before registers stop being it. The `Q_REG_RESIDENT` knob prototype is reverted rather than landed (dead flexibility); the refutation is recorded in the golden file's attention notes (`a0b96494`).

**Conclusion:** the last 7% is per-step pipeline depth — cuDNN's kernel overlaps block *i*'s softmax with block *i+1*'s QK inside one warp (cross-step software pipelining), which our emitter doesn't do, and the `f2` two-chain ILP form that approximates it is register-impossible at hd256 (2×128 O-accumulator registers). That's a substantial emitter feature, out of scope here.

PR #357, opened 2026-07-14
Refute the cross-step-pipelining hypothesis for the hd256 residual (golden notes)

Slonegg:
Follow-up to #356 (merged mid-investigation): documentation-only update to the hd256 attention golden notes.

Three hand-edited pipeline variants of the emitted kernel (deferred-PV; S-double-buffered `QK_{i+1}` textually adjacent to `softmax_i`; on d2/d3 × nt2/nt4, one fully spill-free) were benched through the standard `emmy run --bench` accuracy+timing path by injecting the edited cubin under the original source's content-addressed cache key. All passed the accuracy gate; **all timed identical to baseline** (34.2/34.2, 37.0/36.9, 32.9/32.9).

So the 0.93× residual vs torch SDPA at seq 512 is not the per-step instruction schedule — on top of the earlier refutations (occupancy, registers, split-KV, causal balance). The golden notes now say so and point the next investigation at NCU stall attribution instead of another schedule hypothesis.

### DIFF
--- a/emmy/compiler/pipeline/search/goldens/rtx5090_sm120_gemma4.yaml
+++ b/emmy/compiler/pipeline/search/goldens/rtx5090_sm120_gemma4.yaml
@@ -240,12 +240,14 @@ configs:
   # The PRIMARY post-tile-skip: the nt=4 streaming block (32 keys/step, 32 KB slabs) — 32.9, 7% past
   # the nt=2 entry (pre-skip they were parity alternates). The knob frontier ends here — nt=8 fattens
   # to 55 µs, w2x1 forms reach 35.1 (parity with nt=2, no longer the 97 µs hazard), w1x1 38.6,
-  # d3 34.0, WSPEC bands nvcc-fail on this kernel. The 0.93x residual vs torch SDPA is per-step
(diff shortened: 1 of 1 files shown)


Slonegg:
**Retracting — the measurements behind this PR are invalid.**

The cubin-cache injection harness these conclusions relied on never reached the bench path: the deploy/bench compile resolves a different cache key than the injection wrote, so every "hand-edited variant benched identical to baseline" result actually re-benched the *original* kernel. Discovered when a register-count fingerprint (255 vs the edit's 227) exposed the bypass; with a fixed harness (clobber every cache entry, verify by timing movement), hand edits move timing substantially — e.g. the FA-2-style alternating single-slab staging + Q-in-smem prototype takes the nt8 kernel 55.5 → **32.0 µs**, past the nt4-d2 frontier (32.9).

The occupancy/register refutations in the merged #356 notes remain valid (those were knob/launch-bounds experiments, real compiles), as does the NCU instruction-mix analysis. The cross-step-pipelining question is **reopened**, and the 64-key staging work continues on `feature/flash-alt-staging`.

PR #366, opened 2026-07-15
Seed gemma-4 global-layer (hd512) goldens on RTX 5090 + hd512 flash root-cause + gate_up.s2048 std re-tune

Slonegg:
## What

Gemma-4-12B's 8 `full_attention` layers (of 48) use `global_head_dim=512`, **one KV head** (k_proj weight `(512, 3840)`) and `attention_k_eq_v` (no v_proj — V is a V-norm of k_proj). A layer-0-vs-layer-5 kernel dump diff showed their shapes were covered by **no golden file** — and cold greedy on them ranges from a 2× flash rank miss (attention.hd512, 210 µs vs the 113 µs frontier) to an ~800× scalar misdeploy (q_proj_global, 143.9 ms vs cuBLAS 180 µs) and a watchdog **hang** (hd512+s2048's 16.7M-block scalar).

### 1. 22 new golden entries in `rtx5090_sm120_gemma4.yaml`

q/k/o_proj_global, attention.hd512, qknorm.k512 × {static, dynM, s2048} (matmuls in both lanes), manual pinned `--ab` workflow, every winner 3× reproduced at ≤0.7% spread:

- fm lane 1.19–1.66× past cuBLAS on every global projection; split-K is shape-specific (q wants serial, o requires the split, k needs the deep g8k on the k2 atom — the k4-atom g8k is refused on divisibility).
- At s2048 the fm lane keeps a shallow `g2k` on all three (unlike the hd256 tier's q_proj).
- dynM ≈ static parity holds on all matmuls.

### 2. attention.hd512 flash — root-caused, frontier recorded

Static frontier is **0.87×** vs torch SDPA and codegen-bound: the 64×512 f32 O-accumulator alone is 256 regs/thread (255-reg ceiling, 8% occ); only nt=2 realizes staging (nt4's 66.5 KB slot clamps ring depth to 1 — benched, loses); TMA declines wholesale (head_dim 512 > the 256 box-dim hardware cap); the fm P·V sibling loses outright. **At s2048 it reaches SDPA parity (1.00×)** — the register penalty washes out once the grid saturates, and the split-KV advantage inverts. The dynM twin can't take the static winner's split-KV: cross-CTA split of a symbolic reduce axis is not built. Fix directions past 0.87×: a d_v fold (P·V analogue of split-KV) or cross-step software pipelining; a two-box TMA slab encode.

### 3. Slower-than-ref std-row re-sweep (5090)

- **mlp_gate_up.s2048 std replaced**: `w4x2/f4x8/k4` beats the recorded serial by 9.2% (2191 vs 2413; 0.87× → 0.96×).
- o_proj.s2048 and attention.hd256.s2048 std rows confirmed at their frontier (re-measured 1.00×/all alternatives lose) — no churn.

### 4. Diagnostic fix

The warp-flash STAGE-pin decline warning always claimed "static, N-key-block-divisible kv" regardless of the real gate; it now names the geometry and the full gate list (transport / divisibility / TMA caps / smem budget).

Findings report: `plans/gemma4-global-layer-goldens-findings.md` (plans/ pruned to 10: removed the executed gemma-4 layer-0 findings and the superseded analytic-blame baseline).

## Follow-ups

- 4090 twin seeding (rented box; same workflow).
- hd512 d_v fold / TMA two-box encode — the only routes past 0.87× at short seq.
- The explicit-mask (seq>1024) + hd512 combination has no golden coverage yet.

## Testing

- `make test`: 2385 passed, 41 skipped.
- `make lint`: clean.
- All 22 entries load through `GOLDEN_CONFIGS`; `run --bench --golden` replay validated for `q_proj_global` and `attention.hd512` (values within noise, clean integrity flags).


Slonegg:
Added the **RTX 4090 twin** (d2a1d716) on the provided box: 16 entries in `rtx4090_sm89_gemma4.yaml`, same name set (q/k/o_proj_global, attention.hd512, qknorm.k512 × static/dynM/s2048), winners 3× reproduced and replay-validated on-box.

Highlights vs the 5090:
- fm lane 1.29–1.47× on q/o_proj_global; **k_proj_global below parity on both lanes** (fm 0.93×, std 0.85× — the sm_89 kv-proj residual, worse at N=512).
- **Split-KV does NOT invert at s2048 on this card**: attention alt+g2k 1069 µs (0.87×, 0.1%-stable) beats plain alt (8% jitter) and ring — the transport/split interaction is card-specific.
- Greedy on the unseeded hd512 shapes misdeploys 0.14–0.59× but never hangs on sm_89 (the 5090's s2048 pick blows the watchdog).

Slonegg:
**Follow-up work landed on this branch** (76bc56fa, 6390135e), addressing the four open items from the findings:

1. **Symbolic flash split-KV — built** (was "cross-CTA split of a symbolic reduce axis is not built yet"). A symbolic kv stream now slices at the bn-aligned runtime width `B = ceil(S/(cta·bn))·bn` with the slice's absolute end on a new `Reduction.bound`; the realizer stops at `bound − offset` and re-points every kv mask/clamp at the slice end. Fixed on the way: the split partial's state stores were missing the symbolic-M `m_guard` — the tail CTA's overhanging query rows wrote their `(m,l,O)` into the next head's workspace rows. 12/12 accuracy matrix green (staging × split × tails/empty-slice), new `test_warp_flash_split_kv_symbolic_matches_torch`, full attention suite passes.
   **Measured**: dynM hd512 attention **135.2 → 116.6 µs on the 5090** (+13.7%) and **139.6 → 123.7 µs on the 4090** (+11.4%, 0.1%-stable vs the 9%-jitter plain alt); both golden rows updated. At hd256 dynM the split loses (non-starved grid) — row stands.
2. **hd512 d-split geometry**: `w2x2` probe confirms the twisted moveset enumerates `w<um>x1` only — the d_v fold now has a concrete shape (warp-column d-split), deferred to its own branch.
3. **Banded tile-skip**: scoped, deferred — `SdpaOp` carries no window metadata (HF materializes the mask tensor), so the feature needs the trace-side `sliding_window` stamp + `kv_start` derivation + staged-loop start support; design in the findings.
4. **Layer re-baseline**: 5090 layer-0 clean tune running locally; the 4090 twin is crawling on an unauthenticated HF download and lands later.

Slonegg:
**Layer-0 re-baseline (5090, fresh clean tune)** — the e2e gap collapsed from **2.13× behind eager to 1.26×** (emmy 1769 µs vs eager 1401, torch.compile 1221). Ground-truth launch-order decomposition of what remains:

- The fused **RMSNorm→gate⊗up→GeGLU megakernel is still half the layer (801 µs, 49.9%) and still `d1/sync`** — the per-operand cp.async lockout survived #354 (which cut its cost, not its transport). Still fix priority 1.
- **down_proj reached the smem-tiled tier** (`g4k d1/cp`, 24 KB smem — the smem-less lockout is gone): 307.8 µs, 0.93× vs its golden.
- The rest is **prior rank misses on shapes whose goldens exist**: attention deploys a gmem-direct `w1x1` form at 105.6 µs where the recorded golden is 32–34, and q/k/v deploy un-staged `d1/sync` rows ~1.4× off their golden configs. Those ~150 µs alone ≈ eager parity in the std lane; the fm goldens put the layer past eager — the strongest argument yet for the golden→prior refit as the next cheap step.

4090 twin tune is running (its HF download completed); numbers land in a follow-up.

PR #381, opened 2026-07-16
Gemma-4 findings follow-ups: bool-dtype whole-model trace, -O3 eval verdict, venv-serving excludes

Slonegg:
Addresses the actionable issues from `plans/gemma4-12b-postfix-golden-serving-findings.md` (the post-gate-fix 12B reproduction / serving A/B report).

## Finding 4 (P2) — whole-model tune blocked on `unknown dtype 'bool'`: FIXED

- New graph-level `bool` DataType (`emmy/compiler/dtype.py`) + CUDA C spelling, so the explicit-mask
  construction ops in whole-model traces carry the right `Tensor.dtype`.
- Scalar consts consumed by a bool-output op fall back to f32 — a comparison (`x > 0.5`) compares in the
  operand's domain, so inheriting the consumer's `bool` was semantically wrong.
- The mask subgraph's aten spellings now translate to numpy names in the tracer (`gt/lt/ge/le/eq/ne` →
  `greater/...`, `__or__`/`__and__` → `bitwise_or/and`) instead of dying in `ElementwiseImpl`.
- Verified: the finding's repro (`emmy compile google/gemma-4-12B --dynamic ... --ir torch`) now completes,
  and so does `--ir tensor` (decomposition). The mask subgraph traces as `equal(input_ids, c) → bitwise_or`
  bool nodes. A full whole-model **tune** (lowering + GPU) has not been re-attempted. New unit test:
  `test_trace_bool_mask_ops`.

## Finding 6 workflow note — `eval variants` verdict uses the -O1 lane: FIXED

The pick verdict is now computed from the -O3 reservoir latencies when the pick and ≥1 other row carry one
(the line names the lane it used), falling back to the -O1 tune-ranking lane otherwise. On the findings run's
fm gate⊗up leaderboard (`tune-l0-fm.db`) the verdict now reads `pick: rank 37/113, 1.00x of best (-O3 deploy
latency)` instead of the false `2.22x — misses best`.

## Findings 1–2 (P1, computed-A staging lockout) — assessed, not re-implemented

The plan's literal lever ("stage the RAW B slab async while A rides the compute-fill") is **already in-tree**:
the canonical-B slab fill is a vectorized cp.async flying under the compute fill even at d1 (`_sync_operands`,
since #354), and the B-only prefetch ring exists as the pin-only `STAGE=d<n>/sync`, measured slower on exactly
the gate⊗up shape (occupancy quantization 3→2 CTAs/SM). Status notes correcting the plan's root-cause phrasing
are recorded in the plan file; the remaining levers are in-model-form goldens (the tracked seeding session) or
genuinely new enumeration families (TMA B fill / redundant-statistic split-K).

## Housekeeping

- `venv-serving/` (the serving-A/B venv the findings session created) excluded from ruff + gitignored — it was
  failing `make lint` with 208 third-party findings.
- `plans/` pruned to the 10-file cap (dropped the executed July-8 flash/mma next-steps plan).

Findings 3 (fm whole-program capture anomaly — needs an nsys-on-replay diagnosis, `--profile` is per-kernel ncu)
and 5 (hang-prone scalar-corner enumeration, P3) remain open in the plan.

Test plan: `make test` (2452 passed, 35 skipped), `make lint` clean, plus the repro commands above.


Slonegg:
Follow-up commit: the whole-model tune (the finding-4 repro's real goal) surfaced a second wall — the CUDA renderer had no spellings for the mask ops. Added comparisons + logical combines + a `where` ternary to `op_to_expr`; with that the 12B whole-model tune completed its full search (1475 ok / 19 bench_fail, 33 unique kernels, online prior written). The whole-model `--bench` phase (full 12B eager + torch.compile on the 32 GB desktop card) destabilized the box twice and is skipped by decision — details recorded in the findings plan.

PR #384, opened 2026-07-17
Decode-shape goldens (M=16/32) + whole-step decode CUDA graphs by default in generative serving

Slonegg:
The two serving levers from the 12B serving A/B roadmap (`plans/gemma4-12b-postfix-golden-serving-findings.md`: emmy ~100× behind stock vLLM, structural — cold decode twins + ~96 host launches/token).

## 1. Decode-bucket golden family seeded (RTX 5090)

16 new entries in `goldens/rtx5090_sm120_gemma4.yaml`: the M=16/32 static twins of every projection. Seeded via an fm-superset sweep (`EMMY_FAST_MATH=1 emmy tune -c` per shape, shared per-run DB/prior) + per-shape `run --bench --json` A/Bs + manual `--ab` family pins (`_tune/decode-goldens-5090/`).

- **emmy beats cuBLAS on every decode shape** — up to 5.7× (q_proj.m16 9.4 vs 53.2 µs); the MLP giants are at the weight-streaming roofline (~1.0–1.1×), as they should be.
- std ≡ fm pick on every shape (plain f16_f32 atoms — the big-tile f16acc family doesn't pay at small M), so no `[fm]` siblings.
- The sweep's early shapes (tuned before within-sweep prior transfer) picked configs 20–45% off the `w1x8/f2x2/k[24] + g{4,8}k + d2/tma/ring` family that every later shape converged on — four m16 entries re-pinned manually (finding 1 in the report).
- **Cold-deploy verified** (empty DB + no online prior — the serving runner's exact situation): greedy deploys the golden on every shape except one.
- **Finding 2 (compiler follow-up): ShapeKey `free_prod` is aspect-blind.** 32×8192 == 512×512 == 262144, so `q_proj_global.m32` shares a bucket with `k_proj_global.s512` (same K) and the faster-recorded s512 entry silently shadows it — that shape cold-deploys at eager parity (26.5 vs 16.3 µs pinned). Fix: a `free_max` discriminator in ShapeKey (`S_ext_free_max` is already stamped; keys are recomputed per process, never persisted). The two m16/m32 twin collisions are benign — the shadowing entry records the same config. Documented in the YAML section header + findings report.

## 2. Whole-step decode CUDA graphs — the `serve --generate` default

- `emmy serve --generate` now passes `--compilation-config '{"cudagraph_mode": "FULL_DECODE_ONLY", "cudagraph_capture_sizes": [1,…,bucket]}'` instead of `--enforce-eager`. vLLM 0.23 wraps an undecorated OOT model in its `CUDAGraphWrapper` for full decode graphs without torch.compile, so the entire decode step — embed + 48× (pre → RoPE → paged attention → post) + final norm — records as ONE graph; the ~96 per-token host launches and per-call Python/dlpack overhead vanish at replay. Capture sizes are clamped to the decode bucket (an over-bucket batch takes the symbolic fallback whose host numpy hops abort a capture).
- Opt-out: vLLM's own `--enforce-eager` (forwards untouched); forced automatically when `EMMY_GEN_DECODE_BUCKET=0`. A caller-supplied `--compilation-config` wins over the default.
- Enabling change: `_Program.run_device` issues the raw launch sequence (`run_once` — prebuilt buffers, no alloc, no sync) when `torch.cuda.is_current_stream_capturing()` — nested stream capture and graph launch are both illegal in a capturing stream.
- Tests: `test_gen_capture_gpu.py` (a `run_device` captured inside an outer `torch.cuda.graph` replays LIVE — new inputs flow through), plus four `build_serve_cmd` cases (default config / eager opt-out / user-config wins / bucket-off forces eager).
- Remaining validation: a serving-level capture-vs-eager A/B (TinyLlama or the 12B) — next session.

Findings report: `plans/gemma4-decode-goldens-seeding-findings.md`. Test plan: `make test` (2484 passed, 35 skipped), `make lint` clean, cold-deploy golden A/Bs above.


Slonegg:
## Capture-vs-eager serving A/B on the 12B (RTX 5090, local)

Matched config both arms: vLLM 0.23, `EMMY_GEN_DECODE_BUCKET=32`, fp16, `--max-model-len 512 --max-num-batched-tokens 256 --gpu-memory-utilization 0.90`, concurrency 32, 32 prompts, 64 output tokens, seed 0.

| workload | eager TPOT (ms) | capture TPOT (ms) | eager TTFT (s) | capture TTFT (s) |
| --- | --: | --: | --: | --: |
| in 256 / out 64 | 2 938.8 | 2 795.6 | 113.8 | 157.6 |
| in 8 / out 64 (decode-dominated) | 1 925.1 | **1 924.0** | 10.4 | 7.9 |

**The capture path is validated end-to-end**: with the new default, vLLM captured all 6 decode sizes straight through the 96 emmy program replays + paged attention (`Capturing CUDA graphs (decode, FULL): 6/6`, 12 s, +0.04 GiB), and every request in every run completed correctly — the whole-step graphs replay live.

**TPOT is unchanged because the decode step is GPU-bound at ~1.92 s** — there's no host-launch overhead for capture to erase at this kernel speed (eager's async queuing already hides the ~96 dispatches, as `serving/ARCHITECTURE.md` predicted). A compile-only probe of the M=32 twins shows why the step is ~160× over the ~12 ms/token golden floor: the POST twin's bare matmul mains **do** deploy the new goldens exactly (`w1x8/f2x2/k2 g8k d2/tma/ring` on o_proj/down-proj), but the PRE twin's projections are **computed-A fused forms** (input-norm→q/kv cones at `w1x1/f1x4 d1/sync`, gate⊗up at `w2x2/f2x4 d1/sync`) — the same in-model class as the prefill findings 1–2, now at decode M. The bare goldens still bought 1.9× (bucket-32 TPOT 3 639 → 1 924 ms pre/post this PR).

**Sharpened next lever:** in-model-form decode goldens (seed the fused norm→projection / gate⊗up shapes at M=16/32 from the twins' own dumps) or the computed-A enumeration work. Capture is the prerequisite already in place — once the fused kernels drop, the 96-launch overhead would otherwise dominate again.

Full details in `plans/gemma4-decode-goldens-seeding-findings.md` (updated on the branch); logs under `_tune/decode-goldens-5090/serve-ab-*.log`.

PR #386, opened 2026-07-17
Decode-serving kernel readiness: ShapeKey aspect discriminator + computed-A split-K and d2/sync enumeration

Slonegg:
Follow-up to #384's capture-vs-eager A/B, which showed 12B decode GPU-bound at ~1.92 s/step (~160× over the golden-kernel floor) with the gap pinned to the computed-A fused twin forms. This PR closes the kernel-side gaps so the twin re-tune + serving re-A/B measure kernels, not lockouts.

## 1. ShapeKey `free_max` aspect discriminator

`free_prod` alone can't tell 32×8192 from 512×512, so `k_proj_global.s512` silently shadowed `q_proj_global.m32`'s golden at cold deploy (eager parity instead of its 17.5 µs entry — found by spying `_golden_pick`). `S_ext_free_max` is already stamped on every row and keys are recomputed per process, so the fix is the constructor pair + per-kind builders. Normalized to 0 for dynamic/sweep-kind keys — those join classes stay bit-identical (round-trip tests pin every kind, both sides). Cold-deploy verified: the shadowed shape and the existing prefill/decode goldens all deploy exactly.

## 2. Computed-A decode enumeration

- **`d2/sync` (asymmetric B-only prefetch ring) offered unpinned** as a fork sibling — the measured M=512 occupancy loss inverts at decode M (tiny A slab + stat rows), so the depth is now measured per shape instead of hardwired pin-only. Transposed-B cones still clamp to d1.
- **Redundant-statistic split-K for single-channel computed-A cones** — the real unlock: the un-split norm→q/kv cone grids run 32–64 CTAs on a 170-SM card (92–160 µs vs ~6–11 µs bare-golden twins). The contraction K slices across CTAs while the k-invariant stat prologue stays full-row per partition (recomputed redundantly — cheap exactly on the small-free shapes `_SPLITK_MAX_CTAS` admits); the per-cell cone σ-reindexes to absolute k, the Map-wrapper projection folds into the deferred finalize, and the sync stage resolves against the sliced node. Multi-channel (gate⊗up) stays excluded — its product-monoid carrier isn't the 1-component additive state the finalize folds.
- Accuracy-verified vs fp32 at both depths (`test_fused_cone_splitk_matches_reference`; the d2 cell asserts cp.async engages on a canonical-B slab). Cold greedy already picks split rows on the real M=32 PRE twin: **357 → 247 µs before any tuning**.

## Test updates

Six fused-edge tests met the wider fork: serial-`REDUCE` pins where the ONE-kernel fused form is under test (swizzle inspection, masked-M contract), the unpinned fork-integrity test accepts both forms, and the row-structure test now asserts the split/depth families with the transposed-B caveat documented.

Test plan: `make test` (2501 passed, 35 skipped), `make lint` clean, cold-deploy golden A/Bs, PRE/POST twin probes. Next (not in this PR): tune the four M=32 twin graphs (sliding + global layer classes) into a serving DB and re-run the capture A/B.

### DIFF
--- a/emmy/compiler/pipeline/passes/ARCHITECTURE.md
+++ b/emmy/compiler/pipeline/passes/ARCHITECTURE.md
@@ -143,11 +143,17 @@ codec's atom token and priced by the `MMA_acc_bits` feature; f16 only (mma.sync 
 `wspec_moves()` is the fourth level (bare `WSPEC`, option-0 `""` = uniform SIMT) — offered only on a warp row over a
 resolved **TMA** stage without a cross-CTA split, and resolved/thread-budget-gated at materialization
 (`_wspec_workers`; an ineligible spec degrades to uniform). A computed-A (fused-cone) contraction enumerates its own
-warp-only rows (`_schedule._computed_a_rows` — the mandatory resolved `sync` compute-fill stage crossed with the
-shared `RASTER` launch-order candidates (its B stripes re-stream per M-tile row, exactly the grouped order's L2
-reuse — `gn8` measured −8% on the gemma gate_up fused edge, 5090), no scalar /
-gmem-direct / split-K / WSPEC rows; the compute-producer role for the fused edge is the anticipated `RoleKind`
-extension). The **flash-form fork**: a `TWISTED` streaming contraction pair (the flash tree) offers its
+warp-only rows (`_schedule._computed_a_rows` — the mandatory resolved `sync` compute-fill stage at BOTH depths
+(`d1` + the asymmetric B-only prefetch ring `d2` as fork siblings — the M=512 occupancy loss inverts at decode M,
+so the depth is measured per shape), crossed with the shared `RASTER` launch-order candidates (its B stripes
+re-stream per M-tile row, exactly the grouped order's L2 reuse — `gn8` measured −8% on the gemma gate_up fused
+edge, 5090) and — single-channel nodes only — the **redundant-statistic split-K** rows: the contraction K slices
+across CTAs while the k-invariant stat prologue stays full-row in every partition (each recomputes it, cheap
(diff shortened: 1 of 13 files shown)


Slonegg:
## Expensive phase complete: twin tune + serving re-A/B — **emmy 12B decode TPOT beats stock vLLM**

Tuned the four M=32 twin graphs against this PR's enumeration (~65 min, `_tune/decode-twin-readiness/twins.db`): the split-K cones took the PRE twin **357 → 102 µs**; the POST twin (gate⊗up at its weight-bound floor) 339 → 244; globals 84/252.

Serving re-A/B (matched config from the #384 A/B, decode-dominated in-8/out-64, concurrency 32, `EMMY_TUNE_DB`/`EMMY_ONLINE_FILE` → twins DB):

| | TPOT (ms) | output tok/s |
| --- | --: | --: |
| pre-goldens baseline (#384 A/B) | 1 924 | 15.9 |
| **this PR + tuned twins, captured** | **21.33** | **374.7** |
| this PR + tuned twins, eager | 21.44 | 334.5 |
| stock vLLM (reference, original A/B) | 24.9 | — |

**90× on decode TPOT, and under stock vLLM for the first time.** All requests correct in every run. Capture ≈ eager at batch 32 — the ~96-launch host stream still hides behind ~21 ms of GPU step time; capture stays the default and pays as batches shrink or kernels improve further. The remaining serving wall is TTFT (~4–5 s: symbolic prefill + boot compile) — the packed-varlen prefill / program-caching roadmap items.

Full numbers in `plans/gemma4-decode-goldens-seeding-findings.md` on this branch; logs under `_tune/decode-twin-readiness/`.

PR #387, opened 2026-07-17
Device-resident symbolic prefill: run the pre/post programs on-GPU at every width (TTFT lever 1)

Slonegg:
First lever on the TTFT/prefill wall (the remaining serving gap after #384/#386 took decode TPOT to 21.3 ms, under stock's 24.9): prefill and mixed chunked-prefill steps still ran the per-layer **host numpy path** — ~2 `.cpu()`/numpy hops per layer per step across 48 layers (4–5 s TTFT at small inputs; ~2–3 s per mnbt-256 prefill chunk).

- **`_Program.run_device_sym`** — the symbolic programs run device-resident at any step width: grids sized per step via `set_sym_values` (no re-allocation), inputs uploaded into the buffer prefix device-to-device, launches issued on torch's stream, outputs returned as device views at the resolved shapes. No per-T graph capture: chunked-prefill T varies step to step, and at prefill widths the dispatch hides behind GPU work.
- **`_compile_split(capacity=…)`** — the symbolic programs build with a capacity-width feed (vLLM's `max_num_batched_tokens`, passed as `max_tokens` from `EmmyGenModel`), so the device buffers hold any schedulable step; `set_sym_values` raises past capacity.
- **`EmmyGenModel.forward`** routes every covered width through `_forward_device`; the numpy path survives only for the standalone oracle (no capacity) and as the over-capacity fallback.
- GPU test: the device symbolic path is **bit-identical** to the host rebind path on the same programs (pre q/k/v + post at T above the decode bucket).

Test plan: `make test` (2499 passed, 35 skipped), `make lint` clean.

Next (measurement, queued): tune the symbolic twin graphs into the serving DB on top of #386's enumeration, then the serving A/B at the in-256 prefill-heavy workload for the TTFT before/after.

### DIFF
--- a/emmy/serving/ARCHITECTURE.md
+++ b/emmy/serving/ARCHITECTURE.md
@@ -71,10 +71,14 @@ checkpoint, tokenizer, and sentence-transformers pooling config still come from 
   Attention dims are **per layer** (`layer_meta(L)` → head_dim / num_heads / num_kv / scaling) — Gemma-4's global layers
   use a larger `global_head_dim` than its sliding ones, so each layer's `pre`/`post` compiles at its own width. The caller stitches between
   `pre` and `post` (a reference torch SDPA in the Phase-2 host stitch; vLLM paged `Attention` in Phase 3). **I/O:**
-  prefill / `num_tokens > bucket` use the host numpy `rebind` path; the **decode hot path** (`num_tokens ≤ bucket`)
-  is **device-resident** (Phase A — `run_device` / `embed_device` / `forward_layer_*_device` / `final_norm_device`:
-  captured-replay over the static program with torch↔cupy DLPack zero-copy, no host hop; reuses the embedding
-  runner's pattern). This removed the ~40% host/dispatch overhead and ~2×'d served decode. **Decode bucket:** it
+  the plugin runs **device-resident at every width**: the **decode hot path** (`num_tokens ≤ bucket`) rides the
+  captured static twins (`run_device` — captured-replay, torch↔cupy DLPack zero-copy), and **prefill /
+  chunked-prefill steps** (`bucket < num_tokens ≤ prefill_capacity`, capacity = vLLM's `max_num_batched_tokens`,
+  passed as `max_tokens` at runner build) ride the SYMBOLIC programs' `run_device_sym` — grids sized per step via
+  `set_sym_values` over capacity-built buffers, launches issued on torch's stream, no per-T graph capture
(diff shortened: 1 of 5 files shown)


Slonegg:
## Measurement: symbolic-twin tune + serving A/B — TTFT wall down 72×, emmy at 0.68× matched stock e2e

Setup: this branch (rebased on #386), the four **symbolic** twin graphs tuned into the serving DB (~70 min, on top of the static M=32 evidence), capture default, `EMMY_GEN_DECODE_BUCKET=32`, same matched config as every prior A/B (fp16, mml 512, mnbt 256, util 0.90, concurrency 32, seed 0).

| workload | metric | host prefill (pre-#387) | **device prefill (#387)** | matched stock vLLM |
| in-8 / out-64 | Mean TTFT | 4 119 ms | **125 ms** | — |
| | Mean TPOT | 21.33 ms | 21.50 ms | — |
| | output tok/s | 374.7 | **1 381.9** | — |
| in-256 / out-64 | Mean TTFT | 113 777 ms¹ | **1 579 ms** | 1 310 ms |
| | Mean TPOT | 2 938.8 ms¹ | 29.26 ms | 23.63 ms |
| | req/s | 0.10¹ | **7.06** | 10.33 |
| | bench duration | 332 s¹ | **4.54 s** | 3.10 s |

¹ the original #384-era A/B numbers at this workload class (pre-#386 kernels + host prefill).

- **The prefill host-hop wall is gone**: small-prefill TTFT 4.1 s → 125 ms (33×) from this PR's device path alone; the prefill-heavy workload's TTFT 113.8 s → 1.58 s (72×) with the tuned symbolic programs.
- **Decode held**: 21.5 ms TPOT (vs 21.33 pre-branch) — the routing change didn't disturb the captured decode path.
- **e2e standing vs stock at the matched prefill-heavy workload: 0.68× req/s** (7.06 vs 10.33), from 0.01× at the original A/B. Pure decode remains **ahead** of stock (21.5 vs 24.9 ms TPOT).
- Remaining gap is now prefill-step **compute** (TTFT 1.2×, mixed-step TPOT 1.24× behind stock): the symbolic-path kernels at T≈256 chunks — the in-model computed-A forms at prefill M and the packed-varlen attention item — plus mixed steps running uncaptured by design. Integration overhead is no longer the story anywhere.

All 32 requests succeeded in every run. Logs: `_tune/decode-twin-readiness/serve-{prefill-in256,stock-in256,prefill-in8}.log`.

PR #388, opened 2026-07-17
Static prefill-chunk twin (M = mnbt): exact grids on the hot chunked-prefill width

Slonegg:
Leftover prefill-performance item from the TTFT arc (#387 took TTFT 113.8 s → 1.58 s; the residual 1.2× TTFT / 1.24× mixed-step TPOT vs stock is prefill-step compute). One of the two components is the symbolic masked-tile programs running chunks at off-hint widths — this PR mirrors the proven decode-bucket design at the chunk width:

- **A third static pre/post twin per layer at `M = prefill_bucket`** (default = vLLM's `max_num_batched_tokens` — chunked prefill fills steps to it whenever the queue is deep; `EMMY_GEN_PREFILL_BUCKET` overrides, `0` disables). Weights shared through the per-wrapper constant cache, activations through the runner arena, same compile-failure fallback contract as the decode twins.
- **Routing:** `T ≤ decode_bucket` → decode twin (captured), `decode_bucket < T ≤ prefill_bucket` → chunk twin (pad → run → slice), else the symbolic device path from #387.
- GPU test extended to pin both regimes (chunk-twin allclose vs the host reference — different kernels, fp16 rounding; symbolic bit-exact).

Test plan: `make test` green, `make lint` clean. Measurement queued: tune the prefill-bucket twin graphs, then the in-256 serving A/B — numbers will be posted here like #386/#387.

### DIFF
--- a/emmy/config.py
+++ b/emmy/config.py
@@ -51,6 +51,7 @@ GPU_LOCK = "EMMY_GPU_LOCK"
 NCU_CHILD = "EMMY_NCU_CHILD"
 SERVING_STATIC = "EMMY_SERVING_STATIC"
 GEN_DECODE_BUCKET = "EMMY_GEN_DECODE_BUCKET"
+GEN_PREFILL_BUCKET = "EMMY_GEN_PREFILL_BUCKET"
 READABLE = "EMMY_READABLE"
 
 _CACHE_ROOT = Path.home() / ".cache" / "emmy"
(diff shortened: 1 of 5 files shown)


Slonegg:
## Measurement (joint round for #388 + #389, integration branch, tuned twins DB)

Re-tuned the POST decode twins (the multi-channel split now in the space) + the four new M=256 prefill-chunk twins (~110 min), then the serving A/B (matched config, capture default, bucket 32):

| workload | metric | previous round | this round | matched stock |
| in-256/out-64 | req/s | 7.06 | **7.26** (0.70× stock) | 10.33 |
| | TTFT mean / median | 1 579 / 1 094 ms | **1 480 / 974 ms** | 1 310 / 1 303 ms |
| | TPOT | 29.26 ms | 29.21 ms | 23.63 ms |
| in-8/out-64 | TTFT / TPOT | 125 / 21.50 ms | **120.6 / 21.54 ms** | — |

Honest per-PR attribution:

- **#388 (prefill-chunk twin): a modest, real win** — +3% req/s, −11% median TTFT. The #387 symbolic device path + tune had already absorbed most of the masked-tile overhead, so the exact-grid twin's remaining slice was small. Emmy's median TTFT is now **under stock's** (974 vs 1 303 ms); the means differ by tail shape (emmy P99 3.0 s vs stock 1.9 s — queue ordering under chunked prefill).
- **#389 (multi-channel split): correct but a null result on the target shape.** post32 re-benched unchanged (244 µs): gate⊗up at M=32 was never grid-starved (240 CTAs at N=30720) — its 240-vs-145 residual is the fused compute-fill overhead (the large-M computed-A pipeline class), which split-K doesn't address. The change stands on principle (removes an artificial carrier limitation; the split is a legal, measured fork sibling wherever a multi-fold shape IS starved) but claims no TPOT win here.

Remaining known perf items, all now research-class: the large-M computed-A pipeline (the last ~1.24× on mixed-step TPOT and the gate⊗up residual), packed/mixed-step capture, and portability (goldens/evidence beyond this box). Integration overhead is done.

PR #389, opened 2026-07-17
Multi-channel computed-A split-K: the gate/up fused cone splits over K at decode M

Slonegg:
The last fused decode form above its floor: gate⊗up at M=32 runs 240 µs vs its 145 µs bare-golden twin — it was excluded from #386's redundant-statistic split-K because the multi-fold node's carrier claimed only the primary channel's state ("a multi-channel node never rides the split tiers").

Three-surface generalization, each at the principled spot:

1. **`Contraction.loop`** synthesizes the true **N-component identity-family carrier** for multi-fold nodes (one additive state per channel; `carrier.out` stays the primary component and the coop tier remains contraction-blind), so the split tier folds every channel's partials. `_splitk_option` now reads the carrier off this loop for single- and multi-channel alike.
2. **`030_split_reduce._split_contraction`** generalizes to N components with the residual path's `ws[comp, ksplit, *cell]` packing: one write/load/seed per state, the ⊗-combine (SwiGLU) applied **once** at the finalize after the cross-partition sums. The atomic arm refuses multi-component loudly.
3. **The mma store** gains a raw per-channel arm: when the epilogue is exactly one workspace `Write` per accumulator (the split partial's shape), each channel's C fragment stores to its own slice with no combine — the previous assert demanded a combining projection.

Verified: `test_fused_gate_up_splitk_matches_reference` (M=32, `g4k` pinned) matches the fp32 SwiGLU reference; the symbolic gate⊗up tests pin the serial fold where ONE masked kernel is their contract (the split is now a legal sibling there too). `make test` green, `make lint` clean.

Measurement queued with #388's: re-tune the decode + prefill twins against the widened space, then the serving A/B — numbers will be posted here.

### DIFF
--- a/emmy/compiler/ir/tile/ir.py
+++ b/emmy/compiler/ir/tile/ir.py
@@ -384,13 +384,22 @@ class Contraction(Stmt):
         if len(self.folds) == 1:
             return base
         # Each extra fold channel splices its ``b_load → lift → accum`` triple after the primary's,
-        # reusing the shared A value (the carrier stays the primary channel's — a multi-channel node
-        # never rides the coop / split tiers, so the carrier only serves role/axis reads).
+        # reusing the shared A value. The carrier is the N-COMPONENT identity-family state (one
+        # additive component per channel) — the true product-monoid state, so the cross-CTA split
+        # tier folds every channel's partials (the redundant-statistic split of the gate/up edge);
+        # ``carrier.out`` stays the primary component, and the coop tier remains contraction-blind.
+        from emmy.compiler.ir.stmt.algebra import Carrier, State, Twist  # noqa: PLC0415 — algebra imports leaves
+        from emmy.compiler.ir.stmt.carrier import Channel  # noqa: PLC0415
+
         extra: list[Stmt] = []
         for bl, acc in self.folds[1:]:
             lift = Assign(name=f"{acc}__v", op=ElementwiseImpl("multiply"), args=(self.a_name, bl.names[0]))
             extra += [bl, lift, Accum(name=acc, value=lift.name, op=ElementwiseImpl("add"), axes=(self.k_axis.name,))]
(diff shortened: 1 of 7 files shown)


Slonegg:
## Measurement (joint round for #388 + #389, integration branch, tuned twins DB)

Re-tuned the POST decode twins (the multi-channel split now in the space) + the four new M=256 prefill-chunk twins (~110 min), then the serving A/B (matched config, capture default, bucket 32):

| workload | metric | previous round | this round | matched stock |
| in-256/out-64 | req/s | 7.06 | **7.26** (0.70× stock) | 10.33 |
| | TTFT mean / median | 1 579 / 1 094 ms | **1 480 / 974 ms** | 1 310 / 1 303 ms |
| | TPOT | 29.26 ms | 29.21 ms | 23.63 ms |
| in-8/out-64 | TTFT / TPOT | 125 / 21.50 ms | **120.6 / 21.54 ms** | — |

Honest per-PR attribution:

- **#388 (prefill-chunk twin): a modest, real win** — +3% req/s, −11% median TTFT. The #387 symbolic device path + tune had already absorbed most of the masked-tile overhead, so the exact-grid twin's remaining slice was small. Emmy's median TTFT is now **under stock's** (974 vs 1 303 ms); the means differ by tail shape (emmy P99 3.0 s vs stock 1.9 s — queue ordering under chunked prefill).
- **#389 (multi-channel split): correct but a null result on the target shape.** post32 re-benched unchanged (244 µs): gate⊗up at M=32 was never grid-starved (240 CTAs at N=30720) — its 240-vs-145 residual is the fused compute-fill overhead (the large-M computed-A pipeline class), which split-K doesn't address. The change stands on principle (removes an artificial carrier limitation; the split is a legal, measured fork sibling wherever a multi-fold shape IS starved) but claims no TPOT win here.

Remaining known perf items, all now research-class: the large-M computed-A pipeline (the last ~1.24× on mixed-step TPOT and the gate⊗up residual), packed/mixed-step capture, and portability (goldens/evidence beyond this box). Integration overhead is done.

PR #401, opened 2026-07-20
Golden drift gate: in-model audit via weight-free serving twins + model provenance

Slonegg:
## Why

The isolated golden A/B reproduces pinned configs but is blind to the serving graph drifting out from under the
recordings — the gemma-4 cast-splice class reproduced 68/68 pinned while every in-model deploy missed. The in-model
audit existed only as a manual diagnostics script; this PR makes it a CLI mode and a standing CI gate.

## What

- **`model:` provenance in golden YAMLs** — optional file-level header (per-entry override), stored as
  `GoldenConfig.model`, never part of any join key. Both gemma-4 files tagged `google/gemma-4-12B`. Tagging is what
  opts a file's shapes into the in-model audit.
- **Weight-free twin capture** (`emmy/serving/twins.py`) — a trimmed random-init skeleton from `config.json` alone
  (no checkpoint download; a trace never reads a weight value), traced through the same
  `build_attention_split_wrapper` / `trace_split` path serving compiles. Reproduces the real shapes, including the
  global layers' head_dim 512.
- **Audit library** (`search/audit.py`) + a supported `greedy.golden_audit` sink (replaces the diagnostics script's
  monkey-patch): compile with the golden tier as the only evidence, one MATCH / DRIFT / GAP verdict per consulted
  fork. Forces the deployable nvcc regime (under `-Xcicc -O1` the `H_opt` guard silently skips golden consultation),
  isolates the machine-local online prior, targets the golden file's own card via `Context.from_target` — verdicts
  are machine-independent and GPU-less-safe.
- **`emmy eval golden --in-model`** — audits every model-tagged card; uncovered warp-contraction forks flagged as
  MAJOR GAP; non-zero exit on DRIFT / compile failure.
- **CI gate** (`tests/compiler/test_golden_drift_gate.py`, ~80 s, fully offline via a checked-in gemma-4
  `config.json` fixture, `HF_HUB_OFFLINE` forced). Twins re-trace fresh each run, so tracer/transformers drift is
  covered — a transformers bump that changes the forward goes loudly red, as it would change serving. Coverage is a
  **ratchet**: per-card major-gap baseline that may shrink but never silently grow.

## Findings the audit surfaced on day one

1. **5090 coverage holes** (pinned in the baseline): the global-layer prefill-256 projections — one fork offers 37k
   candidate leaves with no golden — and the dynamic fused gate⊗up cone.
2. **All 9 of the 4090's dynM projection goldens have drifted**: their staged `d2/cp` configs no longer realize on
   the symbolic-M enumeration (sm_89). The earlier "DRIFT 0" audit was 5090-only. The 4090 gate param is `xfail`
   with this reason; needs a 4090 re-record or prune session.

## Testing

- `make test`: 2015 passed, 576 skipped (CUDA-gated), 1 xfailed (the documented 4090 drift)
- `make lint`: clean
- CLI verified: `--in-model` audit (5090 DRIFT 0), exit 2 on bad flag combos / untagged model, exit 1 on drift


Slonegg:
Follow-up pushed (rebased on main): the gate is now **strict on both cards** — the heal session closed every DRIFT and GAP.

- **4090**: the 9 dynM drifts were a layout class (staged transports decline transposed B — proven by `pin_unmatched` probe pins); healed with `.lin` gmem-direct siblings, plus full static/rms/pointwise/fused coverage. **5090**: its 19 gaps closed the same way; `d2/tma` declines trans_b too, so the decline is universal.
- Both cards audit **MATCH 101 / DRIFT 0 / GAP 0**; `EXPECTED_GAPS` baselines are empty and enforced, the 4090 xfail is gone.
- Three schema extensions (rms `heads`, pointwise dtype, reduce dynamic key), move-catalog filtering of sweep winners, and `PLACE@cone` cut anchors for the GeGLU edge (fused only wins at decode M=32 on sm_89).
- Two compiler follow-ups surfaced: cold greedy miscompiles the dynamic fused GeGLU on sm_120 (intermittent NaN + a 41 ms misdeploy), and staged transposed-B support is the ~2x serving-matmul perf lever on both cards. Details in `plans/golden-heal-gemma4-4090-5090-findings.md`.

PR #416, opened 2026-07-22
Controlled cross-engine benchmark workloads

gnull:
Before this branch, `emmy bench` could sweep GPUs, concurrency and request sizes, but not produce a *controlled cross-engine comparison* — the kind where a skeptical reader accepts that two engines did identical work:

- The bench client ran with engine-default sampling and unforced output lengths, so two engines could generate different token counts for the "same" workload, and the prompt set wasn't reproducible across runs (no seed).
- Every point was a single run — a 3% win was indistinguishable from noise.
- No output-quality check of any kind: nothing to prove a fast configuration wasn't a subtly broken one.

This branch adds, from the recipe author's perspective:

- **`benchmark.seed` / `temperature` / `ignore_eos`** recipe fields — seeded identical prompt sets, greedy decoding, forced output lengths. Unset fields emit no flag (existing recipes unchanged).
- **`benchmark.repeats`** — reruns the identical client workload N times against the one deployed server; the JSON result's `metrics` becomes the per-field mean, with `metrics_stddev` and per-repeat raws alongside.
- **`scripts/run_lmeval_gate.py`** (+ optional `eval` dependency group) — scores a fixed GSM8K subset against any OpenAI-compatible endpoint via lm-eval, one labeled JSON row per serving config, for comparing configurations within noise and against published scores.
- **Recipes `gemma-4-12B-it` / `-emmy` / `-emmy-fastmath`** — stock vs plugin vs FAST_MATH lanes, identical workload grid (4k/4k headline, input sweep 512→65k with zipped context sizes, batch 4/8), all pinned to the controls above.
- `validate_gemma4_serve.py` defaults to the `-it` checkpoint (the benchmarked one; architecture identical to base).

**Testing.** Per-commit: 163 unit tests green (`tests/benchmark`, `tests/recipe`, `tests/scripts`) plus ruff. Recipes validated through the dry-run CLI (9 variants each, expected commands). Exercised live on rented 5090s: the controlled-workload client flags across three full serving lanes (stock / emmy / emmy FAST_MATH, five workload points each, bare-metal venv), and the docker recipe path end-to-end via `emmy bench --ssh` against a fixed host — including driver pinning (`deploy.driver_version: "580"`: purge, nvidia-open install, reboot, verify) and a seeded 2-repeat run whose JSON result carried the mean/stddev aggregation. Not yet exercised: the lm-eval gate against a live endpoint and the full GPU `make test`; both are queued on the same VM.

**Review questions.** (1) The emmy-lane recipes pin `--enforce-eager` per the original brief, but `emmy serve --generate` now defaults to `FULL_DECODE_ONLY` decode-graph capture — should the recipes follow the serve default instead? (2) Should `cupy-cuda13x` join the `serving` extra? Its absence broke the first bare-metal boot; the Docker image installs it separately.


gnull:
Field notes from exercising the fixed-host path on a fresh vast.ai VM (Ubuntu 22.04, bare KVM) — two pre-existing `provision_remote` bugs, separate from this PR's changes but hit while testing them:

1. **NVIDIA Container Toolkit install fails silently.** The log printed "installing NVIDIA Container Toolkit..." and provisioning continued, but no toolkit package/binary landed on the host (`nvidia-container-cli`, `nvidia-ctk`, `nvidia-container-runtime-hook` all absent) while `/etc/docker/daemon.json` still got the nvidia runtime entry written. Net effect: the deploy fails later with `could not select device driver "nvidia"` and retries skip the install because the config file looks done. The install step needs an rc check plus a post-install probe (e.g. `docker run --rm --gpus all ... nvidia-smi`).

2. **dockerd restart missing after toolkit install.** Even with the toolkit properly installed (done manually), the first GPU deploy fails until `systemctl restart docker` — the daemon must be restarted after the runtime registration for GPU device requests to resolve.

Both reproduced on driver-pinned provisioning (`deploy.driver_version: "580"` — that part worked flawlessly: purge → nvidia-open-580 → reboot → verify). Happy to fix in a follow-up.

gnull:
GPU `make test` ran on the vast.ai 5090 VM (driver 580.173.02, CUDA 12.8 toolkit): **2622 passed, 1 failed, 36 skipped** (6m34s, `-n` parallel).

The one failure is unrelated to this branch (zero diffs under `emmy/compiler` / `tests/compiler`):

`tests/compiler/pipeline/search/test_bench_record.py::test_bench_leaves_keys_by_offer_site` — asserts the pool key (pre-descent site digest) differs from the terminal op's own `S_*` digest, but on this box the greedy descent stamps no extra `S_*` deltas for the compiled matmul, so the digests coincide and `assert leaves[0].op_sig != own_sig` fails. Reproducible in isolation on this machine; looks environment-dependent (cold prior / no tune DB → zero-step descent) rather than flaky. Worth a look from whoever owns bench_record.

gnull:
Golden drift audit (`emmy eval golden --in-model`) ran on the same rented 5090 that produced the article's benchmark numbers, on this branch: **101/101 MATCH on sm_120 (and 101/101 on sm_89), zero DRIFT, zero GAP, zero compile failures**. The recorded goldens deploy exactly as recorded in the serving graphs — the benchmark numbers in the Gemma-4-12B article draft were produced with the tuned kernels, unaffected by the prior-bypass paths that #417 floors.

Slonegg (approved):

PR #417, opened 2026-07-22
Golden floor: a realizable golden decides even prior-less resolves

gnull:
## What could go wrong before

A greedy deploy is documented as resolving each fork through the evidence hierarchy — the card's recorded goldens first, then measured reservoir/DB evidence, then the prior. In the common path that ordering holds, but the golden tier was implemented *inside* the prior-gated branch of `greedy_decide`, so two paths resolved with the tier silently switched off:

- **Prior load failure** (corrupt / unreadable `online.json`, any `load_prior` exception): every fork of every program in a serve boot fell to emission order (option-0), missing every recorded golden — on the gen-runner serving path that is all ~96 per-layer programs of a boot.
- **`Pipeline.run`'s last-resort emission-order resolve** (after the validity-retry budget exhausts): one over-budget tile anywhere in a program downgraded *every other* fork of that program to option-0, dropping their verified goldens too.

Option-0 is exactly the misdeploy class the goldens exist to pin (degenerate w1x1 tiles, 12–29x off the golden in the cold-gemma findings; `plans/gemma4-m64-decode-goldens-findings.md` measured a 2539.5 µs pick where the seeded golden runs 1215).

A third, latent member of the same class: `_fork_shape_key`'s flash sniff read the `TILE@dd` + `TILE@pj` offer pair off **row 0 only**. A flash fork whose first-emitted leaf isn't a warp leaf would mis-key as `kind=""` → GAP → the attention goldens skipped in favor of the model, silently.

## What the change guarantees

Whenever a recorded golden exists for a fork's shape and its knobs still realize on an offered candidate, the deploy pick is at least as good as that golden — on every resolve path, including the two prior-less ones. Specifically:

- The golden tier is consulted independently of any prior: with `prior=None` (load failure or the explicit emission-order fallback), a realizable golden deploys and the override is **logged loudly** (`golden floor holds — deploying the recorded golden realization`). Only golden-less forks fall to emission order, bit-for-bit as before.
- The drift warning (golden keyed to the shape but no candidate realizes it) now also fires on the prior-less paths — previously silent there because the tier was never consulted.
- `Pipeline.run`'s final option-0 resolve now threads `blocked` through, so the floor can never re-pick a tile that already failed `validate(ctx)` — the escape hatch keeps its lowering guarantee.
- The flash offer sniff scans **every** candidate row (same convention as the computed-A cone's sync-STAGE scan), so the attention-golden join no longer depends on planner emission order.

The prior-loaded path is unchanged — its tier ordering already implements the floor (golden decides before reservoir/DB/model).

## Mechanism found

In `greedy_decide.decide`, the first statement after loading the prior was `if the_prior is None: return _first_leaf(fp.options[0])` — tier (1) of the hierarchy was gated behind tier (3)'s availability. The fix hoists a golden consultation (same flatten → blocklist → row-build → `_golden_pick` as the main path, structural leaves filtered) into that branch; no new abstractions, no config flags.

## Testing

- New `tests/compiler/pipeline/search/test_golden_floor.py` (GPU-free, follows `test_golden_evidence_deploy.py` patterns): floor deploys the golden on a `prior=None` resolve with the override warning; drift warns and falls to option-0; a card without goldens stays pure emission order (parity with `test_option0_decide_matches_no_prior_greedy`); the blocklist keeps the floor off failed tiles; the flash sniff is row-order-invariant. All four behavior tests fail on the pre-fix code.
- `tests/compiler/pipeline/` + `tests/compiler/test_golden_drift_gate.py`: 470 passed, 4 skipped; `tests/compiler/pipeline/search/`: 330 passed (one environment-only failure of `test_selected_kernel_set_identical_across_processes` on the NixOS host — missing `libz` in the spawned subprocess — passes with zlib on `LD_LIBRARY_PATH`; unrelated to this change).
- `ruff check` / `format --check` clean on touched files.

**Pending**: e2e validation on a rented 5090 box — full `make test`, plus a serve boot with a deliberately corrupted `EMMY_ONLINE_FILE` to observe the floor's override warnings on the gen-runner path, and a before/after `emmy eval golden --in-model --model google/gemma-4-12B` (expect MATCH counts unchanged: the audit path already resolves with the offline prior loaded).

https://claude.ai/code/session_01KQZG4Z6Rsr8mV577b4vfqG

### DIFF
--- a/emmy/compiler/pipeline/ARCHITECTURE.md
+++ b/emmy/compiler/pipeline/ARCHITECTURE.md
@@ -262,6 +262,10 @@ deploys the std config there. Goldens are
 **consulted, never trained on**: they enter no reservoir, no checkpoint, no dataset (they are the held-out
 acceptance set). Golden µs is deployable-regime truth and never arbitrates a non--O3 compile. A shape match none of
 whose entries realizes against the offer logs a loud enumeration-drift warning and falls through to the tiers below.
+The tier depends on **no prior**: a resolve with no prior at all — a failed `load_prior` (corrupt/unreadable online
+checkpoint) or `Pipeline.run`'s last-resort emission-order resolve — still consults the goldens and deploys a
+realizable one (the **golden floor**, logged loudly when it overrides option-0), so a broken checkpoint can never
+silently cost a fork its verified golden.
 
 ### `FallbackPrior` and the calibration gate
 
@@ -338,7 +342,8 @@ pinned, so the prior would be blind at the `BM/BN` choice. Instead `greedy_decid
 complete leaves (`fork.flatten_leaves` expands branches depth-first; only knob dicts — materialization stays deferred
 to the chosen leaf) and picks the lowest `Prior.mean_scores` over the full `{H_*, S_*, complete-knob-row}` vector in
 one batched `predict`, invariant to the tree's level order. Cold, the `OfflinePrior` ranks (including a positive
-`MMA_tier` warp preference); option-0 (first leaf, emission order) only if `load_prior` returns nothing entirely.
+`MMA_tier` warp preference); if `load_prior` returns nothing entirely the recorded goldens still floor the pick and
+only the golden-less forks fall to option-0 (first leaf, emission order).
 Greedy benches nothing, so it can only *use* a prior, never train one.
 
 **Every deploy pick breaks ties by candidate content, never enumeration order.** The model can score many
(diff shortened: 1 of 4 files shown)


gnull:
@trifonova-n tldr of the slop in the description: this should prevent emmy from failing to use goldens when prior fails.

Slonegg (approved):

gnull:
Post-merge hardware validation on a rented 5090 (gemma-4-12B-it serving boot, `--max-model-len 8448`):

- **Corrupt `EMMY_ONLINE_FILE` alone**: boot healthy, **0 floor engagements** — the offline cold-start fallback quietly covers this case, so the floor correctly stays out of the way. (Useful scoping fact: a single corrupt prior file never reached the no-prior branch.)
- **Corrupt `EMMY_ONLINE_FILE` + `EMMY_OFFLINE_FILE`**: boot healthy, **1800 floor engagements** — every golden-backed fork logged `deploy: node '<op>' resolved WITHOUT a prior (load failure or emission-order fallback), but the golden floor holds — deploying the recorded golden realization (<µs>)` and the server served normally. Pre-fix code deploys emission-order picks for all of these, silently.

Zero errors in either boot. Together with the pre-merge drift audit (101/101 MATCH on this card), this closes the loop: healthy paths were already golden-correct, and the degraded path is now floored.

PR #433, opened 2026-07-27
First-class multi-output nodes — the MIMO graph foundation

Slonegg:
Executes `plans/multi-output-node-foundation.md` (phases 0–2): first-class multi-input, multi-output graph nodes, retiring the two workarounds that stood in for them.

## Phase 0 — representation + indexes (zero behavior change)
- `Node.outputs: tuple[Tensor, ...]` (non-empty, slot 0 primary); `Node.output` is a read-only alias of slot 0, so ~200 existing call sites read unchanged.
- Buffer names are the edge currency: the graph maintains `_producers` (buffer → (node, slot)) and buffer-keyed `_users` in lockstep with every mutation; new APIs `graph.producer(buf)` / `graph.buffer(buf)` / `graph.buffer_users(buf)`; `graph.users(node_id)` keeps node-granular union semantics.
- SSA per buffer (`add_node` rejects duplicate buffer names anywhere); `Graph.validate()` checks index/SSA consistency (tests only, never production compile paths).
- `splice` / `replace_node` / `rename_node` / `remove_orphans` / `topological_order` / `structural_key` operate on buffer edges. Single-output structural digests are byte-identical to the pre-MIMO scheme (golden/tune evidence keys unchanged).
- Producer-lookup sites that keyed `graph.nodes` by buffer strings migrated to the new APIs (tile/cuda lowering, fusion/lifting/stamp, search, dump, splicer, matcher).

## Phase 1 — plural plumbing
- Serialization: nodes write an `outputs` list; `from_dict` dual-reads the old single-`output` form (old dumps stay loadable).
- `BufferSpec` minted per BUFFER with per-buffer roles (`graph.buffer_role`); `LaunchSpec.writes` names every buffer a launch produces — the slab planner's first-write test reads it (old stored plans fall back to `node_id`).
- Default `Backend.run` stores per-buffer values; a multi-output `forward` returns a tuple matched positionally to `node.outputs`; `Op.infer_output_shapes` plural hook wraps the single-shape hook.
- Eager accuracy compares each graph output against the eager tuple positionally instead of collapsing to `out[0]` (a latent correctness gap).

## Phase 2 — first real client + workaround retirement
- `025_sink_row_reduce` emits `T__sq` as output slot 1 of the producer node; **`AuxOutputOp` is deleted** (and the `plan_from_graph` skip with it).
- `single_node_graph` slices mirror all buffers of a synthetic MIMO boundary as slice inputs and list all root buffers as slice outputs.
[...]

PR #441, opened 2026-07-30
serving: fix the spec-decode capture ladder; route MTP verify steps onto covered widths

gnull:
**Status: ready to merge.** The one blocking item (the c=64 bucket) is resolved by keeping that lane at its
existing width; see *The c=64 decision* below.

Under speculative decoding, vLLM rounds every CUDA-graph capture size *up* to a multiple of
`query_len = num_speculative_tokens + 1`, then pads each step to the first captured size at or above its
width — so emmy routes on the **padded** width. `emmy serve` handed vLLM a sparse power-of-two ladder, and at
depth 2 (`query_len` 3) every rung moved up: `16 → 18`, `32 → 33`, each landing one rung *above* the decode
bucket it existed to serve. Every steady-state verify step therefore missed its static decode twin and ran the
symbolic program. The ladder is now built from dense candidates floored to a multiple of `query_len`, and
`gen_runner._warn_symbolic_decode` reports once per width when a decode-shaped step misses the twin — the absent
signal that let this hide.

**Scope:** the published serving image runs `vllm` directly with no `--compilation-config`, so it never calls
`_gen_graph_args` and inherits vLLM's dense default ladder, which is immune. The ladder fix bites bare-metal
`emmy serve` users, not the shipped image.

## What is verified

**The recipe's decode buckets are worth more than the ladder fix.** `serving_mtp_rtx5090`'s MTP lanes picked the
*tightest* tuned width covering the verify step; they now pick a **well-covered** width. Measured through the
recipe path on the pinned published image (`cloudriftai/vllm-emmy-gemma4:0.23.0-58733e02`), one box, one variable:

| 4k/4k c=4 | bucket 16 | bucket 32 | stock+MTP | new ratio |
|---|--:|--:|--:|--:|
| depth 2 | 234.99 | **331.22** (+41%) | 374.6 | 0.63x → **0.88x** |
| depth 3 | 236.39 | **421.88** (+78%) | 459.0 | 0.52x → **0.92x** |

The bucket-16 column is a **same-box control** run for this PR; it reproduces the article's published 235.1 and
236.6 to within 0.1%, so the gain is the bucket alone and not box variance. Per-token latency improves 28% and
44% (TPOT 15.36 → 11.09 and 15.65 → 8.77 ms).

It also restores a depth ladder those cells were missing: stock gains 23% from depth 2 to depth 3 at this
concurrency, the published cells gained 0.6%, these gain 27%.

**Output quality at bucket 32 is clean at every padding ratio the c=4/c=8 lanes run** (bare-metal `emmy serve`,
GSM8K 200 questions, seed 0, strict exact-match):

| bucket | real step width | padding | score | validates |
|--:|--:|--:|--:|---|
| 32 | 24 | 25% | **0.670** | c=8 depth 2 |
| 32 | 16 | 50% | **0.660** | c=4 depth 3 |
| 32 | 12 | 62% | **0.665** | c=4 depth 2 |

Stock controls on the same box, same harness and tokenizer: 0.665 at concurrency 8, 0.640 at concurrency 4.

**Ladder correctness is directly observable**: a depth-2 boot reports capture sizes
`[3, 6, 15, 24, 30, …, 255]` — every rung a multiple of `query_len`, top rung under the bucket.

`make test`: 2471 passed, 0 failed (the run was cut short by an unrelated box shutdown at 86%; CI covers the tail).

## Also in this PR

- **Docs.** `serving/ARCHITECTURE.md` carries the ladder mechanism and its invariant (*for every reachable verify
  width, the first rung at or above it stays at or below the decode bucket*), the coverage-beats-tightness rule for
  choosing a bucket, and the new warning. `commands/ARCHITECTURE.md` states the rule beside the capture-ladder
  default. `docker/vllm-emmy-gemma4/ARCHITECTURE.md` documents a cost discovered here: a run-time serving-shape
  override is an **execution-plan pack miss**, and the boot then re-runs the whole compiler frontend per program
  (~50 min of host CPU at bucket 256 against a 32-warmed image, with cubins already baked) — longer than the
  compose healthcheck's own boot window.
- **The quality gate now covers what benchmarks serve.** `gsm8k_mtp_rtx5090` gated only depth 3 at emmy's default
  bucket, i.e. a configuration no benchmark runs. It now takes a depth and a decode bucket per lane and adds two:
  emmy at bucket 32 — the width the throughput recipe's c=4 and c=8 MTP lanes route onto — plus a stock control at
  the same depth, which is what distinguishes a broken probe from a broken engine.
- **The investigation report** moves from the branch root to `plans/gemma4-mtp-batched-serving-findings.md`.

## The c=64 decision

`EMMY_GEN_DECODE_BUCKET=192 -> 256` was in this PR and is now **reverted**: bucket 256 is faster but **fails the
output-quality gate** — GSM8K 0.0 with empty completions, on a healthy driver, with the stock control at 0.665 on the
same harness, tokenizer and box. The lane keeps 192, which is untuned (its kernels resolve cold, and the same
configuration has measured anywhere from 10.7 to 332.7 tok/s) but *correct*, and matches the width the article's
published number for that cell was measured at.

The recipe header now records why 192 stays despite the rule pointing at 256, so the next reader does not re-derive
the rule and hit the same failure. The GSM8K gate drops its bucket-256 lane with it. `_warn_symbolic_decode` now
leads with the capture ladder and qualifies its bucket advice, since widening to an uncovered width trades a
throughput loss for a correctness risk.

**The underlying bug is not fixed here** and wants its own issue: a width outside the image's warmed execution-plan
pack resolves its kernels cold, and cold picks appear not to be accuracy-gated. The gate lane removed here is a
ready-made regression probe for it.

## Known deviation from the checklist

`plans/` holds 11 files, over the cap of 10 in CLAUDE.md. Left as-is by decision — this change *adds* a findings
report rather than executing a plan, so enforcing the cap would mean deleting somebody else's plan.

## Corrections to this PR's earlier claims

- **"c=64 reaches 1194.8 tok/s at bucket 256" — not reproducible.** Benched directly against a warm docker server
  at that exact configuration: **748.69** tok/s (TPOT 28.6 ms). The 1194.8 figure was bare-metal; the docker path
  does not show it. And the configuration fails quality regardless, so the cell is not publishable either way.
- **"+30–34% on decode-dominated load" is unreproduced.** That is the original measurement of the ladder fix and
  has not been independently repeated. Everything verified above concerns the *bucket*, not the ladder. The
  ladder's *correctness* is established (rung invariant); its *benefit* is not.
- **"The 4k/4k cells are prefill-bound (TTFT 24–26 s)" does not reproduce on the recipe path.** Mean TTFT is
  ~1.1 s in both c=4 cells, and they moved 41–78% on a decode-width change. That reading came from bare-metal runs
  and does not describe the benchmark the article publishes.
- **The mechanism behind the bucket-256 failure is not padding.** An earlier guess was that the static twin's
  padding is not neutral when the real width is far below the bucket. Bucket 32 at 62% padding scores 0.665, which
  refutes that. The better-fitting hypothesis: bucket 256 is a **pack miss**, so its kernels are resolved cold at
  boot — the same nondeterministic path that produced a 31x throughput spread on this configuration — while bucket
  32 is a pack hit with frozen, warmed picks. Note the shipped golden dataset contains entries that are recorded
  *accuracy failures* rather than measurements, so a cold resolve landing on one would produce garbage with no
  padding involved. If that is right, the defect is that cold-resolved picks are not accuracy-gated, and re-warming
  the image at 256 should make it pass. Untested.

## Still flagged for the team, not addressed here

A **~22% regression on the shipped path**: the published image `58733e02` gives 450.6 tok/s at the c=8 depth-2 cell
where an image built from merged main gives 353.2 at identical knobs on the same box. Candidates by surface: #438
(Featurizer v3 — pick propagation on golden-less shapes, i.e. the cold-resolve path) and #433 (MIMO). Consequence
for anyone rebuilding the article's table: every emmy row is reproducible only with the pinned image, and a rebuild
on current main makes those rows *worse* until this is found.

https://claude.ai/code/session_01Jgz6GKUU6QS2JyZHSw4bU4


gnull:
@Slonegg this changes the decode bucket sizes in the MTP measurement.

Slonegg (approved):

PR #442, opened 2026-07-30
Offline prior: declared weight bounds in the fitter — fix the D_pow2_threads cold-deploy poisoning

6erun:
## The incident

A cold (untuned) `emmy serve --generate` of TinyLlama-1.1B on an RTX 4080 ran **54x behind stock vLLM** (TPOT ~216 ms vs 3.76). Per-kernel attribution pinned the whole gap on ONE deploy pick: the fused norm→gate/up contraction shipped a serial `grid=(8,) block=(128,)` schedule — 10.08 ms/layer, ~150x off the ~66 µs DRAM floor. Full forensics, standalone repro, and the twin-tune closure (TPOT 216 → 4.21 ms, 1.12x stock) are recorded in `plans/emmy-serve-performance-findings.md` (included).

## Root cause

`eval online --dataset nodes --blame` on the twin node data put the offline (cold-start) prior's regret squarely on one term: **`D_pow2_threads` at +150k regret-weighted blame** (114.87x sibling regret at the PLACE+REDUCE+STAGE+TILE fork). The #438 featurizer-v3 refit ran the fitter in random-restart mode (`--samples 20000` — the mode the analytic-prior rework plan explicitly recorded as a trap) and shipped the weight at **686** (the #364 disciplined fit had 56).

The deeper problem is structural: the rank objective is FLAT in a feature whose golden-pool variance is tiny, so its magnitude is unidentified upward — a de-poisoned seed climbs 0 → 630 with byte-identical golden metrics. The surplus is invisible to golden rank but catastrophic at fork scoring, where an undecided prefix scores the feature 0.0: any subtree that decided the knob beats every undecided sibling by `exp(w·scale)`.

## Measured safe window

| D_pow2_threads | static golden top1 | twin fork regret |
|---:|---:|---:|
| 0 | 28/436 | 8.3x |
| 56 | 52/436 | 8.3x |
| **112** | **57/436 (= unbounded)** | **8.3x** |
| 224 | 57/436 | 110x |
| 686 (shipped) | 57/436 | 110x |

The objective's entire legitimate gain saturates at 112; everything above only poisons fork steering.

## The fix

- `fit/linear.py::WEIGHT_BOUNDS` — declared per-feature raw-space |weight| caps (the plan's decision-8 "bound damage" pattern applied to the fit), clamped on the **seed** (a poisoned incumbent heals on the next refit) and on every random-sample / coordinate-descent step. One entry: `D_pow2_threads: 112`.
[...]

PR #443, opened 2026-07-30
Serving: boot-time roofline audit — warn when a deployed pick is far off the DRAM floor

6erun:
Stacked on #442 (shares the serve-performance findings file); GitHub will retarget to main when #442 merges.

## What

At every `EmmyGenRunner` boot, event-time each **static twin program** (one layer per attention class; symbolic programs are skipped — at boot they sit at capacity shape, not a serving width) and compare it against its **weight-streaming floor**: every bound constant must be read at least once per forward, so `floor = const_bytes / dram_bw`, with bandwidth self-calibrated by a device-to-device copy (no per-card peak table to maintain). Any program >10x over its floor logs one loud WARNING naming it and pointing at `emmy tune` / the twin-capture flow.

## Why

The 2026-07-29 TinyLlama/RTX 4080 incident (see `plans/emmy-serve-performance-findings.md` on the base branch): a cold deploy served a fused-norm decode kernel ~150x off the DRAM floor — 10 ms/layer, a 54x serving TPOT gap vs stock vLLM — and nothing surfaced it. The server booted clean and was simply slow; root-causing took a per-kernel attribution session. This PR turns that failure class into one boot log line with the correct remediation attached.

## Design notes

- **Conservative by construction**: the weight floor is a true lower bound (compute-bound programs sit above it) and copy bandwidth undershoots peak DRAM bandwidth, so real ratios read LOW — a flagged program is never a healthy one. Programs with floors under 20 µs are skipped (launch overhead dominates; a mispick there is cheap).
- **Advisory only**: the audit never raises; any internal failure drops to a debug log. A boot warning must not become a boot blocker.
- `const_bytes` is deduped by storage (gemma's K=V projection reuse, the shared-upload cache) so the floor never overcounts.

## Validation

- 4 new CPU unit tests (decision thresholds, warn/quiet behavior, never-raises contract).
- Live A/B on the incident shapes (TinyLlama, decode bucket 8, local 4080):
  - tuned twin evidence: `boot audit clean: 4 static program(s) within 10x of the weight floor`
  - cold boot: `L0.post.decode.m1 runs 68x over its weight-streaming floor (~129 us) — … run emmy tune` (a genuine remaining cold-pick outlier in the M=1 gemv tier; the bucket-8 twins no longer fire cold, corroborating #442's prior fix)
- `make test`: 2701 passed, 159 skipped, 0 failed. `make lint` green.


Slonegg (approved):

Slonegg:
The roofline analysis is a general functionality that we may want in the core

6erun:
Rebuilt onto current main (`5173aa65`) after #442 closed in favor of #451 — this PR is now independent of both:

- Single clean commit: the roofline audit + the incident findings file (`plans/emmy-serve-performance-findings.md`), rescued from the closed #442 branch so the forensic record lands with the feature it motivated.
- `plans/` re-pruned to the cap (dit-v1 landed as #431; the 5090 golden-sweep findings are superseded by `golden-parity-rtx5090-findings.md`).
- Validation on the rebuilt base: lint green; `make test` 2814 passed / 4 failed — the 4 failures (`test_attention_coverage.py::test_warp_chain_gqa_dynamic_matches_torch`) reproduce identically on bare `origin/main`, i.e. pre-existing upstream breakage unrelated to this change.

PR #444, opened 2026-07-30
goldens/rtx5090 gemma-4: the tile-M rule at the m64 / m192 / m256 decode widths

gnull:
## Read this first: what this PR is

Three measurements of the c=64 depth-2 cell, which only make sense together:

| measurement | tok/s |
|---|--:|
| published image `cloudriftai/vllm-emmy-gemma4:0.23.0-58733e02`, docker | 332.7 |
| bare-metal from this branch, **after** the four `norm_linear` rows | 310–353 |
| bare-metal from the same commit, **before** them | **4–6** |

The published image sits where the fixed bare-metal build sits, and both are ~60x above the unfixed build. So
**the published image never carried this misdeploy — a fresh build from `main` does.** The image was baked with its
execution-plan pack written at warm time, which froze a working pick for the fused cone before the golden gap could
bite; a build without that pack re-resolves the cone from the goldens every boot, finds nothing at m192, and takes
the scalar config.

> **This is not a throughput improvement over what ships today** — it is roughly parity with the published artifact.
> What it is: the thing that prevents a **~60x collapse, about 330 tok/s down to about 5, the next time anyone bakes
> an image**, because the pack that currently hides the gap is regenerated from the goldens.

"Restores parity" and "required before the next rebake" are very different propositions, and only the second is
true. The same reasoning applies to any width added to a recipe later: the pack masks a missing fused golden right
up until it is regenerated, which is exactly when nobody is looking for a regression.

---

## The rule

**A wide-N streaming edge must read its weight slab from DRAM exactly once, and the `TILE` codec's M extent is what
decides whether it does.** The extent is `WM · FM · atom_m` (16 for `mma_m16n8k16`), and both ways of getting it
wrong cost about the same ~1.5–1.75x:

- **extent < step width** — M splits into blocks that each re-stream the full slab, so B is read twice or three times;
- **extent > step width** — B is read once, but half to three quarters of the mma is padding.

Every std-lane golden this PR touches violated it. `gate_up_cat` at m64 is `linear(A[64,3840], B[30720,3840])` fp16:
its B slab is 236 MB, so at 1.79 TB/s the floor is 132 us and cuBLAS's 154 us is 86% of peak — the kernel is
bandwidth-bound and nothing else matters. Same shape, same run:

| tile M | geometry | measured | vs cuBLAS |
|--:|---|--:|--:|
| 32 | `w1x8/f2x2/k4` | 228.6 us | 1.48x |
| 128 | `w4x2/f2x4/k2` — *the recorded golden* | 228.3 us | 1.48x |
| **64** | `w1x8/f4x1/k8` | **146.7 us** | **0.95x** |

The tile-M=64 family is flat and wide (`w1x8/f4x1`, `w2x4/f2x2`, `w4x2/f1x4`, `w1x8/f4x4` all land 146.7–152.2), so
the effect is the extent, not one lucky geometry. Two secondary knobs come with it: `k >= 4` is required (the same
winner at `k2` falls back to 222.9), and at m256 `gm8` group-M rasterization is part of the winning row — no std
record at that width had it.

**m192 refines the rule into its general form.** No reachable extent equals 192, so some re-read is unavoidable — and
the winner is a 64-row tile (three M-blocks) *with* `gm8`, which schedules the blocks sharing an N-tile together so
their re-reads are L2 hits: **274.8 us with `gm8`, 434.4 us with the identical tile without it.** So the invariant is
"B leaves DRAM once", and rasterization is a second way to buy it when the extent cannot.

**Where it does not apply** — it is a bandwidth argument, so on narrow-N and long-K edges the tile-M family is 3–6%
*slower* and the existing `w1x8/f2x2` records stand: `q_proj` (N=4096, its 31 MB slab is L2-resident), `o_proj`,
`o_proj_global`, `mlp_down` (K=15360), `kv_proj`, `k_proj_global`.

## Measured

Provenance: **RTX 5090, bare-metal** (not docker), branch at `88b26f6e`, `-O3` (the default for `run`), **std lane
only** — the deployed serving lanes do not set `EMMY_FAST_MATH`, so `f16_f16` siblings are not deployable there and
were excluded from every comparison. Boxes: a 30-core box for the sweeps, a 15-core box for the fused-form and
verification runs. Each recorded row reproduced 2x at <1% spread; all rows carry clean integrity flags (realized-vs-
pinned knobs, wrong-answer check vs torch, arithmetic-intensity floor).

**Every row is verified to *deploy*, not merely to pin** — after the edit the default greedy resolve of each shape
lands on the new row and reproduces its latency, and the drift warning is gone. That distinction is load-bearing:
`--ab` reproduces configs the enumeration would never choose, so a pin-only row looks healthy and changes nothing.

| shape | greedy before | greedy after | vs cuBLAS after |
|---|--:|--:|--:|
| `gate_up_cat.m64` | 228.3 us | **146.8** | 1.48x → 0.94x |
| `gate_up_cat.m256` | 475.1 us | **283.6** | 1.74x → 1.04x |
| **`norm_gate_up.m256`** — the *fused* form serving deploys | 483.5 us | **314.4** | 1.71x → 1.11x |
| `qkv_cat.m64` / `q_proj_global.m64` | 35.1 us | 29.2 | ~~1.31x → 1.05x~~ **claim withdrawn — L2 artifact, see below** |

> **Correction (post-review).** `emmy run --bench` replays one weight slab ~100x, so a slab under the 5090's
> 96 MB L2 is timed L2-resident. The decisive test is whether implied bandwidth (slab / µs) exceeds the 1.79 TB/s
> DRAM peak. `qkv_cat.m64` / `q_proj_global.m64` (63 MB slab) implies **2.20 TB/s** — impossible from DRAM — and its
> *before* number implies exactly 1.79, the DRAM floor, so cache-cold the old config may already be optimal. **That
> throughput claim is withdrawn**; the entries stay (realizable + correct). Four narrow-N m192 rows are flagged
> unconfirmed. Every big-slab row stands, including the whole tile-M rule (headline row 1.61 TB/s, 90% of peak) and
> the ~300x fused misdeploy (immune: big slabs, cold picks ~1000x below DRAM bandwidth, and confirmed end-to-end on
> a live server). Full table in the PR comment.

The m256 fused edge improves because it deploys `PLACE@cone: cut`, so its matmul resolves against the plain
`gate_up_cat.m256` shape and a plain-matmul record floors it.

**Four new m192 rows**, all `w1x8/f4x1/k8` + `d2/tma/ring` + `gm8` (`mlp_down` adds `g2a`), each verified to deploy:
`gate_up_cat` 317.3 → 274.8, `mlp_down` 218.7 → 193.1, `qk_global_cat` 98.5 → 89.8, `qkv_cat` 99.1 → 89.3.

**Why m192 matters more than the 1.1x suggests.** After #441 reverted the c=64 lane to bucket 192, that is the width
the c=64 MTP depth-2 cell actually runs — and it had **zero** records of any kind. That is the documented lottery
behind **10.7 vs 332.7 tok/s on the same serving configuration**: with no golden the fused twins fall to a cold
greedy search whose result is not reproducible. These rows do not make m192 fast (they still trail cuBLAS
1.17x–1.61x, and the header says so); they make it *deterministic*. A predictable 1.2x loss replaces an unbounded one.

## Two bugs

**1. The `gate_up_cat.m64` records were all DRIFT — the biggest kernel in the m64 decode step had no golden floor.**
Every compile logged `no offered candidate realizes any of them ... Investigate enumeration drift`, so the shape fell
through to the prior. The entries still *pin* fine, which is why the isolated golden-reproduction check passes and
only `eval golden --in-model` catches it (that audit also reports GAP 86 against the 4090 file for these twins, so
this is not one shape).

**2. An unrealizable golden is a hazard, not dead weight — 323x.** Recording the best fused-m64 config as
`PLACE@cone: cut` **plus** a `TILE` does not realize (no single offered candidate carries both, though the schema
gate accepts the combination), and the fallthrough lands on a per-cell `fuse`/`b128` scalar config: the edge measured
**54.6 ms** against 168.9 us, reproduced on the global-qk edge too (14.8 ms). Reverted; documented in
`pipeline/ARCHITECTURE.md`. The existing uncut record is the only thing standing between that shape and a 323x
misdeploy.

**Underlying both: nothing in the deploy path checks numbers.** `_golden_pick` ranks candidates by recorded
microseconds only; there is no reference comparison in `CudaBackend.compile` or `build_from_plan`. A legal-but-wrong
knob row deploys silently.

**A better-fitting hypothesis for the bucket-256 corruption (UNTESTED).** The received explanation was that the
twin's padding is not neutral at a wide bucket. Padding is *structurally* inert — every reduction in the twins is
over H, K or head_dim, never the token axis, and vLLM only ever receives `[:t]` slices, so no padded lane reaches a
KV write or the sampler. What fits better: the prefill/decode classification is a **pure width test**
(`decode_ok = 0 < t <= decode_bucket`, no consultation of `attn_metadata`), so at `EMMY_GEN_DECODE_BUCKET=256`
*every prompt shorter than 257 tokens is prefilled through the m256 decode twin*. A defective m256 twin therefore
corrupts prefill as well as decode, which fits a flat 0.0 far better than a partial degradation — and at bucket 32
those same prefills went through the well-covered symbolic programs. This PR's m256 measurement supports the "the
m256 resolve is bad" half directly (the cold pick is 1.74x off cuBLAS and 1.63x off the best config). I did **not**
test the hypothesis; it needs a GSM8K run at bucket 256 with these records in place.

## What did NOT work

- **The m2048 prefill width needs nothing.** Sweeping tile M 128/256 × tile N 64/256 × `gm8` over the six biggest
  m2048 edges: all already 0.99–1.08x cuBLAS, best pin buys 1.01–1.05x, and on `o_proj` and `qkv_cat` no pin beat
  greedy at all. **So the 4k/4k c=4/c=8 TTFT gap (24–26 s against stock's 8.7 s) is not prefill matmul quality.**
- **The YAML's m2048 numbers are stale and should not be used for target selection.** It records `gate_up_cat.m2048`
  at 3644.6 us against cuBLAS 2302 (1.58x); live today is 2197 against 2034 (1.08x).
- **The fused m64 edge gains only 1.11x, and it is unrecordable.** `norm_gate_up.m64` deploys *uncut* and is already
  1.03x cuBLAS (168.9 vs 164.6); its best config (cut + the fixed tile, 152.1 us) cannot be expressed in the golden
  format — see bug 2. Measured, not landed.
- **Six of nine m64 matmul edges have no win available**: `q_proj`, `o_proj`, `o_proj_global`, `mlp_down`,
  `mlp_gate_up_split`, `kv_proj`/`k_proj_global` (several already 0.62–0.94x).
- **`norm_qkv.m64` was never broken** — its greedy already picks `w2x4/f2x4/k4`+`g2a`; best pin is inside noise.
- **Split-K and deeper pipelining do not help the fixed tile**: on the m64 winner `g2k` costs 6%, `g2a` 9%,
  `d3/tma/ring` and `d2/cp/ring` are within noise, `gm8` exactly neutral.
- **`mlp_down.m192` at 1.61x cuBLAS is left open** — its long-K shape wants a family this sweep did not explore.

## No article table cell can be updated from this

The per-kernel wins are real and verified; **the end-to-end consequence is unmeasured, and by my own estimate small.**
I did not get a bare-metal serving A/B. The arithmetic, because it is a useful constraint: the serving-form gain at
m64 is ~17–19 us per layer, ~0.9 ms per decode step over 48 layers, against a step time of roughly 50 ms inferred
from the published 1109 tok/s at c=64 — so **1–2%**, not the 22% that cell is behind.

That same arithmetic says where the c=64 gap is *not*: summing the measured m64 kernels gives ~350 us per layer,
~17 ms per step, against an inferred ~50 ms step. Attention, sampling and scheduling are vLLM's and identical between
the engines, so a 12.7 ms per-step deficit cannot be explained by kernels already running at 0.92–1.07x cuBLAS. **The
next investigation at that cell belongs on the runner's per-step overhead, not on kernels.**

## Claims confirmed / refuted

The briefing carried ten hypotheses. Reporting on them is a first-class result, not an aside.

**Claim 2 — "emmy's advantage over cuBLAS is concentrated at narrow widths and gone by M=32" — REFUTED, and it was
the most load-bearing claim.** It came from cross-M medians over *different kernel populations at each M*, against an
eager baseline. That method is invalid and it points at the wrong width. The correct method is free: the goldens carry
both `emmy_us` and `cublas_us` per entry, so restricting to families that appear at more than one width gives an
apples-to-apples sweep:

| M | 1 | 8 | 16 | 32 | 64 | 256 | 512 | 2048 | 4096 |
|---|--:|--:|--:|--:|--:|--:|--:|--:|--:|
| median emmy/cuBLAS | 0.32 | 0.72 | 0.43 | 0.68 | **0.96** | 0.77 | 0.76 | 0.70 | 0.75 |
| families winning | 9/9 | 20/21 | 8/8 | 24/27 | **15/22** | 24/28 | 19/24 | 19/20 | 15/19 |

There is no cliff at M=32 and no loss of advantage at large M. The one weak width is **M=64** — precisely the width
nobody had looked at — and the cause is the tile-M rule, not a regime change.

**Claim 6 — "the published image freezes kernel choices, so tuning may have no effect through docker" — TRUE but far
narrower than stated; the docker `ARCHITECTURE.md` is misleading.** The pack key is
`{kind, model_type, config_sha, dtype, decode_bucket, max_tokens, prefill_bucket}`, compared with a whole-dict `!=`
and hashed into the pack directory name — so a bucket override is a **whole-pack miss**, not a decode-programs miss,
and on a miss the wheel-shipped goldens *are* consulted. Only the c=1 lanes (bucket 32, no prefill override) hit the
pack. The c=4/c=8 MTP lanes miss because `EMMY_GEN_PREFILL_BUCKET=2048` differs from the warmed 4096; the c=64 lanes
miss on the decode bucket. **Kernel tuning reaches every batched cell in the table.**

**Claim 5 — right about attention, wrong about the boundary.** RoPE is **not** emmy's — the split wrapper returns
un-rotated q/k and vLLM's fused `rotary_embedding` CustomOp runs between the `pre` and `post` halves. The final norm
is a plain torch module; the embedding is a torch gather. The important half holds: there is **no** query-length
branch anywhere in the runner, so an MTP verify step is a byte-identical invocation to a plain decode step of the
same total token count.

**Claim 8 — coverage claim correct, lead not worth taking.** m16 does carry only unfused projections, but no lane
lands on m16 (c=4 moved to bucket 32 in #441), and the padding penalty there is smaller than a new tier repays. The
productive version is what this PR does instead: the widths that *are* deployed (64, 192, 256) had records that
violated the tile rule or no records at all.

**Claim 3 — "the c=8 MTP deficit is not a decode-kernel tuning gap" — CONFIRMED, structurally.** m32 already
satisfies the rule (its `gate_up_cat` record is `w1x8/f2x2/k8`, extent 32 for a 32-row step) and measures 0.97x
cuBLAS. No m32 win exists.

**Claim 7 — mechanism refuted, replacement above.** **Claims 1 and 4** — consistent with everything observed, not
re-measured. **Claim 10** — not reproduced; a greedy-pick integrity scan over 10 of the m64/m256 golden shapes
produced no flagged rows, so those four recorded failures are not reproducing as deploy-time wrong answers on those
shapes today. The remaining 49 shapes are unchecked.

## Checklist

- `ruff check` / `ruff format --check`: **clean** (605 files).
- `make test` on an RTX 5090 box: **2819 passed, 4 failed, 37 skipped**. All 4 failures are **pre-existing on `main`**
  — verified by re-running each individually against the unmodified `origin/main` goldens on the same box, where all
  4 fail identically:
  `test_dynamic_shapes.py::test_qwen_batched_dynamic_matches_eager_b32` (a genuine `CUDA_ERROR_ILLEGAL_ADDRESS` —
  worth someone's attention independently), `test_bench_block.py::test_bench_dry_run_tinyllama_block`,
  `test_gen_pack_gpu.py::test_gen_pack_second_boot_hits_and_matches`,
  `test_gen_runner_gpu.py::test_gen_runner_device_path_matches_host`.
- The suite caught a real invariant break from my own first pass: raising the std row 1.55x left the m64 fast-math
  siblings recorded *slower* than std, which `test_fast_math_rows_never_record_slower_than_std_siblings` forbids.
  Re-tuning the fm lane lands on the same rule (`w4x2/f1x4/k8`, **145.6 us**, a hair under std's 146.7) and the
  superseded `w1x8/f2x2/k2` fm row is dropped. The two accumulate dtypes tying within 1% is itself confirmation of
  the premise the rule rests on — the kernel is DRAM-bound, so the f16-accumulate mma rate buys nothing here.
- **Accuracy on widths outside the image's warmed set — GATED, PASSES.** m64/m192/m256 are all outside the image's
  warm (32), and m192 is live in `serving_mtp_rtx5090`'s c=64 lane, so it was gated before proposing merge. GSM8K,
  200 examples, seed 0, lm-eval 0.4.12, bare-metal RTX 5090, `EMMY_GEN_DECODE_BUCKET=192`, MTP depth 2 with the
  official drafter, BOS-patched tokenizer, stock vLLM as the control **at each concurrency**:

  | lane | concurrency | verify step | exact_match (strict) |
  |---|--:|--:|--:|
  | emmy, bucket 192 | 8 | 24 tok (padded into the 192 twin) | **0.6750** ± 0.0332 |
  | emmy, bucket 192 | 64 | 192 tok (the bucket exactly) | **0.6650** ± 0.0335 |
  | stock vLLM | 8 | — | 0.6900 ± 0.0328 |
  | stock vLLM | 64 | — | 0.6550 ± 0.0337 |

  At ±0.033 stderr the whole spread (0.655–0.690) is about one standard error — statistically indistinguishable.
  Emmy is 0.015 under stock at concurrency 8 and 0.010 **over** it at 64. A numerics break on this harness reads
  ~0.27 (the BOS trap) or 0.0 (the bucket-256 failure); nothing is near either. Two things the controls earned:
  `flexible-extract` is low on **both** engines (0.345–0.370), so it is a harness extraction artifact, not an emmy
  defect; and stock itself scores 0.655 at concurrency 64, **below** the band established at concurrency 8 — the band
  drifts with concurrency, so a lane must be read against a control at its own concurrency. No golden numerics
  change either way: the accumulate dtype is untouched and every row passed the wrong-answer check against torch.

- `ARCHITECTURE.md` updated (`emmy/compiler/pipeline/`). No STYLE/README/CLAUDE changes needed — no new flag, recipe
  field, command or convention. Findings report at `plans/gemma4-m64-m256-tile-tuning-findings.md`.

https://claude.ai/code/session_01Jgz6GKUU6QS2JyZHSw4bU4

### DIFF
--- a/emmy/compiler/pipeline/ARCHITECTURE.md
+++ b/emmy/compiler/pipeline/ARCHITECTURE.md
@@ -263,6 +263,13 @@ deploys the std config there. Goldens are
 **consulted, never trained on**: they enter no reservoir, no checkpoint, no dataset (they are the held-out
 acceptance set). Golden µs is deployable-regime truth and never arbitrates a non--O3 compile. A shape match none of
 whose entries realizes against the offer logs a loud enumeration-drift warning and falls through to the tiers below.
+**That fallthrough is a hazard, not a graceful degradation**, which makes an unrealizable entry worse than an absent
+one: on the gemma-4 m64 gate/up cone the tiers below land on a per-cell `fuse`/`b128` scalar config and the edge
+measures 54.6 ms against the realizing entry's 169 µs — 323×. Two consequences when recording a fused row. Pinning
+`PLACE@cone` **together with** a `TILE` never realizes — no single offered candidate carries both — so record one or
+the other; the schema gate accepts a combination the offer then rejects. And a row must be verified to **deploy**,
+not merely to pin: `--ab` reproduces configs the enumeration would never choose, so a pin-only row looks healthy in
+the isolated golden-reproduction check and only the in-model drift audit (`eval golden --in-model`) catches it.
 The tier depends on **no prior**: a resolve with no prior at all — a failed `load_prior` (corrupt/unreadable online
 checkpoint) or `Pipeline.run`'s last-resort emission-order resolve — still consults the goldens and deploys a
 realizable one (the **golden floor**, logged loudly when it overrides option-0), so a broken checkpoint can never
--- a/emmy/compiler/pipeline/search/goldens/rtx5090_sm120_gemma4.yaml
+++ b/emmy/compiler/pipeline/search/goldens/rtx5090_sm120_gemma4.yaml
@@ -1325,9 +1325,9 @@ configs:
     K: 3840
     dtype: 'fp16'
     trans_b: true
-    knobs: {TILE: 'a:mma_m16n8k16_f16_f32/w1x8/f2x2/k2', REDUCE: 'g8k', RASTER: '', STAGE: 'd2/tma/ring', WSPEC: ''}
-    emmy_us: 33.9
-    cublas_us: 31.5
+    knobs: {TILE: 'a:mma_m16n8k16_f16_f32/w2x4/f2x4/k4', REDUCE: 'g2a', RASTER: '', STAGE: 'd2/tma/ring', WSPEC: ''}
+    emmy_us: 28.6
+    cublas_us: 26.7
   - kernel: matmul
     name: gemma4_12b.k_proj_global.m64.lin
     M: 64
@@ -3155,11 +3155,22 @@ configs:
     emmy_us: 87.3
     cublas_us: 91.7
   # gate_up_cat m64 .lin: the missing layout twin — the cold pick fell to the transposed b<n>t
-  # band on the .lin operand (16.9 ms, the 2026-07-25 per-kernel table's 109x case). Both lanes
-  # trail eager at this width (the fused norm_gate_up.m64 form owns serving at 169.4); the rows
-  # pin the isolated/cut-re-entry deploy to sanity. NOTE the m1 .lin twins are deliberately NOT
-  # recorded: their b64 optima (12.6 us, L2-warm) would out-rank the .t rows on the layout-blind
-  # key and poison the serving (.t) deploy — the interleaved-row removal lesson.
+  # band on the .lin operand (16.9 ms, the 2026-07-25 per-kernel table's 109x case). NOTE the m1
+  # .lin twins are deliberately NOT recorded: their b64 optima (12.6 us, L2-warm) would out-rank
+  # the .t rows on the layout-blind key and poison the serving (.t) deploy — the interleaved-row
+  # removal lesson.
+  #
+  # THE TILE-M RULE AT THIS WIDTH. A wide-N streaming edge is DRAM-bound at M=64: the whole B slab
+  # (30720x3840 fp16 = 236 MB) has to be read exactly once, so the output tile's M extent must be
+  # exactly 64. A 32-row tile splits M into two blocks that each stream the full slab (B read
+  # twice); a 128- or 256-row tile reads it once but wastes half to three quarters of the mma. Both
+  # mistakes cost the same ~1.55x — the std lane measures 228 us at w4x2/f2x4 (tile M 128) AND at
+  # w1x8/f2x2 (tile M 32), against 147 us for every tile-M-64 geometry (w1x8/f4x1, w2x4/f2x2,
+  # w4x2/f1x4, w1x8/f4x4 all land 146.7-152.2). k>=4 is required with it: the same winner at k2
+  # falls back to 222.9. 146.7 us is 1.61 TB/s = 90% of the card's DRAM peak, and 0.95x cuBLAS.
+  # The rule generalizes to the other wide-N m64 edges (qkv_cat / q_proj_global below, 35.1 -> 28.6)
+  # and NOT to the narrow-N or long-K ones — q_proj (N=4096, L2-resident) and mlp_down (K=15360,
+  # K-bound) are 3-6% SLOWER on the tile-M-64 family, so their w1x8/f2x2 records stand.
   - kernel: matmul
     name: gemma4_12b.gate_up_cat.m64.lin
     M: 64
@@ -3167,9 +3178,9 @@ configs:
     K: 3840
     dtype: 'fp16'
     trans_b: true
-    knobs: {TILE: 'a:mma_m16n8k16_f16_f32/w4x2/f2x4/k2', REDUCE: '', RASTER: '', STAGE: 'd2/tma/ring', WSPEC: ''}
-    emmy_us: 233.6
-    cublas_us: 156.0
+    knobs: {TILE: 'a:mma_m16n8k16_f16_f32/w1x8/f4x1/k8', REDUCE: '', RASTER: '', STAGE: 'd2/tma/ring', WSPEC: ''}
+    emmy_us: 146.7
+    cublas_us: 154.0
   - kernel: matmul
     name: gemma4_12b.gate_up_cat.m64.lin
     M: 64
@@ -3187,9 +3198,9 @@ configs:
     K: 3840
     dtype: 'fp16'
     trans_b: true
-    knobs: {TILE: 'a:mma_m16n8k16_f16_f32/w1x8/f2x2/k2', REDUCE: 'g8k', RASTER: '', STAGE: 'd2/tma/ring', WSPEC: ''}
-    emmy_us: 34.0
-    cublas_us: 29.6
+    knobs: {TILE: 'a:mma_m16n8k16_f16_f32/w2x4/f2x4/k4', REDUCE: 'g2a', RASTER: '', STAGE: 'd2/tma/ring', WSPEC: ''}
+    emmy_us: 28.6
+    cublas_us: 26.7
   - kernel: matmul
     name: gemma4_12b.qk_global_cat.m64.lin
     M: 64
@@ -3512,9 +3523,14 @@ configs:
     K: 3840
     dtype: 'fp16'
     trans_b: true
-    knobs: {TILE: 'a:mma_m16n8k16_f16_f32/w4x2/f2x4/k2', REDUCE: 'g4k', RASTER: '', STAGE: 'd2/tma/ring', WSPEC: ''}
-    emmy_us: 481.4
-    cublas_us: 292.0
+    # The std lane's tile-M rule at M=256: `k4` and `gm8` group-M rasterization (L2 reuse across the
+    # N sweep) are both required, and the cross-CTA split is not. w4x2/f2x4 at k2 with a g4k split —
+    # the pre-2026-07-30 record — costs 1.74x cuBLAS; the same tile at k4 with gm8 and no split costs
+    # 1.07x. This row also floors the FUSED norm_gate_up.m256 edge, which deploys `PLACE@cone: cut`
+    # and so resolves its matmul against this shape: that path measures 483.5 -> 296.4 us.
+    knobs: {TILE: 'a:mma_m16n8k16_f16_f32/w4x2/f2x4/k4', REDUCE: '', RASTER: 'gm8', STAGE: 'd2/tma/ring', WSPEC: ''}
+    emmy_us: 291.8
+    cublas_us: 273.7
   - kernel: matmul
     name: gemma4_12b.gate_up_cat.m256.lin
     M: 256
--- a/emmy/compiler/pipeline/search/goldens/rtx5090_sm120_gemma4.yaml
+++ b/emmy/compiler/pipeline/search/goldens/rtx5090_sm120_gemma4.yaml
@@ -4166,3 +4166,59 @@ configs:
     knobs: {TILE: 'a:mma_m16n8k16_f16_f32/w1x8/f4x1/k8', REDUCE: 'g2a', RASTER: 'gm8', STAGE: 'd2/tma/ring', WSPEC: ''}
     emmy_us: 193.1
     cublas_us: 120.0
+  # The FUSED (computed-A) norm->linear cones at this width. These are the rows that matter most
+  # here: with no fused golden the cone cold-resolves to the per-cell `fuse`/`b128` scalar config,
+  # which is not a mild loss but a ~300x one — measured on-card at m192, std lane, against torch's
+  # unfused norm-then-matmul reference:
+  #
+  #   norm_gate_up   132,633 us -> 429.3   (309x)      norm_qkv        35,579 us -> 116.2   (306x)
+  #   norm_qk_global  37,965 us -> 118.2   (321x)      mlp_down_fused  65,067 us -> 265.5   (245x)
+  #
+  # Summed over 48 layers that is ~11.2 s of kernel time per decode step before, ~39 ms after. It is
+  # the mechanism behind the c=64 lane's 10.7 tok/s bare-metal reading, and it is deterministic —
+  # not the run-to-run lottery it was taken for. A computed-A contraction can only ride the sync
+  # compute-fill transports, so the realizable options are `d1/sync` / `d2/sync` and never the TMA
+  # ring; within them all four cones pick the SAME geometry the m64 fused rows already carry
+  # (w1x8/f2x2/k4 + d1/sync), so the fused family is far less width-sensitive than the plain matmuls
+  # above. Recording `PLACE@cone: cut` beside a TILE here would NOT work — that combination never
+  # realizes, and an unrealizable fused row is what produces the scalar config in the first place.
+  - kernel: norm_linear
+    name: gemma4_12b.norm_gate_up.m192.lin
+    M: 192
+    H: 3840
+    N: 30720
+    dtype: 'fp16'
+    trans_b: true
+    knobs: {TILE: 'a:mma_m16n8k16_f16_f32/w1x8/f2x2/k4', REDUCE: '', RASTER: '', STAGE: 'd1/sync', WSPEC: ''}
+    emmy_us: 429.3
+    cublas_us: 266.0
+  - kernel: norm_linear
+    name: gemma4_12b.norm_qkv.m192.lin
+    M: 192
+    H: 3840
+    N: 8192
+    dtype: 'fp16'
+    trans_b: true
+    knobs: {TILE: 'a:mma_m16n8k16_f16_f32/w1x8/f2x2/k4', REDUCE: '', RASTER: '', STAGE: 'd1/sync', WSPEC: ''}
+    emmy_us: 116.2
+    cublas_us: 100.0
+  - kernel: norm_linear
+    name: gemma4_12b.norm_qk_global.m192.lin
+    M: 192
+    H: 3840
+    N: 8704
+    dtype: 'fp16'
+    trans_b: true
+    knobs: {TILE: 'a:mma_m16n8k16_f16_f32/w1x8/f2x2/k4', REDUCE: '', RASTER: '', STAGE: 'd1/sync', WSPEC: ''}
+    emmy_us: 118.2
+    cublas_us: 101.0
+  - kernel: norm_linear
+    name: gemma4_12b.mlp_down_fused.m192.lin
+    M: 192
+    H: 15360
+    N: 3840
+    dtype: 'fp16'
+    trans_b: true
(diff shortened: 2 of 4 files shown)


gnull:
## L2-residency artifact: assessment of every row in this PR

The `--bench` harness replays one weight slab ~100 times, so a slab under the 5090's 96 MB L2 is timed L2-resident
while a live decode step streams ~21.8 GB and is not. Thanks for the catch — it is my method too.

**Slab size alone is only half the test.** Fitting in L2 is *necessary but not sufficient* for a number to be
inflated: a kernel that fits in L2 but runs far below DRAM peak is latency- or compute-bound, and L2 residency has
little room to help it. The decisive test is whether the **implied bandwidth** (slab ÷ measured µs) exceeds what DRAM
can physically deliver (1.79 TB/s). Applying both:

| row | slab | implied BW | verdict |
|---|--:|--:|---|
| `gate_up_cat` m64 / m192 / m256 | 236 MB | 1.61 / 0.86 / 0.81 TB/s | **stands** — exceeds L2 |
| `mlp_down.m192` | 118 MB | 0.61 TB/s | **stands** — exceeds L2 |
| `norm_gate_up.m192` (fused) | 236 MB | 0.55 TB/s | **stands** — exceeds L2 |
| `mlp_down_fused.m192` (fused) | 118 MB | 0.44 TB/s | **stands** — exceeds L2 |
| `qkv_cat.m64` / `q_proj_global.m64` | 63 MB | **2.20 TB/s** | **COMPROMISED** — above DRAM peak |
| `qkv_cat.m192`, `qk_global_cat.m192` | 63 / 67 MB | 0.70 / 0.74 TB/s | unconfirmed |
| `norm_qkv.m192`, `norm_qk_global.m192` (fused) | 63 / 67 MB | 0.54 / 0.57 TB/s | unconfirmed |

**One shape is proven compromised: `qkv_cat.m64` / `q_proj_global.m64`.** 28.6 µs for a 62.9 MB slab implies
2.20 TB/s, which DRAM cannot supply. Worse for my claim: the *before* number (35.1 µs) implies **exactly 1.79 TB/s**,
the DRAM floor — so cache-cold the old config may already be optimal and the recorded 1.22x may be entirely artifact.
**The throughput claim on that pair is withdrawn** (table above updated). Following your point 2 I have kept the
entries: they are realizable and correct, and a golden row's value is not only its speed.

Four narrow-N m192 rows are flagged **unconfirmed** rather than asserted — they fit L2 but run 2.4–3.3x below DRAM
peak, so they are not bandwidth-saturated and the artifact has little room to act.

**Nothing load-bearing is affected.** The tile-M rule was derived from the 236 MB slab against an explicit DRAM floor
— precisely the regime where the artifact cannot occur — and its headline row measures 1.61 TB/s, 90% of peak and
comfortably under it.

**The ~300x fused misdeploy is immune three times over**, which matters because it is now the largest result here:
its two biggest rows exceed L2; its cold picks (35,579 µs for a 63 MB slab ≈ 1.8 GB/s) sit ~1000x *below* DRAM
bandwidth, so they are compute-pathological rather than memory-timed and L2 residency can only **understate** the gap;
and it was confirmed **end to end on a live server** — 4–6 tok/s with a healthy 2.33–2.67 acceptance length, which no
microbenchmark artifact can produce.

The caveat is now documented durably in the findings report (`plans/gemma4-m64-m256-tile-tuning-findings.md`),
including the implied-bandwidth test, since the reproduction section presents `--ab` as the method and anyone
following it on a narrow-N shape will get L2 numbers.

gnull:
## What actually sits between the shipped image and `main` — and why the "22% regression" is not yet established

Context for anyone reading the throughput numbers in this PR or planning a rebake. The article's eleven emmy
lanes pin `cloudriftai/vllm-emmy-gemma4:0.23.0-58733e02`, and there is a widely-repeated claim that an image
built from merged `main` is ~22% slower at the `4k/4k c=8` depth-2 cell (450.6 → 353.2 tok/s, identical knobs,
same box). Two things about that are worth correcting.

### The image's revision is not "ahead of main"

`58733e02` is **not an ancestor of `main`**. It sits on `origin/feature/ttft-tpot-parity`, dated 2026-07-26, and
that branch forked from `65588086` (#428). The often-repeated framing — "the published image's code is ahead of
main because the parity campaign is unmerged" — is stale: **the parity work did land**, as #429
(`c6c84dd5`, "Serving TTFT parity: rider split, m2048 chunk-quantum tier, mnbt headroom"). The branch commit
`f852c561` is its pre-merge form.

So the true delta is not "main is missing the parity work". It is:

**On the image, absent from `main`** (3 commits, all parity-campaign scaffolding whose substance is on `main` via #429):

```
58733e02 parity campaign: final scoreboard, article-protocol recipe, m2048 audit coverage docs
f852c561 serving TTFT parity: chunk+decode rider split, m2048 chunk-quantum tier, mnbt headroom
255f47ac Plan: TTFT sweep + TPOT parity campaign
```

**On `main`, absent from the image** (11 commits — everything merged since the image was cut):

```
274c153e  #441  spec-decode capture ladder
fedfd49c  #440  remove generated DiT artifacts
90d010ef  #432  the article's MTP smoke-test as recipes
cb0ab82b  #439  cloudrift a100
1d0ee241  #438  Featurizer v3 — codec-field propagation gaps + registry simplification
f17a817c  #431  DiT block + PyTorch comparison
2d1125ea  #433  first-class multi-output nodes (MIMO)
80d4c705  #436  4090 deploy-floor siblings (attention.hd512)
62a3c470  #435  eval golden pin-only offer audit
0a055cf5  #434  4090 deploy-floor siblings (mlp_down.m4096)
c6c84dd5  #429  serving TTFT parity (the merged form of f852c561)
```

Of those, most cannot plausibly move a 5090 serving cell: #434/#435/#436 are 4090 goldens, #431 is DiT, #439 is
provisioning, #432 adds recipes, #440 deletes artifacts, #441 changes a code path the gemma-4 image does not use
(it bakes its own `serve.sh`). That leaves **#433 (MIMO)** and **#438 (Featurizer v3)** as the only realistic
candidates — which is what was suspected, but now by elimination over the actual commit range rather than by guess.

### The 22% measurement is confounded and should not be quoted as a code regression

The 353.2 figure was measured on `cloudriftai/vllm-emmy:0.23.0-90d010ef` — the **generic** `vllm-emmy` image —
against 450.6 on `vllm-emmy-gemma4:0.23.0-58733e02`. Those are different image families, and the gemma-4 image
additionally ships a warm/bake layer the generic one has none of: `warm/cubin`, `warm/pack` (a frozen
execution-plan pack) and a hand-written `serve.sh`. So that comparison varies **code revision, image family and
warm-pack presence simultaneously**. It is evidence that *something* costs 22% on that path; it is not evidence
that a code change did.

### A hypothesis that ties this PR's finding to the regression

This PR shows that at decode width 192 the fused norm→linear cone cold-resolves to a `fuse`/`b128` scalar config —
132 ms where 429 µs is available, ~300x, deterministically, on a build from `main`. The published image does **not**
exhibit it: it measures 332.7 tok/s at that cell, and this PR's post-fix bare-metal run sustains 310–353 on the same
box and commit, against 4–6 pre-fix. So the misdeploy is present in fresh `main` builds and absent from the shipped
artifact.

**#438 changed codec-field propagation, which governs picks on shapes with no golden records — exactly the
cold-resolve path, and m192 had no records at all.** That makes one mechanism a candidate for both symptoms: a
cold-resolve pick regression would degrade the golden-less widths (catastrophically at m192) and could plausibly
shift other uncovered shapes by double digits. This is a hypothesis, not a result — nobody has bisected it.

### What would settle it

A same-box, same-image-family, same-warm-state A/B across three points: the published image, an image built from
`main`, and an image built from `main` plus this PR. That sweep is currently running. Until it reports, the honest
position is: **the regression is unquantified and unattributed**, and the only firm statement is the one this PR
makes — a fresh build from `main` misdeploys at m192 by ~300x, and this PR floors it.

Note the practical consequence either way: this PR is **not** a throughput improvement over what ships today. It is
what prevents a ~60x collapse at the c=64 cell the next time anyone bakes an image from `main`.

Slonegg (approved):

PR #447, opened 2026-07-30
Article-repro skill; point the gemma-4 serving experiment at the published image

Slonegg:
## What

Two things.

**1. `experiments/gemma-4-12B/serving_rtx5090` pointed at an image nobody can pull.** Its four emmy lanes pinned `cloudriftai/vllm-emmy:0.23.0-ec45dd6f` — but there is no `cloudriftai/vllm-emmy` repo on Docker Hub, only `vllm-emmy-gemma4`. They now use `cloudriftai/vllm-emmy-gemma4:0.23.0-58733e02`, the latest published tag (same digest as `:latest`), which `serving_mtp_rtx5090` already uses.

That image is warmed at `EMMY_GEN_DECODE_BUCKET=32` and the **standard** kernel fork, so this recipe's bucket-8/64 lanes miss the execution-plan pack and its six `EMMY_FAST_MATH` lanes recompile on first boot. Both are **boot cost, not numerics** — same kernels, same comparable numbers, just a slow first boot that can bump the compose healthcheck on a low-core host. `serving_mtp_rtx5090` already accepts the same trade. The header now documents it instead of leaving the next reader to rediscover it.

**2. A new `reproduce-article-benchmarks` skill.** Re-runs a published article's benchmarks and diffs them, for any article — not just the gemma-4 one:

- reads the article at run time and extracts its tables, lanes, workload points, stack, and self-stated caveats;
- finds the recipes backing it (headers cite articles by title, not URL, so discovery is fuzzy — it confirms the mapping before spending GPU hours) plus any prior committed canonical run, which is the better like-for-like reference;
- classifies each recipe: command recipes stage the working tree and measure HEAD; deploy recipes measure whatever is in the image;
- **drift report** — image tags embed the emmy sha, so "recipe pin vs latest published tag" and "published image vs `origin/main`" are both a `git log`. Says plainly that a published-image run measures the image's code, not HEAD;
- **image choice** — published tag (default), user-supplied, or built locally from current changes by delegating to `release-gemma4-image` through its verify gate, stopping before the push. Includes the scratch-dir override so no one edits a recipe in place, and forbids rewriting the control lanes;
- **hardware choice** — auto-provision, `--ssh`, or `--local`, with a GPU check and an explicit "report the rest as not-run" rule rather than substituting a different card;
- safeguards: one recipe per GPU, wall-clock caps, don't "fix" by-design failures, never edit a recipe to improve a number, keep partial results;
- reporting: lead with the setup, state a noise floor, treat a flagged cell as a lead and check the control lane, and say what wasn't measured.

## Deliberately no baseline data

An earlier draft shipped a transcribed baseline card plus a comparator script. That was removed: a checked-in copy of an article's numbers diverges from the article silently, and a stale reference invents regressions that were never there. The skill fetches the article and reads recipe headers and prior runs each run instead.

## Testing

`make test` — 2267 passed, 1 failed, 610 skipped. The one failure (`tests/compiler/trace/test_dit.py::test_dit_timestep_normalization_preserves_block_output`) is a pre-existing `ModuleNotFoundError` for a missing optional dep, reproduced on a clean stashed tree. `make lint` clean.

Not run: the benchmarks themselves — this branch has no GPU attached, and the recipe change is exactly the kind of thing the new skill exists to validate on a 5090.

## Known adjacent issue, not fixed here

`recipes/gemma-4-12B-it-emmy/recipe.yaml` and `recipes/gemma-4-12B-it-emmy-fastmath/recipe.yaml` pin the same unpullable `cloudriftai/vllm-emmy:0.23.0-ec45dd6f`. They deliberately ride the general compile-on-boot base image (the fastmath one explains why in its header), so switching them to the gemma4 image is a real behavior decision rather than a tag bump — left alone, flagged here.


6erun (approved):

Slonegg:
Addressed the four inline findings in 363e329b — all four were real.

**1. "same kernels" wording (`serving_rtx5090:29`)** — correct catch, the sentence was wrong. It described two different gaps under one claim: a lane at an unwarmed *shape* (different decode bucket) does run the same kernels and only misses the execution-plan pack, but a lane under `EMMY_FAST_MATH` deliberately selects **different** kernels — that is what the fork is — so nothing baked could have served it. Both cost only boot time, which was the point, but conflating them reads as "the fork is cosmetic". Reworded, and the skill carried the same sentence so it's fixed there too.

**2. `WebFetch` naming** — taken. The other skills in this repo describe the action ("web fetch the repo page", "web search") rather than naming a tool, so this was inconsistent with house style; now "fetch the article URL".

**3. `$SCRATCH` undefined** — real hazard, thanks. It was used in five commands with no definition, so unset it expands the Step 5 recipe override to `/<suite>` — writing at the filesystem root. Added a Step 0 that sets it via the session scratchpad or `mktemp -d`.

**4. `preflight_serving_kernels.sh` error handling** — the underlying concern was right and is fixed, but not with `set -e`. The enumeration's exit status was never checked, so a Python failure that still printed partial output would preflight a silent subset of the model's shapes and report a green `N OK, 0 FAIL` — gating a multi-hour rental on partial coverage, which is worse than not preflighting. That's now a hard fail, verified to fire.

I deliberately did **not** add blanket `set -e`, and documented why in the script: the loop exists to attempt every shape and *tally* the failures, so aborting on the first failing compile would report one bad kernel instead of the full picture a gate needs. The other cases raised are already covered — a missing `nvcc` makes each shape fail, increments the counter, and the script exits non-zero off `[ "$fail" -eq 0 ]`. Every step that must not fail silently is checked explicitly instead.

Suppressed-confidence comments, for the record: the `model_slug()` `CalledProcessError` and `spec_from_file_location` loader points are fair but both are genuinely unreachable paths here (the slug script rejects only inputs that sanitize to empty, which the caller reports on separately; the spec load is of a repo-local file that the test suite requires to exist). The `serve-models` `(none)` fallback is a real cosmetic bug — `ls` failing still leaves `sed`'s exit status — but it only shows on a tree with zero model configs, so I've left it rather than expand the diff further. Happy to fix if you'd prefer.

PR #451, opened 2026-07-31
raw-space L2 regularization in the loss

trifonova-n:
## Problem

The offline prior's rank objective is **flat in the magnitude** of any feature whose golden-pool variance is
tiny: above a saturation point, every magnitude scores identically on golden rank, so an unregularized fit picks
arbitrarily there. The surplus magnitude is invisible to every fit-time metric but catastrophic at **fork
scoring**, where a search prefix that has not decided the knob scores the feature as 0.0 — an inflated weight
hands every decided subtree an unconditional win over its undecided siblings. This is the `D_pow2_threads` 686
incident: a cold TinyLlama-1.1B serve on an RTX 4080 deployed a serial fused-norm schedule ~150x off the DRAM
floor (54x serving TPOT gap vs stock vLLM; forensics in `plans/emmy-serve-performance-findings.md` on
`fix/offline-prior-pow2-refit`).

An alternative fix on `fix/offline-prior-pow2-refit` clamps the fitter with a declared per-feature bound
(`WEIGHT_BOUNDS = {"D_pow2_threads": 112.0}`). That works for the one diagnosed feature, but the fitted value
pins exactly at the wall (the shipped weight is the constant, the fit decorative), the constant is tied to one
dataset's measured saturation point, and every future low-variance feature needs its own incident + forensics +
hand-measured bound.

## Fix: identifiability, not a symptom clamp

The fit loss becomes `objective + l2 * l2_penalty` — a **raw-space L2** penalty
(`Σ(w_z/sd)²`; `fit/linear.py`, which owns the loss) applied at every optimizer evaluation site (seed, random
samples, coordinate steps).

- **Raw-space is load-bearing**: the poisoned 686 raw weight was an ordinary O(1) z-space weight (drawn from
  `rng.standard_normal`); plain ridge on the z-space weights cannot see it. `raw = w_z / sd`, so low-variance
  features — exactly the failure class — get the strongest shrinkage.
- **The default strength is a tie-breaker, not a trade-off**: at `DEFAULT_L2 = 1e-6` the shipped artifacts' legit
  magnitudes cost ~0.015 against an O(1) rank objective, far below any genuine rank improvement, while every
  rank-flat direction gains a strict descent toward zero. No rank-worsening step is ever accepted.
- **A poisoned incumbent heals on the next refit** by plateau walk — no per-feature constants, and the mechanism
  protects every feature, including ones not yet diagnosed. `--samples N` random-restart mode (the recorded trap
  that produced the 686) is safe again.
- `emmy fit` grows `--l2` (default `DEFAULT_L2`, `0` disables), recorded in the metrics header's
  `trainer_params` and the artifact's `provenance.args`.

The functional split is respected: the penalty lives in `linear.py` (the loss owner); `rank.py`, `group.py`,
`cv.py`, `run.py` untouched; `objective()` keeps its signature as the pure data-fit term.

## Shipped artifact

`offline_weights.json` is the regularized refit from the poisoned incumbent seed (`emmy fit`, provenance
recorded: `l2=1e-06`, seed 0, 443 static + 79 dynamic cases):

| | incumbent | this PR |
| --- | --- | --- |
| `D_pow2_threads` (static) | 686.3 | **136.5** — the descent stops where rank signal begins (measured golden-objective saturation ~112) |
| static sum-of-squares | 485k | 27k |
| static golden top1 | 57 | 57 (held) |
| dynamic golden median | 19 | 19 (held) |

Note: the dynamic set's `D_pow2_threads` lands at 128.5 (the dynamic fit chains from the static seed and its
rank objective holds it) — same behavior as the declared-bounds branch, whose artifact carries 112 in both sets.
136/128 sits in the band the fork-regret study didn't measure (safe ≤112, poisoned ≥224); a twin-DB fork-regret
re-run would tighten that if wanted.

## Tests

- `test_l2_heals_poisoned_seed_on_rank_flat_feature` — a converged incumbent plus an unidentified magnitude on a
  rank-flat feature: unregularized refit keeps the poison, regularized refit walks it to ~0 while the data term
  holds or improves (the guaranteed property; exact rank equality under descent-from-seed is not guaranteed —
  plateau-walking changes the greedy path).
- `test_l2_default_is_rank_neutral_on_random_restart` — identical golden ranks with and without the default
  penalty under random-restart selection.
- `test_offline_explain_selects_the_dynamic_weight_set` now finds a genuinely-differing sentinel key instead of
  hardcoding one the chained dynamic fit may carry over verbatim.


6erun:
**Independent evaluation against the original incident dataset (the preserved TinyLlama/RTX 4080 twin node DB that produced the 686 diagnosis): this fix resolves the cold-deploy poisoning, equivalently to the declared-bounds branch (#442) on every measured metric.**

| Metric (twin DB, `eval online --dataset nodes --kernel matmul`, offline half) | Poisoned (686) | #442 bounds (cap 112) | This PR (L2, healed to 136.5) |
|---|---:|---:|---:|
| Fork sibling regret @ the incident fork (`matmul free=11264 red=2048`, PLACE+REDUCE+STAGE+TILE) | 114.87x | 8.30x | **8.30x** |
| Golden static top1 | 57 | 57 | 57 (/443, median 62) |

The one open question was the healed value itself: 136.5 sits between the saturation sweep's grid points (safe was proven at ≤112, poisoning at ≥224, nothing measured between). The direct measurement above settles it — 136.5 is on the safe plateau, byte-for-byte the same 8.30x regret as the capped artifact.

Where this approach beats the declared bounds:

- It fixes **identifiability for every rank-flat direction**, not just the one diagnosed feature: total raw sum-of-squares 485k → 27k, and the **dynamic** weight set's `D_pow2_threads` (128.5, healed here) was never covered by #442's static-set measurement at all.
- No hand-measured per-feature constant to maintain; a poisoned incumbent self-heals by plateau walk on any refit.
- Raw-space is the correct subtlety — the poisoned 686 was an ordinary O(1) z-space weight; plain ridge on `w_z` cannot see it.

The only thing the bounds had that this doesn't is a hard stop against a *genuinely rank-improving* climb into pathological territory. The incident evidence says that scenario didn't occur (golden metrics are identical at 112 and 686 — the climb was pure flatness), so I consider it a theoretical edge, not a practical one.

**Verdict: this PR supersedes #442.** Repro data for the table: `_tune/tinyllama-twins-4080/twins/twins.db` (local to the incident box), full forensics in `plans/emmy-serve-performance-findings.md` on the #442 branch — that findings file should land somewhere before #442 closes, and the stacked #443 (boot roofline audit) will be rebuilt onto main once #442's disposition is decided.

6erun (approved):

PR #458, opened 2026-08-02
Serving: attention-output aliasing for every tier (Milestone A, task 4)

6erun:
Final task of Milestone A (low-risk copy removal), stacked on #457 (A3) → #456 (A2). GitHub retargets as the stack merges.

## What

`post_attn_backing(layer, rows)` — previously M=1-only — now routes `rows` to the `attn_out` input backing of exactly the post program `forward_layer_post_device` will run: M=1, decode bucket, exact prefill chunk, or symbolic. Rider widths return `None` (the step splits row-wise across two programs — one contiguous attention output cannot alias two buffers; the copy fallback stands). The caller's `t == 1` gate is removed; under `EMMY_GEN_ALIAS_ATTN` (already default-on) vLLM's paged attention writes directly into the routed backing via `unified_attention_with_output`.

## Why this one wins on eager too

Unlike A2's chaining (whose skip depends on the *output* view surviving to the next call — broken by the eager clone), the alias makes the **upload side** pointer-equal: attention writes into the buffer the post upload reads from, so the skip fires under capture *and* on eager chunk steps. Per layer per step that's the `[T, attn_width]` attention→post copy gone — at gemma's chunk width (2048×4096 fp16) ~16 MB/layer/step.

## Capture-safety contract (unchanged, now per tier)

Views are cached per `(layer, tier)` and minted only on uncaptured steps (`from_dlpack` negotiates a stream sync — capture-fatal); an uncached mid-capture request returns `None` → ordinary copy path. vLLM's per-size uncaptured warmup populates the cache before any capture, same as before.

The routing MUST mirror `forward_layer_post_device` — a mismatch is still correct (upload copies), it just silently loses the skip; the docstring pins this invariant and the test pins the routing.

## Validation

- Tier-routing pointer pins: each width's backing is the routed program's buffer; rider and over-capacity widths → None (`test_gen_prefill_device_gpu.py`).
- Aliased-post value check vs the host path at the decode tier; the sym-aliased capture primitive was already pinned in A2's `test_run_device_sym_aliased_input_backing_replays_live`.
- Full suite: 2818 passed, 159 skipped. Lint green.
- Serving-level effect: measured in the batched Milestone-A 5090 session (chunk cell + over-bucket cell, alias on/off sub-A/B) — launching next.


6erun:
## Milestone A measurement — rented 5090, gemma-4-12B-it, 2026-08-02/03

Protocol: main @ e07ea37f vs this stack's tip (A2+A3+A4; A1 already in base), same box, shared cubin cache, packs rebuilt per arm from identical evidence — a box-local 12-twin tuned DB (m16 + m2048 + sym, both layer classes), which was necessary because the 5090 golden set is orphaned for the current fused-kind forms on this base (known; fixed in unmerged #449 — the twin tune restored sane baselines: chunk TPOT 2760 → 54 ms, over-bucket 58 → 37.5 ms).

| cell | metric | main (x2) | milestone (x2) | alias-off |
| --- | --- | --- | --- | --- |
| chunk p2048/c8 (4000 in/128 out, bucket 16) | median TPOT | 54.34 / 54.18 | **53.93 / 54.00** | 54.38 |
| | median TTFT | 2105 / 2095 | 2200 / 2085 | 2089 |
| over-bucket c64 (8 in/64 out, bucket 16) | median TPOT | 37.54 / 37.78 | **37.21 / 37.83** | — |

- Greedy chat outputs: **content-identical** main vs milestone (3 prompts, temperature 0).
- TPOT: milestone ahead by ~0.2–0.4 ms on the best-run comparison — at the noise edge, direction consistent, **no regression anywhere**. The alias-off sub-run sits ~0.4 ms behind alias-on, same caveat.
- req/s on the over-bucket cell was bimodal (8 vs 22–23) by boot order in BOTH arms symmetrically — TPOT is the robust arm-comparison metric; nothing arm-specific.

**Milestone verdict, per the plan's own gate:** the integration D2D copies Milestone A removes are µs-class against 37–54 ms steps — measured, they are NOT where the remaining serving gap lives. That answers the plan's pre-Milestone-B question: `bind_io`'s TPOT justification is weak; the remaining gap belongs to glue-kernel time and eager host overhead (lock/DLPack dispatch), i.e. compiler fusion and whole-chunk-step capture. Milestone B's strongest surviving motivation is Phase 5 (Torch-owned memory → KV capacity), not copy removal.

Milestone A still stands on its merits: strictly fewer graph nodes/copies, correctness pinned at every seam, no regression, and the measurement itself.

Slonegg (approved):

PR #460, opened 2026-08-02
golden gate: MAJOR-GAP hazard baseline + pins beat routing entries (stacked on #449)

6erun:
Stacked on #449 (base branch `feature/gemma4-mtp-golden-coverage`) — two commits. Context: on main the
2026-08-02 5090 serving audit showed every fused twin family cold-resolving (114–1014x over the weight-streaming
floor, p2048/c8 chunk cell at TPOT ~2760 ms) while the drift gate stayed green; #449 closes the coverage, this PR
makes the gate class-aware so the loss can't ride a routine baseline edit again, and unblocks the re-record path
for the four keys #449 had to leave open.

## 1. Drift gate: warp-contraction gaps ratchet through their own `EXPECTED_MAJOR_GAPS` baseline

The #446 PLACE retirement uncovered all 18 fused serving forks and the gate stayed green because the keys were
appended to the generic `EXPECTED_GAPS` — indistinguishable in review from an aux-key edit. Now:

- `major_gap_keys` (warp forks with a real reduce — the misdeploy/hang class) ratchets against its own
  `EXPECTED_MAJOR_GAPS`, exact-match in both directions, with a failure message that says to close (seed a golden
  or a routing row) rather than list.
- Every entry carries a dated reason + a concrete burn-down condition; the 5090 set is exactly the four
  `mlp_down_fused` keys #449 documents, the 4090 set is the known deferred mirror re-tune, grouped by width.
- `test_no_major_key_hides_in_the_generic_baseline` keeps major keys out of `EXPECTED_GAPS` statically.

Verified off-GPU on #449 + current main: RTX 5090 audits MATCH 176 / DRIFT 0 with only the four down majors;
both card gates + the invariant test pass.

## 2. Placement routing: schedule-family pins are authoritative over recorded routing entries

#449's finding "the staged computed-A form no longer realizes in the isolated snippet (the tile-IR 1s window)"
turned out to be misattributed — bisect shows the pinned fused replay already fails at 0ffea99d (the commit that
*recorded* the m192/m256/dynM `.cut` routing rows) and realizes clean at its parent, while the 1s commits change
nothing: the snippet's computed-A cone still binds and offers `d1/sync` on current main. The actual bug:
`route_cut` consulted the recorded same-shape `.cut` ROUTING golden inside pinned compiles too (only `PLACE` pins
were authoritative), so a pinned `d*/sync` fused row silently compiled the cut's pieces and gated
`realized (off)`. Every fused golden replay was failing against its own recorded spelling as soon as a same-shape
`.cut` row landed — and #449's 14 new cut rows extend that lockout to every serving width.

Fix: `route_cut` skips the recorded routing entry whenever any schedule-family pin (TILE/STAGE/REDUCE/WORK/
RASTER/WSPEC; bare `EMMY_<KNOB>` var or `EMMY_KNOBS` aggregate key) is live — matching `Knob.narrow`'s rule that
pins out-rank every golden tier. Explicit `PLACE` pins keep full routing control; the unpinned deploy path is
untouched. New test `test_schedule_pin_suppresses_the_routing_entry` (fails without the fix). Bare schedule pins
apply compile-wide, so the suppression is compile-wide too — same scope as the pins themselves.

Consequence: fused rows bench honestly again — the third commit uses exactly that. The #449 findings-plan section on
the snippet realization loss should be read with this correction (also: at m192 the gate_up fused form is routed
away in-model as well — the in-model fused MATCH at that width comes from the down node, which has no legal cut
seam).

## 3. The four `mlp_down_fused` keys seeded — the 5090 major baseline is EMPTY

With the routing fix in place, the four remaining down keys (m8/m64/m2048/m4096) were tuned + benched on a
vast.ai 5090 (isolated `.lin` snippets, one shared cold DB/prior; `-O3` A/B vs eager): m8 105.4 µs (eager 100),
m64 133.5 (99.4), m2048 2066.0 (1190.2), m4096 3238.6 (2259.1). The wide widths trail eager on the std-lane
FP32-accumulate wall — honest pins, not wins (the unseeded fork cold-resolved ~100x off in serving). Two
record-time findings, both noted in the YAML block: m8/m64 record the `w1x8 f2x2/k4 g4k` spelling (the snippet
greedy's `w1x16`/`g*a` winners are not offered on the in-model geglu→down cone and audited DRIFT), and all four
dicts drop the snippet `record_knobs`' `REDUCE@a1: ''` stamp (a snippet-nest-only axis; the canonical-spelling
gate rejects the key and the axis-keyed pin made the split rows unmatchable in-model). Pinned replays on the
5090 reproduce the records within the noise band (m8 105.2, m2048 2100). `EXPECTED_MAJOR_GAPS` for the 5090 is
now `set()` — the in-model audit reads MATCH 184 / DRIFT 0 with `major_gap_keys` empty.

## Validation

- `make test`: full suite green (2814 passed, incl. the two audit-gate cards and the new routing test).
- `make lint`: clean.
- Off-GPU in-model audit (`emmy eval golden --in-model`): 5090 MATCH 184 / DRIFT 0 / `major_gap_keys` empty.
- On the 5090: pinned replays of the new rows reproduce the recorded µs; the greedy (unpinned) pick now resolves
  through them.

### DIFF
--- a/emmy/compiler/pipeline/passes/lowering/tile/_cut.py
+++ b/emmy/compiler/pipeline/passes/lowering/tile/_cut.py
@@ -59,6 +59,24 @@ def _place_pins() -> dict[str, str]:
     return pins
 
 
+#: The schedule knob families a golden / ``--ab`` row pins. A live pin from any of them marks a
+#: pinned re-record compile, where the pin — not a recorded routing entry — must decide the form.
+_SCHEDULE_FAMILIES = ("TILE", "STAGE", "REDUCE", "WORK", "RASTER", "WSPEC")
+
+
+def _schedule_pins_live() -> bool:
+    """Whether any schedule-family knob pin is live (a bare ``EMMY_<KNOB>`` var or an
+    ``EMMY_KNOBS`` aggregate key). Pins are authoritative over every golden tier — a recorded
+    ROUTING entry must not reroute a pinned compile: the pinned fused row would silently compile
+    the cut's pieces and gate as ``realized (off)`` (the 2026-07-31 fused re-record dead end,
+    where every fused golden replay failed against its own recorded spelling as soon as a
+    same-shape ``.cut`` routing row landed). Bare schedule pins apply compile-wide, so this
+    suppression is compile-wide too — matching ``Knob.narrow``'s bare-pin scope."""
+    if any(config.knob_raw(f) is not None for f in _SCHEDULE_FAMILIES):
+        return True
+    return any(family_of(k) in _SCHEDULE_FAMILIES for k in parse_knob_spec(config.knobs_aggregate()))
+
+
 def _card_has_routing(gpu_name, cap) -> bool:
     """Whether ANY routing golden exists for this card — the cheap gate that keeps the per-kernel
     seam scan off the common compile (no pins, no routing entries → recognition is untouched)."""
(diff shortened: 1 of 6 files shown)


Slonegg (approved):

6erun:
@Slonegg this one is stacked on your branch, so I'll wait until you merge it and delete the branch.

Slonegg:
> @Slonegg this one is stacked on your branch, so I'll wait until you merge it and delete the branch.

Got you. Merged into #449

PR #482, opened 2026-08-11
Experiment: replace merge-loop gates with placement cuts

Slonegg:
## Summary

This experiment removes every merge_loop_ops materialization-policy gate except the 8x aggregate compute blow-out cap. Structural region ownership, genuine splicer rejection, and the decided-cut workspace fence remain invariants.

The existing try_flash / _recognize path is extended across the boundaries exposed by gate-free fusion; no second recognizer or pass is added. The branch is rebased on main at fcbc880f (#483).

The final cleanup removes 508 lines and leaves merge_loop_ops focused on region splicing plus the compute cap.

## RTX 5090 results

- Swept all 472 regimes across the four canonical RTX 5090 golden files: 52 generic, 397 Gemma 4, 10 Laguna, and 13 OLMoE.
- Repeated O3 A/B replay promoted two Gemma pointwise schedules: cut_combine.m8 and cut_cone_scale.m32 switch from TILE=f4 to untiled, about 0.795 us versus 0.858 us.
- Flash remains recovered without a merge gate: attention.hd64 measures 9.6 us standard / 8.3 us fast-math; normalized gemma4_12b.attention.hd256 measures 36.5 us / 34.1 us.
- Schedule pins do not repair gate-free normalized-linear/MLP compute multiplication; O1 search remains orders of magnitude behind recorded materialized kernels.

## Future perf gaps

- Eleven Gemma regimes had no successful O1 candidate: LM-head m32/m64 and nine large explicit cut twins.
- LM-head m64 isolated O3 compilation did not finish before wrap-up; m32/m64 are the first follow-up targets.
- Remaining functional experiment gaps are transcendental duplication across contraction columns and the known tiny two-linear runtime miscompile.

The full history and methodology are in plans/merge-loop-cuts-experiment.md. Raw sweep rankings, logs, paired JSON, and the future-session checklist are under the ignored _tune/merge-loop-cuts-5090-20260810/ directory.

## Verification

- Ruff and git diff checks pass.
- Focused fusion, flash-recognition, and golden-config suite: 84 passed; the only assertion failure is the documented transcendental-duplication experiment gap.
- O3 flash and two promoted golden A/B rows have clean integrity flags.
- The duration guard separately reports one 14.3s test missing from tests/durations.json.

PR #490, opened 2026-08-12
fusion: restore the multi-load-of-reducer refusal — close the #482 gemma MLP offer drift

6erun:
## Summary

#482 removed every merge-loop materialization gate except the 8x compute cap, and landed with a known consequence (its plan's Verification section): gate-free fusion splices the RMSNorm producer of the gemma-4 normalized-linear/MLP cone into BOTH the gate and up contractions, so no offered candidate realizes the recorded `mlp_geglu` goldens (m32/m4096, std + fast-math) and a fresh deploy falls past the golden tier to the cold offline prior. Measured live on a rented 5090 (2026-08-12, fm lane): the boot's roofline audit flags the post twins at 39,623x (decode m32), 15,173x (m1) and 104,464x (chunk m4096) over their weight-streaming floors — the same class as the 2026-08-02 zero-coverage incident.

This restores exactly the exception the experiment plan itself names as the smallest non-placement hardening ("restoring only the refusal for a reduce-heavy producer read through more than one Load"): a reduce-heavy producer read through more than one `Load` across the region stays materialized. A rowmax-bearing sink is exempt, so the softmax assembly still fuses onto its score producer and the #482 flash-recognizer follow-up stands; a mask-add discount keeps multi-mask sliding-window scores under the reduce-heavy threshold.

## Validation

Off-GPU in-model twin audit (gemma-4-12b serving matrix, RTX 5090 card, `audit_card`):

| lane | main (02a94511) | this branch | #483 baseline |
| --- | --- | --- | --- |
| std | MATCH 71 / DRIFT 4 | **MATCH 197 / DRIFT 2** | MATCH 199 / DRIFT 2 |
| fast-math | MATCH 1 / DRIFT 2 | **MATCH 15 / DRIFT 2** | MATCH 15 / DRIFT 2 |

The geglu drifts are gone in both lanes; the residual DRIFT rows are the pre-existing `gate_up_cat.m1024.lin` and `mlp_down.dynM` entries (documented before #482), and std GAP improves 48 → 38 vs #483.

- `tests/compiler/passes/test_fusion_rules.py`: 53 passed, 1 xfailed — includes two new cases (refusal fires for a twice-loaded reduce-heavy producer; correctness of the materialized result)
- `tests/compiler -k "golden or placement or flash or fusion"` on an RTX 4080: 320 passed, 11 skipped, 1 xfailed, no new failures — flash recovery and placement routing intact
- `ruff check` / `format --check`: clean
- Full `make test` on this box carries pre-existing failures from the transformers version squeeze (Laguna tests need >5.12.1 while gemma-4 serving under vLLM 0.23 breaks at 5.15) — independent of this change, fix in flight separately

## Notes

- The 5090 re-baseline bench sweep this incident interrupted is running at #483 (last-good) and can re-validate this branch end-to-end on the same box afterwards if wanted.
- `emmy eval golden --serving-config` on main also fails the #483 realization-matrix pre-check (4,171 missing realizations) — that is a separate gap between the golden file and the stricter gate, not addressed here.

### DIFF
--- a/emmy/compiler/pipeline/passes/ARCHITECTURE.md
+++ b/emmy/compiler/pipeline/passes/ARCHITECTURE.md
@@ -221,15 +221,14 @@ How to comply:
 - **When generalizing an existing rule, normalize its incidental divergences** (one dtype rule, one index rule)
   and name the behavioral deltas explicitly in the commit — don't preserve two behaviors behind one entry point.
 
-Loop fusion's profitability guard follows the same structural rule. `loop/fusion/010_merge_loop_ops` counts aggregate
-arithmetic and reads, and separately counts transcendental executions. The separate count prevents an expensive
-pointwise producer such as GELU's `tanh`, originally evaluated once per `(M,K)` element, from moving under a
-contraction's `N` loop merely because the contraction's much larger cheap-FMA count hides that duplication. Flash
-attention is the structural exception: a merged softmax-then-P@V offer deliberately streams `exp(score)` without
-materializing the probability matrix, so the tile recognizer owns that composite and the generic transcendental
-brake does not split it. The automatic brake is also limited to unary/single-source activation cones. A multi-source
-gated activation such as GeGLU/SwiGLU remains an evidence-controlled placement choice, preserving its GPU golden
-coverage until that entire cone has been measured.
+Loop fusion's materialization policy follows the same structural rule. `loop/fusion/010_merge_loop_ops` keeps two
+guards: the aggregate compute cap (a merge is refused when summed leaf work grows more than 8× over the region's
+own) and the multi-load-of-reducer refusal (a reduce-heavy producer read through more than one `Load` across the
+rest of the region stays materialized — splicing it would re-execute its reduce per demand site, the boundary that
+keeps a fused normalized-linear/MLP cone realizing its recorded goldens). Flash attention is the structural
+exception: a rowmax-bearing sink may swallow its score producer, because the tile recognizer owns that
+softmax-then-P@V composite and re-synthesizes it without the duplication. Every other historical merge gate is
+gone; those boundaries are placement decisions now.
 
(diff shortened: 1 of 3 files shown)


6erun:
Live-validated on the rented 5090 (main + this branch, fresh pack dir, fm lane, small_c1 cell): boot logs **0 enumeration-drift warnings** (main logs 5+ on the geglu shapes and falls to the cold prior), roofline shows only the pre-existing chunk-m4096 flags (23–29x, identical class/values to the #483 baseline boot — not introduced here), and the bench lands at TTFT 110.9 ms / TPOT 19.11 ms / 51.2 tok/s vs the #483 baseline's 115.6 / 19.10 / 51.1 on the same box — #483-class serving restored on main.

Slonegg (changes requested):
No gate restoration in loop fusion. Find another way. I am trying to get rid of all these loop fusion gate hacks.

6erun:
Ack — dropping the fusion-gate approach. Investigating the placement-layer repair instead: whether the gate-free geglu cone still offers a legal cut seam at the norm feeder (then the fix is routing rows re-keyed to the new tree, no fusion changes), and failing that, factoring the shared reduce-heavy subterm at the cut/recognition layer — the same move #482 used for flash's feeder workspace, which would also close its listed 'transcendental duplication across contraction columns' gap. Will update this PR or replace it.

6erun:
Superseded by #496 per the review direction: the root cause turned out to be a recognition bug (order-sensitive `_cone_value_key` failing the documented value-tree dedup on the per-channel duplicated norm cone), fixed with a commutative-canonical sort in `_atomize.py` — no fusion changes. #496 also documents the experiment showing routing rows cannot repair this seam, and flags the remaining #482 flat-cell class (o_proj+norm / down+norm deploying scalar with no golden consultation) as a separate open gap.

PR #496, opened 2026-08-12
recognize: commutative-canonical cone keys — gemma-4 geglu goldens realize in-model again

6erun:
## Summary

Replaces #490 per review direction (no fusion gates) — the fix lives in the recognition layer and does not touch loop fusion.

**Root cause of the gemma-4 5090 geglu drift** (the #482-exposed regression: cold deploys at 39,623x–104,464x roofline, the 2026-08-02 incident class): gate-free fusion inlines the RMSNorm cone once per gate/up channel, and one copy spells the normalize multiply's operands the other way round (`x̂·s` vs `s·x̂`). `_cone_value_key` was positional, so the two copies keyed as different A values, `bind_prologue_contraction` declined the shared-producer composition, the kernel demoted to the scalar tier, and no offered candidate realized the recorded `mlp_geglu` m32/m4096 rows (std + fast-math). Sorting a commutative op's child keys makes the value-tree equality the docstring already promises hold.

**Why not routing rows:** verified by experiment — a bare `PLACE=cut` pin realizes legally, but the parent piece re-recognizes to a flat cell (two Accums in one K loop), never consults the golden tier, and deploys a single-option scalar kernel. The seam that matters is raw duplicated statements, not a tree site; no PLACE key set can repair it. The old #449 routing entries are bare `PLACE: cut` rows for other kinds — there was never a path-keyed entry for these shapes to re-key.

## Validation (off-GPU in-model twin audit, gemma-4-12b serving matrix, RTX 5090 card)

- main: std MATCH 71 / **DRIFT 4**, fm MATCH 1 / **DRIFT 2** (all six = the geglu forks)
- this branch: std MATCH 75 / **DRIFT 0**, fm MATCH 3 / **DRIFT 0** — all six drifts become MATCHes, GAPs unchanged, no new drift; the recognized tree/seams match the #483 `norm_gate_up` shape and the `.lin` row wins at its recorded 178.6 µs
- `tests/compiler/passes/ + tests/compiler/pipeline/`: 807 passed, 1 skipped, 1 xfailed (includes a new regression test: duplicated cone with commuted args still shares one A)
- ruff clean

Live 5090 boot + bench validation is queued and will be posted here.

## Remaining #482 gap this does NOT fix (flagged, not addressed)

The in-model MATCH count is still far below #483 (75 vs 199 std): #482's broader merges lift the o_proj+norm and down+norm kernels to flat cells (multiple top-level reduces), which deploy as single-option scalar kernels with **no golden consultation at all** — the recorded `mlp_down_fused.*.cut` routing entries silently no-op (entry found, no seam — observed live on the post32 twin). This class is invisible to the DRIFT gate because no fork exists. It needs its own decision (likely recognition-side typing of multi-Accum cells, or the shared-subterm feeder factoring #482's plan lists as the 'transcendental duplication across contraction columns' gap).


Slonegg (approved):

6erun:
**Live 5090 validation (fresh box, main-equivalent tree = d44a9b4a + this fix):** the fix itself works live — **0 enumeration-drift warnings** (was 5+ before). But the boot roofline shows serving is still broken by the flat-cell class this PR flagged:

```
L0/L5.post.decode.m1    17,248x / 19,124x over floor
L0/L5.post.decode.m32   23,799x / 26,496x over floor
L0/L5.post.chunk.m4096  80,982x / 89,632x over floor
L0/L5.pre.chunk.m4096   29x / 28x  (the known pre-existing class — healthy by comparison)
```

The bench did not complete (at these per-launch costs the profiling pass alone exceeds the health window). So the o_proj+norm / down+norm flat cells are **on the hot path and dominate**: they deploy as fork-less single-option scalar kernels with no golden consultation, invisible to the DRIFT gate. **gemma-4 5090 serving on main remains unusable; fcbc880f (#483) is still the last servable state.** The multi-Accum flat-cell decision (recognition-side typing, feeder factoring per the 'transcendental duplication across contraction columns' gap, or narrowing the merge) needs an owner — cross-posting to #482.

PR #499, opened 2026-08-13
NVFP4 support in the emmy compiler

gnull:
NVFP4 in the emmy compiler: reading the format, decoding it as ordinary graph algebra, and multiplying it on Blackwell's block-scaled 4-bit tensor cores.

NVFP4 is NVIDIA's 4-bit float format — e2m1 values packed two per byte, an fp8 e4m3 scale per 16 values, one f32 scale per tensor. This PR compiles such a checkpoint two ways. **W4A4** puts both operands on the tensor core at 4 bits, lowering to `mma.sync.aligned.m16n8k64...kind::mxf4nvf4.block_scale`, which `ptxas` assembles only for `sm_120a`; this is the scope. **W4A16** keeps the weights packed and the activations 16-bit; it is scaffolding that serving runs today. No pass below the frontend knows what NVFP4 is — the loader writes the format into the graph as ordinary algebra at model-read time, and the kernel side RECOGNIZES the shape that algebra makes.

## Results

512×4096×4096 on an RTX 5090, every row CUDA-graph captured. The comparisons answer three different questions, so they are grouped by question rather than ranked together.

| | µs | |
| --- | --- | --- |
| **emmy W4A4, whole linear** (deploy path) | **42.1** | quantize 7.6 + matmul 34.5 |
| **emmy W4A4, the multiply alone** | **34.5** | the block-scaled cell |
| emmy W4A16, whole linear | 126.8 | the scaffolding path |
| | | |
| *Same program* — quantize→dequantize→matmul in f16, torch eager | 2720.6 | emmy is **65× faster** |
| *Same program* — the same, under `torch.compile` | 145.6 | emmy is **3.5× faster** |
| | | |
| *Same instruction* — cuBLASLt NVFP4 GEMM via `torch._scaled_mm` | **16.9** | **cuBLAS is 2.0× faster** than emmy's cell |
| *Same instruction* — vLLM's CUTLASS fp4 GEMM | not measured | vLLM is not installable on the measurement host |
| | | |
| *No quantization* — f16 linear, torch eager | 94 | emmy's W4A4 is **2.2× faster** than not quantizing |
| *No quantization* — f16 linear, `torch.compile` | 94 | |

Read them as three separate claims. Against **the same declared program** — the quantize→dequantize→matmul this PR compiles — emmy is 3.5× faster than `torch.compile` and 65× faster than eager; that is the comparison this PR is for. Against **the same instruction**, cuBLAS's own NVFP4 GEMM is 2.0× faster than the cell emmy generates, which is the honest gap and the clearest thing to aim at next. Against **no quantization at all**, the whole W4A4 linear still comes in at 2.2× the f16 baseline, so the format is paying for itself end to end. W4A16, by the same yardstick, is 1.35× SLOWER than eager f16 — which is why it is called scaffolding here and not a result.

One caveat on the `_scaled_mm` row: it is fed synthetic codes and scales and was not checked for numerical equivalence with our program. It is a throughput reference for the same instruction class, not a verified-equivalent computation.

**The tuning story sits in one number.** Compiled whole-program instead of per-target, the same W4A4 linear costs 4012.8 µs — because the activation quantize, the same kernel identity, costs 7.6 µs resolved as its own target and 3970.6 µs fused, where it lands on a grid of 2. A **522× spread from placement alone**. That is a finding about the compile lane, not about NVFP4, and it is written up below.

## Reproducing these numbers

Each command below was run on the PR head before being written here. The full trail — every arm, the sweep driver, the clean-DB reset, the baseline harnesses — is in `plans/nvfp4-golden-tuning.md` (branch-lifetime note; it does not outlive this PR).

Freeze both arms once, so nothing but the declared scheme differs:

```
emmy compile -c "torch.nn.Linear(4096, 4096, bias=False).half()(torch.randn(512, 4096).half())" --quantize nvfp4       --ir torch -o w4a4.ir  --dump-dir dump-nvfp4
emmy compile -c "torch.nn.Linear(4096, 4096, bias=False).half()(torch.randn(512, 4096).half())" --quantize nvfp4-w4a16 --ir torch -o w4a16.ir --dump-dir dump-nvfp4-w4a16
```

The two headline emmy numbers, through the recorded goldens (42.1 and 126.8):

```
emmy trace dump-nvfp4/00_input.json --target sm_120 -o golden-nvfp4.yaml
emmy run --golden-file golden-nvfp4.yaml --bench --iters 200 --warmup 20
```

The whole-program contrast that produces the 522× line (4012.8):

```
emmy run dump-nvfp4/00_input.json --bench --iters 200 --warmup 20
```

The no-quantization reference (94), through emmy's own bench flow:

```
emmy run -c "torch.nn.Linear(4096, 4096, bias=False).half().cuda()(torch.randn(512, 4096).half().cuda())" --bench --bench-backends eager,tcompile,emmy --iters 200 --warmup 20
```

The same-program (2720.6 / 145.6) and same-instruction (16.9) baselines cannot go through `emmy run` — the tracer has no `aten.bucketize`, and `_scaled_mm` is not a traced op — so they use small capture harnesses kept with the trail in `plans/nvfp4-golden-tuning.md`.

The serving verification (223 passed, 0 failed):

```
pytest tests/serving -q -p no:randomly -n 8 --dist=loadgroup
```

## Scope

**In:** the compile side of the native path — the format's dtype and loader, the encode, the block-scaled atom and its arch-suffixed compile path, the static activation speller, packed-pair recognition on both operands, the schedule offer, the four-slab staging, the drain, and the W4A16 scaffolding that keeps serving working.

**Out, and named in the charter below:** serving the native path (the runner does not spell 4-bit activations, so every serving number here is the coded W4A16 trunk), the hybrid linear-attention slice, and the 27B.

## Serving today, with its reservations

The coded W4A16 trunk works and is re-verified on this head: **223 passed, 29 skipped, 2 xpassed, 0 failed** across `tests/serving` on the 5090. It serves DENSE models through emmy's generation runner, weights bound packed from the checkpoint and decoded in-kernel.

Three reservations, plainly. It is **W4A16, not W4A4** — activations stay 16-bit end to end. It is **correct but not yet fast**: the decode-step profile in the appendix shows where the time goes, and the comparison against vanilla vLLM is not done. And the **hybrid slice is not in this PR at all**.

For context on why this work exists: vLLM serves these checkpoints through a prebuilt kernel library shipped per release — on Blackwell that includes a CUTLASS fp4 path, with W4A16-style dequantize-and-GEMM fallbacks on older cards. Emmy instead GENERATES the kernel for the shape in front of it and tunes it per card, which is what makes a per-shape result like the 522× placement finding visible at all. **These vLLM specifics are stated from general knowledge and were not verified here** — vLLM is not installable on the measurement host (its `serving` extra pins a torch that would replace the one every compiler test runs against), which is also why the CUTLASS row above is unmeasured.

## Findings a reviewer should know

Three are about the tuning surface rather than about NVFP4, and they are why the whole-program number is 100× the deploy number. Each is evidence-backed.

**The whole-program lane misprices the activation quantize by ~500×.** Same kernel identity, 7.6 µs per target and 3970.6 µs fused onto a grid of 2. Measured exhaustively: 125 axis-scoped `REDUCE` combinations and 33 kernel-global `WORK` values, 158 pinned configurations, none of which moves it off grid 2.

**`WORK` and `PLACE` are kernel-global as PINS while the fork assigns them per kernel.** Greedy really does give the quantize `t4` and the matmul `w2x4` in one program, but no pin can say so: `PLACE=cut` takes the quantize from 3974 µs to 1420 µs and simultaneously costs the matmul its native cell (34.3 → 127.4 µs, worse than W4A16).

**An enumeration-driven sweep cannot find `PLACE`.** `enumerate_graph` returns knob rows; `PLACE` is a pin, not a knob. The 158-config sweep had that blind spot by construction, and only a hand test found the 2.6× whole-program lever.

**The evidence that greedy leaves its own per-kernel freedom unused is INDIRECT** — greedy's clean-DB total (4012.8 µs) is worse than the best single global pin (1560.4 µs), which a fork exploiting per-kernel freedom should not be. The direct test is named in the charter and was deferred by choice.

## Known gaps

Each entry says what a reader would reasonably expect to be possible, what is there instead, and why to the extent it is understood. The last part is tagged by how it is known: **measured** (a run backs it), **informed reading** (a plausible cause, not profiled), **by decision** or **by scope** (deliberate), or **not diagnosed**.

- [ ] **The block-scaled cell on a real layer.** *Expect:* `emmy compile nvidia/Qwen3-8B-NVFP4 --layer 0 --target sm_120 --ir cuda` puts the cell on most of the layer's marked matmuls. *Instead:* it lands on one kernel — 6 dump files carry the instruction, none carry the W4A16 byte-slab drain — against five kernels before the regression this branch fixed. *Why:* **measured** — recognition is not the problem, since the pair matcher binds both operands on every synthetic shape tried, with and without a fused norm; what is left is which sites the search SELECTS. The 522× placement spread on the quantize kernel is the same class of problem on a shape where it could be isolated, and is the best evidence for where to look.
- [ ] **`q_proj` and `k_proj` as contractions.** *Expect:* they bind like the other projections and get a tensor-core tier enumerated. *Instead:* they never bind as contractions at all, so nothing is enumerated and they stay scalar reduces. *Why:* **not diagnosed.** This sits underneath the coverage count above and is the concrete next thing to chase there.
- [ ] **The `gate` + `up` two-channel kernel.** *Expect:* the fused gate/up node takes the pair reading through both channels. *Instead:* the reading accepts any channel arity now, but every contraction in all three real W4A4 graphs is single-channel, so the widening has never run on a real shape and carries no numeric parity of its own — the per-channel residue and the per-channel scale-fragment name are where a wrong scale would be silent. *Why:* **measured** (a discriminator run over the three graphs).
- [ ] **Matching cuBLAS on the same instruction.** *Expect:* a generated cell reaches `torch._scaled_mm`'s 16.9 µs at 512×4096×4096. *Instead:* 34.5 µs, a 2.0× gap. *Why:* **informed reading, not measured** — a young generated kernel against a mature vendor library. No profile has been taken to attribute the gap, so the cause is a guess and the number is the only fact.
- [ ] **Serving the native path.** *Expect:* `emmy serve nvidia/Qwen3-8B-NVFP4 --generate` runs the declared W4A4 program. *Instead:* it runs the coded W4A16 trunk. *Why:* **known by code path** — the activation spell lives in the compile command's stamp, and serving's program seam was never taught to carry 4-bit activations. This is the follow-up's first item.
- [ ] **Serving speed.** *Expect:* the serving path's kernel picks are tuned and its goldens recorded, as a shipped path's would be. *Instead:* the picks are whatever the prior selects untuned, and no goldens are recorded for them. *Why:* **by decision** — W4A16 is scaffolding here, kept working and deliberately not polished, and recording goldens for it would dress up a path this PR is not claiming.
- [ ] **The hybrid 27B slice.** *Expect:* the full-attention quarter of a Qwen3.8-27B checkpoint served through emmy while the linear-attention layers stay on stock vLLM modules. *Instead:* the carve is removed from this PR and the per-layer-kind dispatch it would need never existed. *Why:* **by scope** — the carve's machinery was never joined to a consumer, so it ships from the follow-up rather than half-wired from here. Distance to done is four pieces of very unequal size: the carve returns as ONE revert (`b3056eea0` was built as a single revertable unit precisely so it could); the unpack mismatch in its consumer is a small verified fix (`vllm_model_gen.py` destructures three tensors where a gated `pre` returns four); the serving-side dispatch is the real work and has never existed at all — per-layer-kind routing plus the DeltaNet cache wiring, days rather than hours; and the whole thing then wants box verification. It starts from proven ground rather than a guess: the compile-side evidence is already recorded, both layer kinds carving bit-exactly against their eager blocks. **Target models:** the serving target is the Qwen3.8-27B family vLLM's own recipes serve (`Qwen/Qwen3.8-27B`, per recipes.vllm.ai). No official `nvidia/` NVFP4 repack of 3.8 exists yet, so `nvidia/Qwen3.6-27B-NVFP4` remains the architecture-identical official stand-in. The lead 3.8 candidate is **`Inferact/Qwen3.8-27B-NVFP4`** — the repack recipes.vllm.ai itself serves for the nvfp4 / rtx_5090 variant, and verified by reading its published config rather than by any run on our hardware: modelopt producer, `quant_algo` NVFP4, `model_type` qwen3_5. Its exclude list carries a consequence worth stating, because it makes the hybrid work SMALLER than the shape above suggests: every DeltaNet-layer projection stays unquantized across all 48 layers (`in_proj_qkv`, `in_proj_z`, `in_proj_b`, `in_proj_a`, `conv1d`), as do `lm_head`, the embeddings, the visual tower and the MTP head. So the stock-vLLM DeltaNet path this slice hands the linear-attention layers to would be handling NO quantized weights in this repack — the per-layer-kind dispatch has to route correctly, but it does not additionally have to teach vLLM's DeltaNet anything about NVFP4. Alternates, both checked the same way and neither recipes-endorsed: `RadixArk/Qwen3.8-27B-NVFP4` (modelopt MIXED_PRECISION, same layout convention as nvidia's 3.6) and `gittensor-model-hub/Qwen3.8-27B-NVFP4-RTX5090` (modelopt pure NVFP4). For all three, a config read is the ONLY verification anyone has done — no weights have been loaded, no parity taken, nothing run. **Calibration quality is unvouched** on every community repack until our own qualification runs say otherwise, and that applies to the lead candidate exactly as much as to the alternates.
- [ ] **Twins for a quantized model.** *Expect:* `scripts/capture_gen_twins.py` on an NVFP4 checkpoint produces the CODED graph serving actually runs. *Instead:* it produces float16 twins. *Why:* **known by code path** — `serving/twins.py` has no NVFP4 arm. Release infrastructure is outside this PR's scope.
- [ ] **`graph.to_dict()` on an NVFP4 graph.** *Expect:* it round-trips like any other graph. *Instead:* a known round-trip defect on the packed constants. *Why:* **by decision, unchased** — nothing this PR needs goes through that path, so it was left alone rather than fixed opportunistically.
- [ ] **Pinning a schedule per kernel.** *Expect:* `EMMY_KNOBS="WORK=…"` addresses one kernel the way `REDUCE@a2` does, and `PLACE` appears among the rows `enumerate_graph` returns. *Instead:* `WORK` and `PLACE` are whole-program pins with no `@<axis>` key, and `PLACE` is a pin rather than a knob, so the enumeration never returns it at all. *Why:* **measured** — 158 pinned configurations over the offered space, and `PLACE` found only by hand afterwards. The deploy path routes around the limit by resolving each target separately; greedy does not, on the indirect evidence in the findings above. The direct test is deferred to the follow-up.
- [ ] **sm_120 goldens, and the identity re-seed #554 calls for.** *Expect:* this card carries recorded goldens for the model shapes. *Instead:* two synthetic-shape goldens from this PR's measurements and nothing else. *Why:* **known** — #554 orphaned kernel identities keyed on old tree shapes, and the re-seed has not been done.
- [ ] **Whole-model logits comparison.** *Expect:* `emmy run <model>` compares whole-model logits against torch. *Instead:* an arity mismatch stops it, and per-layer parity stands in. *Why:* **measured** — it reproduces on pristine main, so it is not this branch's.
- [ ] **Direct failing tests** for the warp tier's packed decline and for `match_packed_kblock_b`'s upper bounds (single packed load, single gather). *Expect:* each refusal has a test that fails when the refusal is wrong. *Instead:* they are covered indirectly through end-to-end shapes. *Why:* **not written yet.**

Two environment-class notes, neither a defect in this branch. **Twin memory** is ~47 GB host RSS for an 8B whole-dict load, which is the ceiling to raise before the 27B. And the suite's **session-end durations check** is a hard error that trips on GitHub's shared runner whenever a test drifts across its 5 s threshold; the tests it names are main's, it surfaces one or two at a time, and the failure reads as `N passed, 0 failed, exit 1`.

## The follow-up PR's charter

Named so this PR's scope is legible, and so the follow-up starts from a clean seam rather than half-wired machinery.

- **Serving-side W4A4 wiring** — per-layer-kind dispatch, the runner spelling 4-bit activations, the operand plumbing that makes serving run the native path.
- **The hybrid slice** — the linear-attention carve and the output-gate seam, removed here in one revertable commit (`b3056eea0`), picked up from there; the recorded bit-exact carve result is the ground it starts from, and the serving-side per-layer-kind dispatch plus DeltaNet cache wiring is the part that has to be built. Targets named in Known gaps: `Inferact/Qwen3.8-27B-NVFP4` as the lead candidate (the repack recipes.vllm.ai serves; its config was read, not run, and its exclude list leaves the DeltaNet layers unquantized), `nvidia/Qwen3.6-27B-NVFP4` as the official architecture-identical stand-in, `RadixArk/…` and `gittensor-model-hub/…` as alternates — calibration unvouched on every community repack.
- **The 27B slice**, once the twin memory ceiling is addressed.
- **The twins arm** — `serving/twins.py` has no NVFP4 arm, so `capture_gen_twins.py` produces float16 twins rather than the coded graph.
- **The evidence-path direct test** — does deployment-by-evidence reach per-kernel configurations that whole-program pinning cannot express?
- **Quantize producer-fusion** — the 522× spread itself: can the whole-program lane place the quantize the way per-target resolution does?
- **Closing the 2× to cuBLAS** on the block-scaled cell.
- **The parity-gated consolidations** — folding `_k_block_guard` into `_addr`'s `gmem_axis_step`, and sharing `_spell_static_fp4_decode` with `_spell_fp4_one`.

## AI;DR (AI; Didn't Read)

Everything below is the accumulated working record — the decision log, the per-card measurements, the merge history, and the reasoning behind choices a reviewer does not need in order to read the diff. It is kept because the arguments outlive their conclusions, and because someone arriving through `git blame` will want the argument rather than the outcome. Nothing here is required to review the change.

### Design decisions, in the order they were made

1. **The stored element is the packed byte pair.** `f4e2m1x2` is a dtype whose one stored element carries two logical values (`logical_elems == 2`); arrays stay byte-identical to the checkpoint. Rejected: one code per byte (2x memory, breaks checkpoint identity).
2. **Quantization is spelled, not typed.** Following the repo's fp8 design: bits, scales and decode become ordinary constants and ops in the graph at birth; no pass below the frontend knows checkpoint formats. A repo invariant test enforces the boundary.
3. **The two scale levels fuse to one f16 tensor.** e4m3 values are exact in f16, so the fusion rounds exactly once; the numpy oracle and the kernels share that single rounding story.
4. **The decode is a pair-table gather.** The packed weight indexes a 256x2 value table — an embedding lookup — rather than in-graph bit arithmetic (~3x the nodes for the same term; the lowering, not the graph, decides the emitted kernel code). The table binds as a computed constant (`source_graph`), since literal constants carry scalars only.
5. **Recognition is conservative and marker-first.** Checkpoint recognition requires the exact scheme (16-element groups; MXFP4's 32-element blocks decline). The kernel-side matcher identifies the packed operand by dtype marker and proves the block scale's k-invariance ((X + k)/B with X a multiple of B); any unproved shape declines to the compute fill, which computes the same values.
6. **W4A16 applies the scale in the drain, not the accumulator.** The mma lowering tier's fragments are f16/bf16 regardless, so multiplying the block scale into decoded values at the fragment load sits inside that tier's existing precision contract. Accumulator-side application belongs to native-fp4 operands (phase 4).
7. **Staging is three slabs.** A and the raw packed bits copy asynchronously; the block scales compute-fill into a small third slab. The bits slab indexes the checkpoint layout directly (`bits[n, k/2]`) — the dtype contract makes the layout canonical, so the decode cone's reshape arithmetic never reaches the kernel.
8. **The tuner distinguishes the packed stage without a new row stamp.** The search tree's candidate rows carried no signal that a byte slab moves a quarter of a 16-bit weight stream's bytes, so the prior priced both alike. A per-row stamp was the obvious fix and broke measurement attribution (the recorded-bench path recovers a row by its stamp digest), so the landed feature, `MMA_b_store_bits`, derives from facts already on the row. `FEATURIZER_VERSION` is unchanged and no prior was refit.
9. **The hybrid carves around a torch-side core.** For the linear-attention layers: pre = the input layernorm and the four input projections; core = causal convolution + gated delta rule + gated norm (torch/vLLM territory); post = the output projection, the post-attention layernorm and the MLP. For the full-attention layers: the fused query/gate projection's gate half rides the seam — a fourth pre output that post applies before its projection. Both carves reproduce their eager layers exactly (max_abs 0.0) on real 27B weights.
10. **Checkpoint-name translation reads transformers' own conversion registry**, in both directions: checkpoint keys to twin parameter names at load, twin paths back to checkpoint keys in the serving lane. Rejected: a hand-written prefix rule (serves one family; the registry serves every family with a mapping).
11. **One flag names the arch-suffixed target, not the instruction that wanted it.** `KernelSpec.uses_tma` used to request `sm_<cc>a`, because TMA was the only instruction family needing the suffix. The block-scaled fp4 mma needs it for an unrelated reason: ptxas assembles that instruction on `sm_120a` and refuses it on `sm_100a`. So the flag is now `arch_specific`, which names the need rather than either instruction. A plan stored under the old key still reads, so packs baked before the rename keep loading instead of quietly dropping to the unsuffixed target. Rejected: a second parallel flag (two flags, one question).
12. **Probes settled the scale-fragment layout, not the ISA text.** The PTX ISA documents the block-scale operands' thread-id selector as [0..3]. At `scale_vec::4X` ptxas accepts only [0..1]. The lane maps say which lane supplies which row's or column's scales. Probes on the hardware found them. The prelude records them beside the PTX it emits. A reader who trusts the documented range meets an assembler error rather than a wrong answer.

13. **The branch adopts main's #513 rather than working around it.** Main deleted every pass refusal taken for speed ("passes enumerate, evidence decides"), renamed the `STAGE` transport tokens, and removed fusion's policy guards. Merging it meant re-checking every refusal this branch makes. None of them refused for profitability. The TMA box limits and alignment are what the hardware accepts. The k-block guard declines a cone it cannot prove, and it declines *to a fork sibling*, which is already the shape #513 wants. The cc 12.x condition names the only architectures on which ptxas will assemble the instruction. They are legality facts and they stay. Two refusals did not survive as written. The matmul resolvers used to decline a split stage, and `Stage.split` no longer exists on main at all. Main also deleted `AtomKind.materialized_edges_only`. The fp4 atom used it to say no staged drain exists for a packed-plus-scales operand. That is a missing-code fact, so it moved to the stage resolver where that code would live.
14. **The block-scaled matmul op lands in this PR, not a follow-up.** Reaching the fp4 atom needs the warp tier's packed decline relaxed. That decline is legality today: every decode-based atom would read the wrong bytes. It is also the one rule on this branch that would otherwise sit in tension with #513's design. Keeping the op here means the branch never lands carrying that tension: the deviation and its resolution are the same PR. The one planned follow-up PR remains the serving-contract redesign — the runner rework that would serve the linear-attention layers too.
15. **A plan carries the graph dtype of every constant it binds.** Binding a constant needs a fact the checkpoint does not hold: whether the consuming node wants an fp8 tensor's RAW BITS or its decoded values. The graph-keyed loader reads it off the node; the plan-keyed one, which serving uses, had no such field and read everything at the stored dtype. `WeightSpec` therefore gains `graph_dtype`, written at plan construction, so a plan describes its own binding without needing the graph back. The alternative — having the caller pass the raw-bits paths from the graph — is a smaller diff that leaves a plan rebound from a saved pack still binding wrong. The field is deliberately not called `source_dtype`: `ConstantOp` already has that name for the STORED dtype, the opposite fact, and one helper is duck-typed over both types. Old packs omit the key, read `None`, and keep the historical behaviour.

16. **Decompose-and-recognize — no new matmul op.** Supersedes decision 14's vehicle while keeping its scope commitment: this PR does not build the dedicated block-scaled matmul op 14 named — a frontend op carrying (codes, scales) operand pairs with its own lowering — because a frontend op cannot survive decomposition (the repo's layering invariant) and the instruction's W4A4 program is not a schedule fork of today's matmul. Instead the speller writes the activation quantization into the graph as shared-vocabulary algebra at model-read time (`spell_static_fp4_activations`, beside the fp8 activation speller), and kernel-side RECOGNITION reaches the native path: a matmul whose both operands are pair-table-gather-times-blockscale decode chains over packed buffers lowers to the block-scaled atom. No format-named op survives decomposition — the repo invariant that killed the dedicated-op sketch stands, and `bind_bilinear`'s reserved both-computed slot is where the recognition lands.

17. **W4A4 is the declared numerical contract**, scoped to the matmuls the checkpoint marks (packed weights beside a stored per-linear `input_scale`). The graph itself declares the activation's quantize→dequantize round trip, so the program changes numerically — Σ x̂·ŵ, program-wide including decode — and the numpy backend computes the same declared program, staying the parity oracle for every lowering. Measured on the 8B's layer 0: the declared program sits at median 1.5e-3 / max 1.9e-2 from eager torch — the calibration's quantization delta. One inherent parity property came with it: behind a computed producer the two backends reach the encodes with epsilon-different upstream values, so a block whose scale ratio lands within that epsilon of a rounding boundary flips one code; parity there is distributional (median exact, rare bounded flips), where direct-feed comparisons stay tight.

18. **The block-scaled lowering carries a bounded gap against the declared program, rather than the declared program moving to meet it.** NVFP4 stores two scale levels — one e4m3 byte per 16-element block, one f32 per tensor — and decision 3 collapses them into a single f16 number per block ("the fused scale") because the 16-bit drain wants one multiply per block at its fragment's own dtype. The block-scaled instruction cannot take that number: it applies the raw e4m3 pair itself, per block, and leaves the per-tensor level to an output multiplier. So the declared `Σ a·f16(sa·s2a) · b·f16(sb·s2b)` and the emitted `s2a·s2b · Σ_blocks sa·sb · Σ a·b` are not the same expression — no reassociation connects them, unlike split-K and the mul-hoist, whose forks change rounding only. The gap is one f16 rounding of a per-block constant on each side, so it is bounded by roughly 2^-11 relative per side; the native path is therefore checked against numpy to a tolerance rather than as an exact oracle, and every other lowering keeps the exact check. **Follow-up worth trying: drop the f16 fusion on both sides**, spelling `code · from_f8e4m3(bits) · scale_2` with the levels apart, which makes the instruction's form an ordinary reassociation of a block-invariant factor and restores the exact oracle. It was costed at about two days and deferred: the quantize half divides by the fused scale, so unfusing moves the activation codes themselves rather than only the scale's rounding; the W4A16 drain's own loader types the scale at the fragment dtype and would need a second spelling; and every recorded W4A16 number would need re-measuring.
19. **A dtype's `logical_elems` is derived, not a stored field.** `DataType` is a frozen dataclass and `structural.form` renders a dataclass by walking its fields, so adding `logical_elems` as a field put it into the identity of every kernel whose body mentions any dtype — while discriminating nothing, since the value is fixed per dtype and the name is already the first thing `form` renders. The cost was recorded evidence going unreachable: 492 of the 859 realization rows in `recipes/*/golden/*.yaml` derived a different kernel key on this branch than on main, and nothing in the suite checks those files. Derived as a property over the packed-pair names it is invisible to the walk: `DataType`'s rendering returns to main's tuple for tuple, the recipe-golden drift falls to 130 rows, and those 130 are exactly what main plus this branch's `expr.py` index-arithmetic change produces on its own — the two trees agree row for row across all 859. The realization corpus reads the same from the other side: 17 stale cases become 2, and the 2 are the `imap-reshape` pair the index change legitimately moves (restamped; regeneration refused nothing, so both keep their `offered` and `realized` verdicts). Rejected: restamping all 17 and leaving the field, which fixes the corpus the suite does check and leaves the 127 contraction kernel names in the goldens it does not.
20. **A fusion cut releases whatever is left feeding only what departed.** The packed-buffer refusal (decision 16's boundary) drops the buffer's readers and their descendants, but left each consumer's own scale reconstruction behind — nodes feeding nothing but a matmul that had already gone. A cut materializes every buffer crossing it, so such a value is stored once and read once, at whatever shape it happens to carry: here the last node is a broadcast, and the boundary stored one f16 per logical element where its source held one e4m3 per 16-element block. That is what kept the block-scaled cell unreachable after the quantize hoist — the consumer's index then names k outside any block divisor, and a packed reading that cannot prove k-block invariance declines, so only the weight edge bound and the cell needs both. Releasing the survivor puts the reconstruction beside its matmul and lands the boundary on the raw block scales, which is the narrower buffer and the one a consumer can index block-wise; the activation edge then reads exactly as the weight edge does, and as it did before maximal fusion landed. Two nodes never leave this way: one writing the packed buffer itself, which is the boundary the refusal exists to place, and one read from outside the region, whose value has to be stored regardless. **This does not revise decision 3.** The two scale levels still multiply in f32 and round once to f16 — what moved is where that product is computed, not whether it rounds — so decision 18's bounded gap and its tolerance stand unchanged, and the follow-up 18 names (dropping the f16 fusion to restore the exact oracle) is still open and still costed the same. Rejected: teaching the matcher to see through a materialized broadcast, which would let a lowering coverage gap be repaired by widening recognition instead of by placing the boundary where the values live.

21. **The whole-program compile lane prices the activation quantize ~500× worse than the deploy path, and the pin grammar cannot correct it.** Measured on the 512×4096×4096 W4A4 linear: `k_reshape_reduce_e3e181` costs 7.6 µs when the working golden resolves it as its own target and 3970.6 µs when greedy compiles the program whole, where it fuses onto a grid of 2. An exhaustive prior-free sweep says no schedule pin recovers it: all **125** axis-scoped `REDUCE@a2/a3/a4` combinations the walk offers leave it at 3957–3975 µs and grid 2, and so do all **33** kernel-global `WORK` values (3974–3990 µs). Two structural reasons, both worth fixing separately from this PR. First, `WORK` and `PLACE` are KERNEL-GLOBAL as pins while the fork assigns them per kernel — greedy really does give the quantize `t4` and the matmul `w2x4` in one program — so a pin that helps one kernel drags the other with it: `PLACE=cut` takes the quantize from 3974 µs to 1420 µs and simultaneously costs the matmul its native cell (34.3 µs → 127.4 µs, worse than W4A16). Second, the sweep could not have found `PLACE` on its own: `enumerate_graph` returns knob rows, and `PLACE` is a PIN rather than a knob (`tests/compiler/realization/ARCHITECTURE.md` states the split), so any sweep driven by the enumeration has that blind spot by construction. What this does NOT block is deployment: the golden path already resolves per target and already reaches 7.6 µs, which is why the whole-linear number above is 42.1 µs and not 4012.8. The evidence that greedy is leaving its own per-kernel freedom unused is INDIRECT — greedy's clean-DB total (4012.8 µs) is worse than the best single global pin (1560.4 µs), which a fork exploiting per-kernel freedom should not be. The direct test — record these goldens, then re-run greedy with that evidence in place and see whether the quantize lands on its resolved-alone configuration while the matmul keeps the native cell — was deferred by choice, and is named in the serving follow-up's charter.

### Measured results (RTX 4090, sm_89)

Everything in this Measured-results section — the 4090 and 5090 tables, the serving runs, the decode-step profile — is the SCAFFOLDING stage's evidence: the W4A16 convert path, kept working but no longer the deliverable (see Summary).

Packed byte slab vs the compute fill — the generic lowering that evaluates the decode per element into shared memory — with identical pins except the stage, on Qwen3-8B linear shapes. M is the row count of the activation (tokens per step); the k/v row shows the lower and upper of its two same-shaped projections; `~` marks a pair's rounded midpoint.

| linear | N x K | M=4 | M=256 |
| --- | --- | --- | --- |
| q_proj | 4096x4096 | 2.91x | 4.31x |
| k_proj / v_proj | 1024x4096 | 3.18x | 3.79–4.01x |
| o_proj | 4096x4096 | 3.06x | 4.30x |
| gate_proj / up_proj | 12288x4096 | ~2.9x | ~2.0x |
| down_proj | 4096x12288 | 3.86x | 4.28x |

Parity against the decoded reference holds on every shape (f16 2.5e-4; bf16 2.1e-3 — the formats' own mantissa widths). An autotune run over one packed shape lands both candidate families in the tune DB, and greedy selection picks the packed row. On the real mixed-precision Qwen3.6-27B, both layer carves reproduce their eager layers at max_abs 0.0 against twin-loaded weights.

One fact the real checkpoint corrected: nvidia/Qwen3.6-27B-NVFP4 is mixed-precision. Its config places the attention and delta-net projections in fp8 and the MLP (`gate/up/down_proj`) plus `lm_head` in NVFP4; `in_proj_b`/`in_proj_a` are unquantized bf16 (verified on the checkpoint's own config groups and layer-0 tensors; its ignore list covers `mtp*` only). The MLP holds ~72% of layer weight bytes, so the packed kernel path carries the bulk of the weight traffic.

#### RTX 5090 (sm_120), -O3, CUDA-graph captured

The same packed-vs-fill comparison on the target card, same pins except the stage:

| linear | N x K | M=4 | M=256 |
| --- | --- | --- | --- |
| q_proj | 4096x4096 | 2.32x | 5.15x |
| k_proj / v_proj | 1024x4096 | 4.42x | 4.64x |
| o_proj | 4096x4096 | 2.37x | 5.09x |
| gate_proj / up_proj | 12288x4096 | 3.37x | 2.62x |
| down_proj | 4096x12288 | 4.63x | 4.63x |

Those rows are the W4A16 byte-slab reading against the generic compute fill — the SCAFFOLDING path's own efficiency, not the native one. The native path's efficiency is the table at the top of this description, and it is a different comparison: W4A4's block-scaled cell against W4A16's mma at the same shape, 34.5 µs vs 126.8 µs (3.8×) on the multiply and 42.1 vs 126.8 (3.0×) on the whole linear at 512×4096×4096. The two tables are not commensurable — one prices a staging choice inside W4A16, the other prices W4A4 against W4A16 — and the native one is the number this PR is about.

Sweeping M into prefill widths halves the advantage at M=512, after which it holds flat: q_proj runs about 5.2x at M=256, 2.60x at M=512 and 2.18x at M=4096. Two ratios say why. Achieved bandwidth falls to 2-5% of peak above M=256, so weight traffic stops governing — which is the regime the native fp4 atom targets. But the kernel plateaus at roughly 82-89 TFLOP/s, about a fifth of the card's fp16 peak, so at prefill the binding limit is not the tensor-core rate. Doubling the arithmetic rate on a kernel running at a fifth of peak need not deliver anything near 2x; the prefill gap deserves its own profile before anyone draws that conclusion.

#### Serving, end to end

**Working** (`a9ee223f9`). Serving keeps an NVFP4 trunk coded, and the numbers match. The *trunk* here means a model's dense per-layer weights, as against its routed expert weights. *Coded* means the loader hands the graph the checkpoint's packed trio — the e2m1 codes, their e4m3 block scales, the f32 per-tensor scale — rather than decoding it to a dense tile first. Measured on an RTX 5090 against `nvidia/Qwen3-8B-NVFP4`, one decoder layer, seeded input, with nothing patched — the runner deciding for itself. It chose `compress_trunk=True` and compiled 35 kernels, 4 of them carrying the packed byte-slab drain. Its post half matched eager torch at max|diff| 1.95e-3, rel 3.67e-4. The pre half's q/k/v agree with the decoded trunk bit-for-bit. Getting there took three defects. Each looked like the whole story at the time. **One:** a staged slab took 16-byte vector writes while declaring only its element's alignment, so an f16 slab could sit at 8 mod 16 — general, not NVFP4's, and it closed the same latent hazard in three attention kernels (`9d7c9e39c`). **Two:** the packed operand rebuilt its gmem address from the K-chunk offset. A split-K partition cuts the contraction axis across CTAs and hangs the slice's base on that axis, so rebuilding from the offset dropped it and every partition re-read the first slice's bytes. The block scales, which the fill evaluates from the decode cone's own body, read theirs correctly (`a0f61965d`). **Three, the double decode:** a plan reaches the checkpoint without its graph, so the plan-keyed loader read every key at its stored dtype and decoded the e4m3 block scales that the graph's own `from_f8e4m3` cone then decoded again (`9c72e0643`). An earlier attempt to turn this on (`5bb1cdc06`) lasted a day: parity showed it computing garbage, so `d8b5f77f6` took it back out. The refusal then stood until all three fixes landed. None of the three announced itself: every kernel compiled, every load succeeded, and the numbers were quietly wrong. Parity against eager torch is what caught each one, which is why it is now a pre-commit check for anything that changes what kernels compute rather than an item further down a list.

#### Speed: measured for the decoded trunk, not yet obtainable for the coded one

Decode-step profile on an RTX 5090, bucket 16, all 36 layers, runner-only (vLLM's attention excluded), same session for both:

| trunk | wall/step | GPU kernels/step | host + dispatch |
| --- | --- | --- | --- |
| decoded | 160.03 ms (6.2 tok/s) | 146.17 ms (91.3%) | 13.86 ms (8.7%) |
| coded | no number — see below | | |

The coded trunk has no figure because neither run could profile it. Two runs, capped at 50 and 90 minutes, both hit their timeout without reaching the timed-step phase, and the capture says why over 2483 repetitions: `[cuda] kernel '<whole-program e2e window>' still pending after 69.24s`. One program window runs past 69 seconds where the decoded path does all 36 layers in 160 ms. The runner's own roofline check names the same thing at compile time — `L0.post.decode.m16 runs 58431x over its roofline floor (~46 us)`.

This is a kernel-selection result, not a correctness one, and the two are cleanly separable here. Per-layer parity is independent of it and holds: the coded trunk matches eager torch at 1.95e-3 and agrees with the decoded trunk bit-for-bit on q/k/v. What is missing is tuning — nobody has tuned the packed serving twins on this GPU, so the deploy pick falls back to something far off the floor. **This branch therefore claims correctness for the coded NVFP4 trunk and claims nothing about its speed.** The golden re-seed in the gaps below is what stands between here and a usable number; an untuned pick landing 58431x off the floor is what #554's orphaning of kernel identities looks like from the serving side.

A historical figure of 147.8 ms/step appears in earlier notes for this model. It came from a different box, on code four syncs back, with a different kernel count, so it is a reference point rather than a control; the same-session decoded run above is the control.

### The serving decision: the full-attention slice first

The serving runner's per-layer seam is attention-shaped in several places — the two public pre/post signatures, the runner's per-layer metadata, the program tiers. Two paths were on the table. **The serving-contract redesign** threads a per-layer token-mixer kind through metadata, signatures and tiers, and serves every layer kind uniformly; it is the right end state and it is large. **The full-attention slice first** takes the half that fits the existing seam, and leaves the linear-attention layers to stock vLLM modules.

**The decision: the slice, and its machinery ships in the follow-up rather than here.** The Qwen3.5 family gives every fourth layer ordinary attention — 16 of Qwen3.6-27B's 64 — so the slice accelerates that quarter's token mixing and its MLP. This PR carries the compile-side evidence that the carve is exact, and none of the machinery: what was written here was never joined to a consumer, so removing it costs no working capability and leaves the follow-up a clean seam. The redesign remains the eventual answer; nothing in the slice forecloses it.

### Merged with main

`origin/main` moved 22 commits ahead of this branch, one of them architectural: **#513, "Remove performance gates: passes enumerate, evidence decides."** The merge adopts it rather than working around it. Decision 13 records the refusal review it forced. The study behind it read every commit in the gap. Four of the 22 touch a Python file this PR also changed.

The merge also fixed a bug this branch could not have fixed itself. Loop fusion was leaving down_proj's whole product materialized — an M x K x N buffer, 412 GB for an 8B at serving capacity — which is why NVFP4 serving died at engine init with a `cupy` out-of-memory for exactly that size. It is not an NVFP4 bug: an unquantized Qwen3-0.6B shows the same buffer, merely small enough to survive. Bisecting put the first good commit at #513, which removed fusion's three policy guards. The boundedness cap it kept (`_BLOWUP_FACTOR`) is not what left the buffer behind. A one-layer check now matches pristine main.

The full suite on an RTX 5090 after the merge: 3793 passed, 4 failed. Two of those four fail only because the box holds an rsync copy rather than a git checkout, one is the pre-existing tolerance case in the known gaps below, and the fourth was two stale transport pins in a test the merge auto-merged, fixed in a follow-up commit. A fresh 5090 box at the current tip now delivers sync #5's owed on-device verification: 3817 passed, 1 failed — and the one failure (`test_fused_causal_sdpa_split_partition_keeps_absolute_predicate_coordinates`, NaN output) reproduces on pristine main at the merge point on the same card, so it is main's sm_120 defect, not this branch's. The `test_quantized_checkpoint_e2e_cuda` tolerance case from the gaps below passes on that box — after sync #5, whose #613/#594 moved the f16/bf16 reference numerics; the old deterministic 0.0113 miss has not reappeared since.

Three mechanical consequences ride along. The merge renames the `STAGE` transport tokens (`sync`, `cp.async`, `tma` become `smem`, `smem-async`, `smem-tma`) through the packed byte-slab code and its pins. `AtomKind.materialized_edges_only` is gone, so the fp4 atom states that refusal where the missing code is. And model goldens moved to `recipes/<model>/golden/` (#537), while the model-agnostic hardware goldens stay in `emmy/compiler/pipeline/search/goldens/`.

Three further syncs followed. **#543/#544/#545** rebuilt state-merge internals across the same lowering files this branch extends; the packed byte-slab rows survived, and #545 restored the warp tier for the fused norm-into-linear shape a serving program compiles its projections into. **#547/#549** realized flash attention through the existing schedule vocabulary. Both sides had restructured how the A-side operand reaches shared memory, which is where they collided. This branch had extracted that decision into one helper so the packed path could share it. Main had added arms that fill the slab from a nested contraction on the tensor core. The merge keeps the helper answering which side of the transport A rides, and lets the caller swap in the chained fill. **#554** rebuilt recognition itself. The old design parsed a loop nest into scheduled terms while lifting it; the new one lifts any loop nest into a tree of `Fold` terms first, then runs classification as stages over that tree. That deleted `_atomize.py` outright. This branch's packed reading moved to a module of its own, `emmy/compiler/pipeline/passes/lowering/tile/_packed.py`, with its logic unchanged — `bind_bilinear` binds a packed weight as a computed-B projection exactly as its predecessor did.

**Sync #5** brought 73 commits. #629 derives a contraction's operand roles from the grid axes an operand's index mentions rather than from the order the trace emitted the arguments, which is the arm a packed weight already took — its decode cone is never a direct column-indexed load. #634 replaced the torch and tensor IR dumps with Rust-like pseudocode; the NVFP4 dtype and ops need nothing from it, established by putting the decode cone through the new printer at both stages. #580's retune and #526's serving-twin matrix leave the goldens question where it was, since no golden on main carries an NVFP4 shape. #628 and #636 refuse a cross-CTA structural tune winner whose parent pins cannot reproduce every measured child decision, which is the shape a split-K winner has. One resolution was a judgement call rather than a splice: main's `_needs_fill` now routes multi-channel products into the compute-fill block where the packed byte-slab dispatch sits, so that dispatch tests the node instead of the requested transport. Measurement then showed the mma resolver already declines a multi-channel node on its own, so the change hardens a predicate rather than repairing a wrong row; a test pins the refusal either way.

### Why these interim models

Qwen3.8-27B weights shipped mid-work (2026-08-15) with exactly the Qwen3.5-family geometry this branch targets. The OFFICIAL NVFP4 repack is still pending. Community repacks now exist: a modelopt mixed-precision one laid out like nvidia's 3.6, a pure-NVFP4 one, and a compressed-tensors variant. This branch defers them by choice rather than hitting a blocker, because 3.6 shares the architecture — everything proven here carries over, and 3.8 remains a checkpoint swap for whenever we want it. The ladder stands: nvidia/Qwen3-8B-NVFP4 (dense, pure NVFP4) proved the format machinery in isolation; nvidia/Qwen3.6-27B-NVFP4 (hybrid, mixed-precision) proved the architecture carves and the mixed loader; Qwen3.8-27B slots in as a checkpoint swap plus re-tuning when its repack lands.

### Unrelated fixes included

- `nbytes_of("long long")` raised instead of returning 8 (missing table entry; surfaced by a new whole-table consistency test).
- `_attn_aliased`'s docstring claimed the device-buffer aliasing is M=1-only; it holds at every width.
- The scalar-leaf codegen fix landed separately as PR #512 (merged).
- `SdpaOp`'s numpy forward filled masked scores by subtracting `mask * 1e9`; under NumPy 2 promotion the 1e9 casts to the scores dtype first (inf in f16) and `0 * inf` turned every VISIBLE position NaN, so the numpy reference of any masked f16 attention graph was all NaN. Masked positions now fill with -inf through `np.where`; f32 references agree to rounding.
- The reverse checkpoint renamer passed a `reverse=` keyword that only transformers 5.15 has, while this project pins `>=5.13,<5.15`. On a supported install it raised `TypeError` from both callers — the generation runner, and the constant retarget that `emmy compile` / `emmy run` reach on any quantized checkpoint. Dropping the keyword leaves the key mapping identical, since the call passes no converters and the flag only orders renamings against converters.

### Verification

CI is green on the final head. On an RTX 5090 with CUDA present, the full serial suite is **4235 passed / 1 failed**, the single failure being `test_freeze.py` reading a git-LFS pointer because that box is seeded from bundles and carries no LFS objects; `tests/compiler/e2e` is 371 passed and `tests/serving` 223 passed, both zero-failure. `scripts/digest_kernels.py` exits 0 with 23 of its 26 rows byte-identical to main; the three that differ carry identical kernel NAMES and differ only by the `__align__` this branch's `9d7c9e39c` puts on a slab aligned to its fill chunk rather than its element.

https://claude.ai/code/session_01Vgf2BkkXJcBWZvDwEgDSgG


gnull:
@Slonegg here are the currently produced IRs for layer 0 of `nvidia/Qwen3-8B-NVFP4`. It's what I got running `emmy compile nvidia/Qwen3-8B-NVFP4 --layer 0 --ir <ir>` on the current PR head [ec1b6f9](https://github.com/cloudrift-ai/emmy/pull/499/changes/ec1b6f9ee814fd9cdcf01846a2c4031fb05709fa).

- [torch.txt](https://gist.github.com/gnull/93de1337a4ca44d951c1f579de1f7bca/raw/0ec6842055d6187dc3f7e852d97795e0894a1c7e/torch.txt)
- [loop.txt](https://gist.github.com/gnull/93de1337a4ca44d951c1f579de1f7bca/raw/0ec6842055d6187dc3f7e852d97795e0894a1c7e/loop.txt)
- [tile.txt](https://gist.github.com/gnull/93de1337a4ca44d951c1f579de1f7bca/raw/0ec6842055d6187dc3f7e852d97795e0894a1c7e/tile.txt)
- [cuda.txt](https://gist.github.com/gnull/93de1337a4ca44d951c1f579de1f7bca/raw/0ec6842055d6187dc3f7e852d97795e0894a1c7e/cuda.txt)

Slonegg:
> @Slonegg here are the currently produced IRs for layer 0 of `nvidia/Qwen3-8B-NVFP4`. It's what I got running `emmy compile nvidia/Qwen3-8B-NVFP4 --layer 0 --ir <ir>` on the current PR head [ec1b6f9](https://github.com/cloudrift-ai/emmy/pull/499/changes/ec1b6f9ee814fd9cdcf01846a2c4031fb05709fa).
> 
>     * [torch.txt](https://gist.github.com/gnull/93de1337a4ca44d951c1f579de1f7bca/raw/0ec6842055d6187dc3f7e852d97795e0894a1c7e/torch.txt)
> 
>     * [loop.txt](https://gist.github.com/gnull/93de1337a4ca44d951c1f579de1f7bca/raw/0ec6842055d6187dc3f7e852d97795e0894a1c7e/loop.txt)
> 
>     * [tile.txt](https://gist.github.com/gnull/93de1337a4ca44d951c1f579de1f7bca/raw/0ec6842055d6187dc3f7e852d97795e0894a1c7e/tile.txt)
> 
>     * [cuda.txt](https://gist.github.com/gnull/93de1337a4ca44d951c1f579de1f7bca/raw/0ec6842055d6187dc3f7e852d97795e0894a1c7e/cuda.txt)

I don't quite see where the quantized scale is. Maybe ask AI to generate the diff between loop IR / tile IR of normal FP16 layer and NVFP4 layer.

gnull:
## The native fp4 path, dumped

Everything below comes from `emmy compile -c`, at `0f21108b5` plus the `--quantize` flag, rendered for `sm_120`. Two invocations: one whose activation feeds TWO linears, so loop fusion leaves its packed codes in memory, and one with a single linear, whose quantize stays inline and whose codes are compute-filled into the slab instead. Between them they cover both ways an operand's codes reach the cell.

`--quantize nvfp4` does not synthesize the decode algebra. It quantizes the traced weights, writes a real checkpoint (into `--dump-dir` when given, else a temp dir whose path it logs) and runs the ordinary spellers over it, so what compiles is exactly what compiling that directory would give. The weight side is derived from each tensor; the activation scale is calibrated over the trace's ONE example input, which is a real calibration and a poor one, and it is written into the checkpoint so the number can be read rather than assumed.

Full dumps, one file per stage: shared — [torch](https://gist.github.com/gnull/d3df988299226d508d92f915547173d8#file-oneliner-shared-torch-txt) · [tensor](https://gist.github.com/gnull/d3df988299226d508d92f915547173d8#file-oneliner-shared-tensor-txt) · [loop](https://gist.github.com/gnull/d3df988299226d508d92f915547173d8#file-oneliner-shared-loop-txt) · [tile](https://gist.github.com/gnull/d3df988299226d508d92f915547173d8#file-oneliner-shared-tile-txt) · [kernel](https://gist.github.com/gnull/d3df988299226d508d92f915547173d8#file-oneliner-shared-kernel-txt) · [cuda](https://gist.github.com/gnull/d3df988299226d508d92f915547173d8#file-oneliner-shared-cuda-txt); single — [torch](https://gist.github.com/gnull/d3df988299226d508d92f915547173d8#file-oneliner-single-torch-txt) · [tensor](https://gist.github.com/gnull/d3df988299226d508d92f915547173d8#file-oneliner-single-tensor-txt) · [loop](https://gist.github.com/gnull/d3df988299226d508d92f915547173d8#file-oneliner-single-loop-txt) · [tile](https://gist.github.com/gnull/d3df988299226d508d92f915547173d8#file-oneliner-single-tile-txt) · [kernel](https://gist.github.com/gnull/d3df988299226d508d92f915547173d8#file-oneliner-single-kernel-txt) · [cuda](https://gist.github.com/gnull/d3df988299226d508d92f915547173d8#file-oneliner-single-cuda-txt). Layer 0 of the 8B is in the [follow-up comment](#issuecomment-5427873943).

---

# A — a shared activation: codes reach the matmuls from memory

```
emmy compile --quantize nvfp4 --target sm_120 --ir loop -c "
class QK(nn.Module):
    def __init__(self):
        super().__init__()
        self.q = nn.Linear(512, 128, bias=False)
        self.k = nn.Linear(512, 128, bias=False)
    def forward(self, x):
        return self.q(x) @ self.k(x).transpose(-1, -2)
QK()(torch.randn(64, 512))"
```

Two linears read one activation, and the `q @ kᵀ` keeps them in separate kernels. The quantize half runs once: kernel 0 reduces each 16-element block to one e4m3 byte, kernel 1 encodes and packs.

```
=== 0: k_reshape_21b1bd ===
    in0 = load x_static_fp4_scale_2[0, 0, 0]
    v0 = multiply(in0, 6)
    v1 = reciprocal(v0)
    for a0 in 0..2048
        for a1 in 0..16
            in2 = load x[(a0 // 32), (((a0 % 32) * 16) + a1)]
            v2 = abs(in2)
            acc0 <- maximum(acc0, v2)
        v3 = multiply(acc0, v1)
        f8e4m3 v4 = to_f8e4m3(v3)
        x_static_fp4_scale_bits[(a0 // 32), (a0 % 32), 0] = v4

=== 1: k_reshape_pointwise_40ffbf ===
    in0 = load x_static_fp4_scale_2[0, 0, 0]
    for a0 in 0..64
        for a1 in 0..256
            in3 = load x_static_fp4_scale_bits[a0, (a1 / 8), 0]
            v0 = from_f8e4m3(in3)
            v1 = multiply(in0, v0)
            f16 v2 = copy(v1)
            f32 v3 = copy(v2)
            v4 = maximum(1e-12, v3)
            v5 = reciprocal(v4)
            in4 = load x[a0, ((a1 * 2) + 1)]
            v6 = multiply(in4, v5)
            v7 = reciprocal(v4)
            in5 = load x[a0, (a1 * 2)]
            i32 v8 = to_f4e2m1(v6)
            v9 = multiply(in5, v7)
            v10 = left_shift(v8, 4)
            i32 v11 = to_f4e2m1(v9)
            v12 = bitwise_or(v11, v10)
            f4e2m1x2 v13 = copy(v12)
            x_static_fp4_bits[a0, a1] = v13
```

Both linears then load those codes, and both bind the cell:

```
=== 0: k_reshape_21b1bd ===
=== 1: k_reshape_pointwise_40ffbf ===
=== 2: linear_1_reduce -> linear_1_reduce ===
    Fold[a3 in 0..256 ⊂ a2] contraction   ⟨TILE=mma_m16n8k64_e2m1_f32/f1x4/k4 STAGE=d1/smem-async⟩
=== 3: linear_reduce -> linear_reduce ===
    Fold[a3 in 0..256 ⊂ a2] contraction   ⟨TILE=mma_m16n8k64_e2m1_f32/f1x4/k4 STAGE=d1/smem-async⟩
=== 4: matmul -> matmul ===
    Fold[a3 in 0..16 ⊂ a2] contraction   ⟨TILE=f2⟩
```

Kernel 4 is the `q @ kᵀ`, which has no quantized operand and takes an ordinary register tile.

---

# B — a single linear: codes are computed into the slab

```
emmy compile --quantize nvfp4 --target sm_120 --ir kernel \
  -c "nn.Linear(512, 128, bias=False)(torch.randn(64, 512))"
```

One consumer means no fan-out, so fusion keeps the quantize inside the matmul and there is no buffer to copy. Three slabs copy — the weight's codes and BOTH block-scale slabs, because the cell takes the stored e4m3 byte itself — and the activation's codes slab is filled by the encode.

```
        RegFragment a:f4e2m1x2 _a0[4] (16x8x64)
        RegFragment b:f4e2m1x2 _b0[2] (16x8x64)
        RegFragment b:f4e2m1x2 _b1[2] (16x8x64)
        RegFragment b:f4e2m1x2 _b2[2] (16x8x64)
        RegFragment b:f4e2m1x2 _b3[2] (16x8x64)
        RegFragment c:f32 _c0_[4][4] (16x8x64)
        Smem unsigned char _a_smem[16, 128] align=16
        Smem unsigned char _b_smem[32, 144] align=16
        Smem __nv_fp8_e4m3 _as_smem[16, 16] align=16
        Smem __nv_fp8_e4m3 _bs_smem[32, 16] align=16
                cp.async[16B] _b_smem[((_fb * 16) / 128), ((_fb * 16) % 128)] <- p_weight_bits[((a2_b * 32) + ((_fb * 16) / 128)), ((((((a2_b * 32) +  …
                cp.async[16B] _as_smem[((_fas * 16) / 16), ((_fas * 16) % 16)] <- input_static_fp4_scale_bits[((a1_b * 16) + ((_fas * 16) / 16)), (((( …
                cp.async[16B] _bs_smem[((_fbs * 16) / 16), ((_fbs * 16) % 16)] <- p_weight_scale_bits[((a2_b * 32) + ((_fbs * 16) / 16)), ((((((a2_b * …
            for _fa in ((((a1_u * 1) + a2_u) * 32) + _lane)..128:32
            cp.async.commit_group
            cp.async.wait_group(0)
            Sync
                BlockScaleLoad _sfa0 <- _as_smem[((a1_u * 16) + 0), (_ki / 16)] (sfa, ldm=16)
                LdmatrixLoad _a0 <- _a_smem[((a1_u * 16) + 0), (_ki / 2)] (byte gather, ldm=128)
                BlockScaleLoad _sfb0 <- _bs_smem[((a2_u * 32) + 0), (_ki / 16)] (sfb, ldm=16)
                LdmatrixLoad _b0 <- _b_smem[((a2_u * 32) + 0), (_ki / 2)] (byte gather, ldm=144)
                BlockScaleLoad _sfb1 <- _bs_smem[((a2_u * 32) + 8), (_ki / 16)] (sfb, ldm=16)
                LdmatrixLoad _b1 <- _b_smem[((a2_u * 32) + 8), (_ki / 2)] (byte gather, ldm=144)
                BlockScaleLoad _sfb2 <- _bs_smem[((a2_u * 32) + 16), (_ki / 16)] (sfb, ldm=16)
                LdmatrixLoad _b2 <- _b_smem[((a2_u * 32) + 16), (_ki / 2)] (byte gather, ldm=144)
                BlockScaleLoad _sfb3 <- _bs_smem[((a2_u * 32) + 24), (_ki / 16)] (sfb, ldm=16)
                LdmatrixLoad _b3 <- _b_smem[((a2_u * 32) + 24), (_ki / 2)] (byte gather, ldm=144)
                MmaSyncPtx _c0_[0] += _a0 @ _b0 (m16n8k64 e2m1 scales:_sfa0,_sfb0)
                MmaSyncPtx _c0_[1] += _a0 @ _b1 (m16n8k64 e2m1 scales:_sfa0,_sfb1)
                MmaSyncPtx _c0_[2] += _a0 @ _b2 (m16n8k64 e2m1 scales:_sfa0,_sfb2)
                MmaSyncPtx _c0_[3] += _a0 @ _b3 (m16n8k64 e2m1 scales:_sfa0,_sfb3)
            Sync
            FragmentApply(_c0_[_r0] *= multiply(_c0_[_r0][], _bs_alpha1))
```

The emitted kernel, with the encode inside the fill and the instruction below it:

```cuda
                float in9__c14 = input_static_fp4_blk[((a1_b * 16 + _fa / 8) * 512 + (((a1_b * 16 + _fa / 8) * 512 + (a0 * 256 + (_ks + (16 * (_fa % 8) + 14) * 2))) / 16 % 32 * 16 + ((a1_b * 16 + _fa / 8) * 512 + (a0 * 256 + (_ks + (16 * (_fa % 8) + 14) * 2))) % 16)) / 2 % 256 * 2 + (a1_b * 16 + _fa / 8) * 512];
                float in9__c15 = input_static_fp4_blk[((a1_b * 16 + _fa / 8) * 512 + (((a1_b * 16 + _fa / 8) * 512 + (a0 * 256 + (_ks + (16 * (_fa % 8) + 15) * 2))) / 16 % 32 * 16 + ((a1_b * 16 + _fa / 8) * 512 + (a0 * 256 + (_ks + (16 * (_fa % 8) + 15) * 2))) % 16)) / 2 % 256 * 2 + (a1_b * 16 + _fa / 8) * 512];
                int in2 = input_static_fp4_shift[0];
                unsigned char v21__c0 = (emmy_to_f4e2m1((in9__c0 * (1.0f / v10__c0))) | (emmy_to_f4e2m1((in8__c0 * (1.0f / v10__c0))) << in2));
                unsigned char v21__c1 = (emmy_to_f4e2m1((in9__c1 * (1.0f / v10__c1))) | (emmy_to_f4e2m1((in8__c1 * (1.0f / v10__c1))) << in2));
        ...
            for (int _ki = 0; _ki < 256; _ki += 64) {
                unsigned _sfa0 = emmy_mma_load_sfa_f4(reinterpret_cast<const unsigned char*>(&_as_smem[a1_u * 16 * 16 + _ki / 16]), 16);
                emmy_mma_load_a_smem_b8v<unsigned char>(_a0, &_a_smem[a1_u * 16 * 128 + _ki / 2], 128);
                unsigned _sfb0 = emmy_mma_load_sfb_f4(reinterpret_cast<const unsigned char*>(&_bs_smem[a2_u * 32 * 16 + _ki / 16]), 16);
                emmy_mma_load_b_smem_trans_b8v<unsigned char>(_b0, &_b_smem[a2_u * 32 * 144 + _ki / 2], 144);
                unsigned _sfb1 = emmy_mma_load_sfb_f4(reinterpret_cast<const unsigned char*>(&_bs_smem[(a2_u * 32 + 8) * 16 + _ki / 16]), 16);
        ...
        #pragma unroll
        for (int _r0 = 0; _r0 < 4; _r0 += 1) {
            const int _g = (threadIdx.x & 31) >> 2; const int _t = (threadIdx.x & 31) & 3;
            atomicAdd(&linear[(a1_b * 16 + a1_u * 16) * 128 + (a2_b * 32 + a2_u * 32 + _r0 * 8) + _g * 128 + _t * 2 + 0], _c0_[_r0][0]);
            atomicAdd(&linear[(a1_b * 16 + a1_u * 16) * 128 + (a2_b * 32 + a2_u * 32 + _r0 * 8) + _g * 128 + _t * 2 + 1], _c0_[_r0][1]);
            atomicAdd(&linear[(a1_b * 16 + a1_u * 16) * 128 + (a2_b * 32 + a2_u * 32 + _r0 * 8) + (_g + 8) * 128 + _t * 2 + 0], _c0_[_r0][2]);
            atomicAdd(&linear[(a1_b * 16 + a1_u * 16) * 128 + (a2_b * 32 + a2_u * 32 + _r0 * 8) + (_g + 8) * 128 + _t * 2 + 1], _c0_[_r0][3]);
        }
}
```

No `EMMY_F4_LUT`, no per-element `__hmul`, no 16-bit fragment: the tensor core multiplies the 4-bit codes and applies the e4m3 block scales itself. The two per-tensor scale levels are the `_bs_alpha` product, applied once per output element after the K loop.

---

# C — what still does not reach the cell

Two shapes decline, and both reproduce as one-liners.

**A two-channel node.** Change `q @ kᵀ` above to a `cat` and the two linears fuse into ONE kernel with two channels sharing an A. The pair reading requires a single channel, so the node declines — and so does the W4A16 packed reading, which has the same restriction, meaning that shape has never been on any packed path. This is the `gate`+`up` kernel of a real layer:

```
emmy compile --quantize nvfp4 --target sm_120 --ir tile -c "
class GateUp(nn.Module):
    def __init__(self):
        super().__init__()
        self.gate = nn.Linear(512, 128, bias=False)
        self.up = nn.Linear(512, 128, bias=False)
    def forward(self, x):
        return torch.cat([self.gate(x), self.up(x)], -1)
GateUp()(torch.randn(64, 512))"
```
```
=== 2: k_linear_reduce_1cab9c ===
    ├─ operand[acc0]: Fold[a2 in 0..512] planar   ⟨REDUCE=coop⟩   ‹computed›
```

**An output whose trailing axes are not the matmul's rows and columns.** A projection that reshapes into heads gives `token x head x head_dim`, and contraction binding takes the trailing free-axis pair as `(m, n)` — so it reads head/head_dim and declines before any of this is consulted. That is `q_proj`/`k_proj` in a real layer, it is pre-existing and unrelated to NVFP4, and `loop/canonicalize/010_fuse_split_free_axes` exists for exactly that shape and is not firing.

Coverage on layer 0 of `nvidia/Qwen3-8B-NVFP4` — three of six matmul kernels — and the dumps for it are in the [follow-up comment](#issuecomment-5427873943).

---

## The measured claim

```
./venv/bin/pytest tests/compiler/passes/test_nvfp4_w4a4.py -q \
  -k "block_scaled_cell or binds_both_operands or reaches_its_matmuls"     # 3 passed in 50s
```

`test_the_block_scaled_cell_runs_and_holds_its_declared_tolerance` runs the native kernel against numpy on a DIRECT-FED activation — no computed producer, so both backends encode bit-identical codes — and holds median 1.8e-5 / max 5.6e-4 relative. That is the fused-scale gap of decision 18 and not the exact oracle; behind a computed producer the known boundary flips dominate instead, which is why the tight bound needs the direct-fed shape.

Full suite on the same box: 1 failed, 3825 passed, 51 skipped — the one failure being the sm_120 fused-causal NaN that reproduces on pristine main.

gnull:
## Layer 0 of `nvidia/Qwen3-8B-NVFP4`, dumped

The native fp4 path on a real model, at `0f21108b5` on an RTX 5090. `--layer 0` defaults to seq 512.

Full dumps, one file per stage: [tile](https://gist.github.com/gnull/d3df988299226d508d92f915547173d8#file-layer0-tile-txt) · [kernel](https://gist.github.com/gnull/d3df988299226d508d92f915547173d8#file-layer0-kernel-txt) · [cuda](https://gist.github.com/gnull/d3df988299226d508d92f915547173d8#file-layer0-cuda-txt) · [loop](https://gist.github.com/gnull/d3df988299226d508d92f915547173d8#file-layer0-loop-txt) · [tensor](https://gist.github.com/gnull/d3df988299226d508d92f915547173d8#file-layer0-tensor-txt).

### Which matmuls reach the cell

```
emmy compile nvidia/Qwen3-8B-NVFP4 --layer 0 --ir tile |
  awk '/^=== /{k=$0} /contraction|planar/{if (k ~ /linear/) {print k; print "    " $0; k=""}}'
```

```
=== 3: k_linear_reduce_dc5a93 ===
    Fold[a2 in 0..4096] contraction   ⟨TILE=mma_m16n8k64_e2m1_f32/f1x8/k8 STAGE=d1/smem-async⟩
=== 4: k_linear_0af9a4 ===
    ├─ operand[acc0]: Fold[a3 in 0..4096] planar   ⟨REDUCE=coop⟩   ‹computed›
=== 5: k_linear_cf0665 ===
    ├─ operand[acc0]: Fold[a3 in 0..4096] planar   ⟨REDUCE=coop⟩   ‹computed›
=== 9: k_linear_fcb6af ===
    ├─ operand[acc0]: Fold[a2 in 0..4096] contraction   ⟨TILE=mma_m16n8k64_e2m1_f32/f1x8/k8 STAGE=d1/smem-async⟩   ‹computed›
=== 13: k_linear_reduce_1828e5 ===
    ├─ operand[acc0]: Fold[a3 in 0..4096] planar   ⟨REDUCE=coop⟩   ‹computed›
=== 15: k_linear_d1ccdf ===
    ├─ operand[acc0]: Fold[a2 in 0..12288] contraction   ⟨TILE=mma_m16n8k64_e2m1_f32/f1x8/k8 STAGE=d1/smem-async⟩   ‹computed›
```

Three of the layer's six matmul kernels: 3 is `v_proj`, 9 is `o_proj`, 15 is `down_proj`. Still planar: 4 and 5 are `q_proj`/`k_proj`, whose output is token x head x head_dim so contraction binding reads the wrong axis pair as `(m, n)`; 13 is the fused `gate`+`up` kernel, a two-channel node the pair reading declines.

### The bound term

`v_proj`, with both operand edges packed decode chains — the activation's indexed by the row axis `a0`, the weight's by the column axis `a1`:

```
=== 3: k_linear_reduce_dc5a93 ===
    place  free=(a0, a1)  grid=(a0, a1)
    work   w4x1
    Fold[a2 in 0..4096] contraction   ⟨TILE=mma_m16n8k64_e2m1_f32/f1x8/k8 STAGE=d1/smem-async⟩
    ├─ operand[v9]: Fold  free  ‹pointwise›   ‹computed›
    │  └─ lift: λ() -> (v9)
    │       in0 = load mul_1_static_fp4_scale_2[0, 0, 0, 0]
    │       in2 = load mul_1_static_fp4_scale_bits[0, a0, ((((a0 * 4096) + a2) / 16) % 256), 0]
    │       v0 = from_f8e4m3(in2)
    │       v2 = multiply(in0, v0)
    │       in5 = load mul_1_static_fp4_bits[0, a0, ((((a0 * 4096) + ((((((a0 * 4096) + a2) / 16) % 256) * 16) + (((a0 * 4096) + a2) % 16))) / 2) % 2048)]
    │       f16 v5 = copy(v2)
    │       i32 v6 = copy(in5)
    │       in7 = load mul_1_static_fp4_r3_f4_pairs[(int)v6, (((a0 * 4096) + ((((((a0 * 4096) + a2) / 16) % 256) * 16) + (((a0 * 4096) + a2) % 16))) % 2)]
    │       v9 = multiply(in7, v5)
    ├─ operand[v8]: Fold  free  ‹pointwise›   ‹computed›
```

### Slabs: copied and filled

`v_proj`'s activation fans out to q/k/v, so loop fusion leaves its codes in memory and all four slabs copy:

```
        Smem unsigned char _a_smem[64, 272] align=16
        Smem unsigned char _b_smem[64, 272] align=16
        Smem __nv_fp8_e4m3 _as_smem[64, 32] align=16
        Smem __nv_fp8_e4m3 _bs_smem[64, 32] align=16
                cp.async[16B] _a_smem[((_fa * 16) / 256), ((_fa * 16) % 256)] <- mul_1_static_fp4_bits[0, ((a0_b * 64) + ((_fa * 16) / 256)), ((((((a0 …
                cp.async[16B] _b_smem[((_fb * 16) / 256), ((_fb * 16) % 256)] <- p_attn_v_proj_weight_bits[((a1_b * 64) + ((_fb * 16) / 256)), ((((((a …
                cp.async[16B] _as_smem[((_fas * 16) / 32), ((_fas * 16) % 32)] <- mul_1_static_fp4_scale_bits[0, ((a0_b * 64) + ((_fas * 16) / 32)), ( …
                cp.async[16B] _bs_smem[((_fbs * 16) / 32), ((_fbs * 16) % 32)] <- p_attn_v_proj_weight_scale_bits[((a1_b * 64) + ((_fbs * 16) / 32)),  …
```

`o_proj` has a single consumer, so its quantize stays inline and there is no buffer to copy. Three slabs copy — the weight's codes and BOTH block-scale slabs, because the cell takes the stored e4m3 byte itself — and the activation's codes slab is filled by the encode:

```
        Smem unsigned char _a_smem[128, 256] align=16
        Smem unsigned char _b_smem[128, 272] align=16
        Smem __nv_fp8_e4m3 _as_smem[128, 32] align=16
        Smem __nv_fp8_e4m3 _bs_smem[128, 32] align=16
                cp.async[16B] _b_smem[((_fb * 16) / 256), ((_fb * 16) % 256)] <- p_attn_o_proj_weight_bits[((a1_b * 128) + ((_fb * 16) / 256)), (((((( …
                cp.async[16B] _as_smem[((_fas * 16) / 32), ((_fas * 16) % 32)] <- reshape_static_fp4_scale_bits[0, ((a0_b * 128) + ((_fas * 16) / 32)) …
                cp.async[16B] _bs_smem[((_fbs * 16) / 32), ((_fbs * 16) % 32)] <- p_attn_o_proj_weight_scale_bits[((a1_b * 128) + ((_fbs * 16) / 32)), …
            for _fas in ((((a0_u * 2) + a1_u) * 32) + _lane)..256:512
            for _fa in ((((a0_u * 2) + a1_u) * 32) + _lane)..2048:512
        ... the encode, 16 cells wide, then one vectorised 16-byte store into the slab ...
```

### The emitted CUDA

```
emmy compile nvidia/Qwen3-8B-NVFP4 --layer 0 --ir cuda > /tmp/l0.cu
grep -c "emmy_mma_m16n8k64_e2m1_f32(_c" /tmp/l0.cu   # 24  the block-scaled instruction
grep -c "EMMY_F4_LUT"                    /tmp/l0.cu   # 0   no operand decoded through the value table
grep -c "emmy_mma_m16n8k16_f16_f32(_c"   /tmp/l0.cu   # 0   no 16-bit tensor core left in the layer
```

`v_proj`'s inner loop, eliding the repeated `_b1`..`_b6`. One A fragment against eight B fragments, so eight `mma` calls share one `_sfa0`:

```cuda
            for (int _ki = 0; _ki < 512; _ki += 64) {
                unsigned _sfa0 = emmy_mma_load_sfa_f4(reinterpret_cast<const unsigned char*>(&_as_smem[a0_u * 16 * 32 + _ki / 16]), 32);
                emmy_mma_load_a_smem_b8v<unsigned char>(_a0, &_a_smem[a0_u * 16 * 272 + _ki / 2], 272);
                unsigned _sfb0 = emmy_mma_load_sfb_f4(reinterpret_cast<const unsigned char*>(&_bs_smem[a1_u * 64 * 32 + _ki / 16]), 32);
                emmy_mma_load_b_smem_trans_b8v<unsigned char>(_b0, &_b_smem[a1_u * 64 * 272 + _ki / 2], 272);
        ...
                emmy_mma_m16n8k64_e2m1_f32(_c0_[0], _a0, _b0, _sfa0, _sfb0);
                emmy_mma_m16n8k64_e2m1_f32(_c0_[1], _a0, _b1, _sfa0, _sfb1);
```

No table lookup, no per-element multiply, no 16-bit fragment: the tensor core multiplies the 4-bit codes and applies the e4m3 block scales itself. The two per-tensor scale levels ride the epilogue, applied once per output element after the K loop.

Slonegg (approved):

PR #505, opened 2026-08-14
Optimize Llama FP16 kernels on V100

Slonegg:
## Summary

- keep loop fusion maximal and target-independent; every sensible materialization boundary is exposed as a scheduler placement fork
- integrate the PR #504 two-level architecture: outer search chooses scoped `PLACE` rows and inner search independently schedules every resulting kernel
- persist exact placement and schedule transcripts, structural fingerprints, per-child CUDA rows, and fresh end-to-end costs; stale plans fail loudly
- automatically split verified working plans into canonical PLACE-only routing and schedule-only Loop-child records, while requiring honest whole-plan and child references
- repair FP16 contraction semantics, multi-channel transports, typed reduction boundaries, reshape-aware addressing, atomic split legality, materialized flash masks, symbolic geometry, and split-K placement
- add the pinned single-V100 FP16 serving envelope for Llama 3.1 8B at 65,536 tokens

## Placement contract

- fusion remains maximal and has no target, model, or operation-name branches
- placement offers `PLACE`, `PLACE@a`, `PLACE@map`, `PLACE@broadcast`, `PLACE@nested`, and `PLACE@product` forks where structurally applicable
- option 0 preserves maximal fusion; routing evidence or an explicit placement pin selects a cut
- hardware capabilities constrain schedule legality only; they do not hardcode a placement decision
- working-golden replay records the complete unpooled decision trajectory and every child schedule

## Current validation

- pre-placement Llama serving inventory: 2 frontend programs, 6 context-complete Loop programs, 24 M=1/M=32/M=512/dynamic realizations
- product plus additive reduction remains one outer tuning target; no expanded product tensor is exported as a standalone tuning ABI
- 71 exact replay/schema/promotion tests passed
- 112 placement/recognition/fusion tests passed, with 1 expected xfail
- 104 cross-platform compiler tests passed off GPU
- fresh GPU validation: 7 SM120 cases and 3 V100 cases passed, including reshape-aware MMA addressing, dense-copy legality, atomic split, biased flash, and SDPA compact-placement replay
- repository-wide Ruff check and format check pass; fusion implementation remains unchanged
- latest CI: lint passes; 3,449 tests pass and the sole failure is the intentionally absent canonical Llama V100 golden

## Preliminary V100 diagnostics

These numbers predate the final placement-fork replay and are retained only as a regression baseline:

- pre M=1: Emmy 87 us vs eager 245 us, strict pass
- pre M=512: Emmy 957 us vs eager 1274 us, strict pass
[...]

PR #504, opened 2026-08-14
Separate fusion, placement, and schedule search for H200 tuning

Slonegg:
## Summary

- Keep loop fusion deterministic and maximal, independent of placement, schedule, hardware, and search evidence.
- Add a post-recognition `PLACE@path=fuse|cut` structural fork owned by the outer two-level tuner; schedule every resulting recognized kernel independently in the inner tuner.
- Preserve recognized Tile IR across cuts, separate placement lineage from per-kernel schedule evidence, and exclude graph-splice fragments from ordinary predecessor timing evidence.
- Add exact multi-kernel tune-winner persistence and replay: placement, component cache keys and multiplicities, directly searched schedule pins, and ordered CUDA-row integrity data.
- Add a generic materialized-score flash schedule: workspace score → C-fragment load → online softmax/C-to-A repack → tensor-core P@V, including split-KV realization.
- Repair generic packed Q/K recognition, lexical closure, multi-channel sync staging, multi-output eager reference checks, and nested-only schedule-site materialization.
- Enforce the ownership invariant with static and dynamic tests: fusion cannot depend on target, placement, schedule, or search; recognition is algebra-only; schedule and lowering use named capabilities and atom metadata.
- Strengthen compiler, IR, pipeline, pass, kernel-lowering, command, and test architecture documentation around fusion → placement → schedule ownership.

Fusion rules and fusion tests are intentionally unchanged.

## H200 result

Targeted Qwen3-0.6B layer-0, seq=512 target `k_sdpa_mean_linear_reduce_28c822.` on one NVIDIA H200 141GB. The winner was searched at deploy optimization (`--nvcc-flags ""`), persisted as one exact kernel set with `PLACE@dd=cut`, and replayed in five independent strict processes with 10 warmups and 100 measured iterations.

| Metric | Median | Five-run range |
| --- | ---: | ---: |
| Eager PyTorch whole forward | 653.01 us | 652.82–653.12 us |
| Exact Emmy winner, captured whole program | 270.76 us | 270.40–271.05 us |
| Speedup | **2.4116x** | **2.4085–2.4154x** |

Every run has `strict.status=pass`, one `pinned[]` winner with `status=ok`, no flags, direct eager correctness at `rtol=atol=1e-3`, five captured launches, and `timing_semantics=whole_program_e2e`. The kernel-set manifest and ordered source-hash inventory are identical across all five processes. The new P@V residue is about 18.3 us plus a 6.0 us finalize; the tuned QK producer is about 210.5 us.

[...]

PR #508, opened 2026-08-14
Unify placement as a measured structural fork (supersedes #504, #505)

Slonegg:
## Summary

Supersedes #504 and #505 — one branch carrying both PRs' functionality under a single layer contract:

1. **Fusion** stays greedy-maximal, deterministic, and algebra-only.
2. **Recognition** is algebra-only and total over what it accepts.
3. **Placement** (`_cut.py`) is the single owner of fuse/cut: with no pin or routing golden, every recognized
   root offers one structural fork — the maximal fused form plus one option per realizable seam. GPU product
   identity lives in exactly one function (`_routing_entry`), selecting measured evidence. Enforced by the AST
   pass contract tests.
4. **Schedule** enumerates the capability-legal domain via named `Context` predicates; enumeration does not rank.
5. **Policy** resolves every fork through the deploy evidence hierarchy; capability-derived cold leads rank rows
   when no evidence exists.

## What each superseded PR contributes

**From #504** (via its original head, merged, plus ports from its rebased head `305b3de9`): the `PLACE` structural
fork, `Op.decision_knobs`, the hardware-capability contract tests, the two-level outer/inner boundary and
measured-winner assembly; ported fixes — nested-SSA sync-fill suffixing, nested-only TILE ownership, strict-run
row accounting, accuracy size-mismatch failures, `PLACE_cut_site_<path>` features, and the fail-closed
`no_exact_pin` tune marker with stale-winner supersede.

**From #505**: the FP16 numeric contract (f32 virtual product, typed reduction boundaries, f32 split-K
workspace, mixed-radix index rejoin, N-channel emitters), the exact placement/schedule plan persistence
(`replay_plan` + promotion + `tune --verify-replay-plans`), and its V100 capability knowledge — reborn as the
**capability-derived placement cold lead** (deploy tier 4): a contraction whose computed operand no
target-available atom consumes inline leads the option that cuts that edge, so a cold sm_70 deploy recovers the
hardware contraction tier. Pins, routing goldens, and trusted Σ-pricing sit above it; measured evidence overrides.

**Not carried over** (recoverable from the closed branches): #504's `ranking.kernel_set` persistence
(superseded by `replay_plan`) and its materialized-score flash lowering, which is entangled with a
`lowering/schedule` pass-package split this branch does not adopt; #505's deterministic in-`_cut.py` capability
cuts (forbidden by the fork contract — replaced by the cold lead) and its V100 serving envelope, whose golden
was never captured.

## Defects found and fixed during two-GPU validation

- `bind_contraction_channels` ran every contraction through the group path, whose shared-epilogue check rejects
  the k-invariant mul-hoist — every fp8-B matmul lost its warp tier (4 tests recovered).
- A trace-native multiply+reduce pair (decomposed matmul) kept its full pre-reduction product materialized —
  **120 GB of f32 for one 512-token Llama MLP block**. The product-reduction recomposition now accepts
  nested-accumulation producers when the workspace is statically wider than every edge it bridges, the
  nested-reduction lift leads the cold option order as a work repair, and a just-lifted workspace never re-merges
  (the lift and the recomposition are inverse rewrites).
- Cold option order hoisted ordinary value-seam cuts above the fused base whenever any repair existed.
- A selecting fold's f16 carrier under a cross-thread combine mixed `__half` with the combine's f32 tree — nvcc
  rejected the kernel (V100 seq-512 attention).
- Reconstituted/spliced collapse loops carried no kernel name; two anonymous `__partial` kernels collided in one
  program's kernel table and launched one function with the other's arguments (intermittent segfault in the
  serving boot audit).

## Verification

- **Local RTX 5090**: `make test` — remaining failures are pre-existing on `origin/main` on this box (verified
  per-test against a pristine main worktree: serving-generation vLLM drift, TMA-reshape, quant/EXL3 e2e — the
  EXL3 one roots in dotted checkpoint buffer names rendered into a kernel signature) or pass-in-isolation
  ordering flakes. No failure introduced by this branch; four fp8 failures and the device-prefill serving test
  were *fixed* relative to the merge base.
- **V100 (8× SXM3 32 GB)**: `make test` fully green — 3460 passed, 0 failed (exit code reflects only the
  5090-measured durations baseline). Whole Llama 3.1 8B block compiles, runs, and passes accuracy vs eager at
  seq 1 and seq 512.
- Cold V100 perf without goldens trails eager (one cold gate-up kernel is 87% of block time) — as designed;
  the deploy hierarchy is fully cold on that card. Capturing the tuned V100 inventory (and with it the serving
  envelope) is the follow-up, per `plans/placement-fork-unification.md`.


Slonegg:
The three standalone correctness fixes in this branch (sync-fill SSA suffixing, nested TILE ownership, strict-run accounting) are extracted to #510 so they can land independently. They were cherry-picked with `-x`, so once #510 merges this branch rebases over them cleanly.

PR #550, opened 2026-08-19
Swizzle every slab the fused flash kernel reads

Slonegg:
## What

The fused single-kernel flash arm at `D = 128` was **shared-memory bank-conflict bound**, not tile bound. Three of its four slabs drained plain row-major, and the swizzle the fourth used under-spread a row wider than its atom. This swizzles all four and fixes the derivation.

Supersedes this PR's original scope (recording the A100 measurements) — the table it recorded no longer describes the kernel.

## Why the old reading was wrong

The plan said the wall was the tile: 254 registers, four slabs, one block per SM, 16% compute. ncu on an A100 says otherwise — **78% of the kernel's shared wavefronts were conflict replays**, L1/TEX at 80% against 7% compute, ncu's block limit from registers was 2, and local (spill) traffic was negligible.

## The three fixes

Each measured on its own, `(1, 8, 1024, 128)` f16, `w4x1`/`f1x16/k8`:

| change | us |
| --- | ---: |
| baseline | 108.7 |
| the score's staged key + query slabs swizzle on their `cp.async` fill | 84.8 |
| the weight tile swizzles on its FRAGMENT store (`RegStore.swizzle`) | 64.9 |
| the XOR shifts by the slab's own ROW stride, not the swizzle atom | 50.1 (best schedule) |

The third is why the wall was `D = 128` **specifically**. The XOR's shift is `log2(atom elems)`, which reads the row index only while a slab row IS one 128 B atom. A 128-elem fp16 head-dim row is two atoms, so consecutive rows collapsed onto a quarter of the chunk positions. ncu measures 7.8-way at `D = 128` and **no load conflict at all** at `D = 64` on the same kernel — that was the decisive experiment.

Hardware (TMA) slabs keep the atom shift: the copy engine fixes their permutation and their descriptor already splits its box down to the atom. A row that is one atom keeps the plain spelling, so **every slab that already drained conflict-free renders byte-identically**.

## Result — A100-SXM4-80GB, f16, `(1, 8, S, 128)`, best schedule pinned

| S | torch | fused | |
| --- | ---: | ---: | --- |
| 256 | 13 | **9.0** | `w4x1`/`f1x8/k8` |
| 512 | 24 | **19.0** | `w4x1`/`f1x16/k8` |
| 1024 | 48 | 50.1 | `w8x1`/`f1x16/k8` |
| 2048 | 133 | 179.0 | `w4x1`/`f1x16/k8` |

Beats torch at `S ≤ 512`, parity at `S = 1024`, still 1.35x behind at `S = 2048` — and the residual grows with `S`, so it is not a fixed overhead. L1/TEX is down to 61% and compute up to 24%; the next lever (the P round trip, which FA-2 keeps in registers) is recorded in the plan.

The COLD pick is untouched by this and still lands badly: unpinned it takes `f1x8` at every shape, which happens to be best at `S = 256` (9.0) and is 231 us at `S = 512`. That is the schedule-evidence problem the docstring already documents.

## A bug found on the way

The per-cell `RegStore` rewrite rebuilt the node field by field, so it silently dropped the new `swizzle` — leaving the weight tile stored row-major under a drain that reads it XOR-permuted. A **wrong kernel with no loud failure**, and the same trap had already cost `atomic` once. It is a `dataclasses.replace` now, which cannot regress that way when a field is added.

## Verification

- `make test` on the A100: **3550 passed**. The 7 failures are pre-existing on this box (fp8 needs sm_89+; two are environment) and reproduce identically on a pristine `HEAD` checkout.
- Every number above is `emmy run --bench --strict` (correctness checked against eager), `EMMY_NVCC_FLAGS=` (O3), empty tune DB and prior. Emmy's timings repeat to ±0.2 us across runs; torch's drift a few percent, more at `S = 2048` (132-148 observed).
- New tests: the swizzle-shift derivation (`software_swizzle` / `swizzle_xor` spellings, incl. the byte-identity cases), and a schedule-independent assertion that every slab in the fused kernel is addressed through ONE mode — all uses XOR or none do, which is exactly what the `RegStore` field-drop broke.
- The 5090 table in the docstring predates this and has NOT been re-measured; it is marked as such.
- `ruff check` / `ruff format --check` clean.

### DIFF
--- a/emmy/compiler/ir/kernel/ir.py
+++ b/emmy/compiler/ir/kernel/ir.py
@@ -23,7 +23,7 @@ Tile IR and are materialized away before reaching this layer. A
 
 from __future__ import annotations
 
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from functools import cached_property
 
 from emmy.compiler.dtype import F32, DataType
@@ -479,7 +479,7 @@ class CpAsyncCopy(Stmt):
     def pretty(self, indent: str = "") -> list[str]:
         smem_idx = ", ".join(e.pretty() for e in self.smem_index)
         src_idx = ", ".join(e.pretty() for e in self.src_index)
-        swz = f" swz={self.swizzle}" if self.swizzle in LDMATRIX_SWIZZLE_XOR else ""
+        swz = f" swz={self.swizzle}" if swizzle_xor(self.swizzle) else ""
         return [f"{indent}cp.async[{self.nbytes}B] {self.smem}[{smem_idx}]{swz} <- {self.src}[{src_idx}]"]
 
     def render(self, ctx: RenderCtx) -> list[str]:
@@ -487,8 +487,8 @@ class CpAsyncCopy(Stmt):
 
         smem_flat = render_index(self.smem, self.smem_index, ctx)
         src_flat = render_index(self.src, self.src_index, ctx)
-        if self.swizzle in LDMATRIX_SWIZZLE_XOR:
-            smem_flat = f"emmy_swizzle_{self.swizzle.lower()}({smem_flat})"
+        if swizzle_xor(self.swizzle):
+            smem_flat = f"{swizzle_fn(self.swizzle)}({smem_flat})"
         pad = _pad(ctx.indent)
         # ``emmy_cp_async_{cg,ca}`` (the cp.async prelude) does the ``cvta`` internally, so this is a
         # single call — no ``_smem_addr`` local and no wrapping ``{ }`` block. .cg is 16-byte-only.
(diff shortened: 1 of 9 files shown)


Slonegg:
Rebased onto `main` (`a3a65b79`, which brings #551's pipeline-strategy refactor and #552) — clean, no conflicts.

Re-ran `make test` on the A100 after the rebase: **3243 passed**, same 7 pre-existing failures as before (fp8 needs sm_89+; two are environment), no new ones. The pass count drops from 3550 only because #551 consolidated the search/pipeline suites (~6.7k lines of tests removed on main); this branch's own diff to `tests/` adds tests and removes none.

`ruff check` / `ruff format --check` clean on the rebased tree.

PR #558, opened 2026-08-20
Reverify DeepSeek V4 Flash 0731 on 16x V100 SXM3, and fix what blocked it

6erun:
Reverifies the maintained DeepSeek V4 Flash 0731 recipe on 16× V100 SXM3, and fixes the three defects that reverification turned up. Every number below was measured on the supplied host.

## Serving

The recipe qualifies: 24/24 requests over three steady repeats, 0 failures, 30.98 ± 0.09 output tok/s, mean TPOT 201.4 ± 0.5 ms. An exact 1,048,576-token request returned HTTP 200. Tool calling and separated reasoning both verified — reasoning separation is opt-in per request via `chat_template_kwargs`, which the report now documents.

Short-context throughput is ~9% below the 2026-08-11 qualification on the previous host, concentrated almost entirely in prefill (TTFT +48.7%, TPOT +2.8%). Both runs are internally stable, but host and driver changed together, so the report states the delta without attributing it.

Two stale recipe comments corrected: the pinned image *is* published, and the configuration is not fully zero-JIT (eight Triton kernels compile once during the first warm-up).

## Three defects fixed

**A fresh 16× V100 host could not deploy at all.** Every vLLM worker died at `cudaGetDeviceCount()` with `error 802: system not yet initialized` while `nvidia-smi` listed all 16 GPUs. The GPUs sit behind 12 NVSwitches and NVIDIA Fabric Manager was never installed — nothing in `provision_remote` installed it. Added as the last provisioning step, gated on a non-empty `/proc/driver/nvidia-nvswitch/devices/`, pinned to the running driver's exact version resolved from `apt-cache madison` (Ubuntu and NVIDIA publish different revision suffixes).

**`emmy run --golden` re-parsed the whole inventory once per target.** At ~15 s per load for this 3.4 MB, 279-target golden, the replay spent over an hour re-parsing before doing useful work. `_run_golden_targets` already holds the parsed document, so it now hands it to each target. Full replay: **4 min 19 s**. Separately, `validate_golden_file` rebuilt a program graph per *config* though only 13 distinct programs exist — deduped, validation drops 4.5 s → 1.5 s.

**A realized cross-CTA `REDUCE` split was reported as unrealized.** Pinning `REDUCE=g2a` always failed the realized-vs-pinned gate — `"REDUCE=g2a realized (off)"`, row left unbenched — even though the split did happen: the emitted CUDA halves the K loop from 16,384 to 8,192, adds the partition axis and closes with `atomicAdd`. #539 made a split mint brand-new kernels and had `knob.consume_kernel_row` strip their schedule row, so no piece may carry the `g<n>` it came from (`test_split_fresh_kernels` asserts exactly that). The receipt is the piece's sliced reduce axis, exactly as a realized `PLACE` cut is structural. The gate now skips that stage the way it already skips `PLACE`, and keeps gating the rest of the value. Emitted CUDA is byte-identical — codegen untouched.

## Compiler golden

Coverage was re-derived from the model config and confirmed by tracing seven layers: all 43 decoder layers reduce to **three** distinct graphs (`compress_ratio` 0 / 4 / 128, at 945 / 1,156 / 1,087 IR nodes). Layers 41–42 (`dspark_target_layer_ids`) trace identically to their ordinary siblings, and the old golden's fourth representative was redundant. The MTP head (`num_nextn_predict_layers: 1`) is **not** traceable — it is not exposed as a decoder layer — and the golden does not contain it.

An equal-budget hybrid-vs-MCTS search ran over the nine realizations that had headroom (same inventory-only base, empty DB/prior, separate cubin caches, `--max-candidates 8 --patience 4 --seed 731`, MCTS first). The O1 ranking lane misranks this model badly, so all conclusions come from repeated deployable O3:

| Target | greedy | MCTS winner | hybrid proposal |
| --- | ---: | ---: | ---: |
| `model-seam.k_linear_db1eb0` | 2,855 / 2,856 µs | 2,805 / 2,511 µs | **1,015 / 902 µs** |
| `layer0.i003.k_linear_reduce_f6a146` | **910 / 910 µs** | 2,526 / 2,520 µs | 2,368 / 2,487 µs |

One target promoted: `REDUCE=coop, WORK=t128` on the model-seam projection, 2.8–3.2× faster than greedy. With the gate fixed, the eight `k_linear_reduce_f6a146` rows benched for the first time since #539 (16/16 runs, no unreproducible-pin error) and were re-measured; `g2a` ties with greedy on all eight, so their configuration stands and only the stale 2026-08-11 measurements needed refreshing.

Across the golden, realizations materially slower than greedy fall **41 → 11**, and the eleven that remain are sub-microsecond pointwise kernels. The file repository-validates, strictly decodes 279/279, and replays every target on sm_70.

## Known limits

- The exemption above means a `g<n>` pin that genuinely never split can no longer be told apart from one that did — knob stamps carry no evidence either way. Documented in `emmy/compiler/pipeline/ARCHITECTURE.md`.
- Tuning ran against the committed inventory, not a freshly re-derived one: three representative traces ran 4h45m in the Loop splicer (`_ensure_dep` under `build_merged_region`) without emitting post-fusion targets. Coverage is proven at the graph level, not the post-fusion target level. A whole-model trace is unbounded here — it passed 830 GiB RSS before being stopped.
- The A/B was scoped to the nine targets with headroom; a full 279-target arm measured out at ~10 h per arm. It is labelled a scoped kernel result, not a whole-model performance claim.
- On a Volta host, `torch 2.13` pulls `nvidia-cuda-nvrtc 13.0.88` and `cupy-cuda12x` prefers it; CUDA 13 dropped sm_70, so every CuPy JIT path fails with `invalid value for --gpu-architecture`. Worked around with `LD_PRELOAD` of the CUDA 12.9 NVRTC. Not fixed here — tracked separately.

## Testing

`make test` cannot complete on macOS/arm64: `tests/compiler` collection segfaults (exit 139), reproduced on a clean tree at the base commit. Everything else passes locally (1,077 passed, 76 skipped), `tests/compiler/passes` + `ir` pass file-by-file (785 passed, 0 failures), and the full suite including `tests/compiler` was run on the Linux V100 host.

### DIFF
--- a/emmy/provisioning/remote.py
+++ b/emmy/provisioning/remote.py
@@ -25,6 +25,7 @@ async def provision_remote(
        (reboots and waits for the host to come back if anything was installed)
     4. Install NVIDIA Container Toolkit if not found
     5. Add user to docker group
+    6. Start NVIDIA Fabric Manager if the host has NVSwitches
     """
     dry_run = bool(getattr(host, "dry_run", False))
 
@@ -73,6 +74,10 @@ async def provision_remote(
 
     # 5. Add user to docker group
     await host.run("groups | grep -q docker || sudo usermod -aG docker $(whoami)")
+
+    # 6. Start Fabric Manager on NVSwitch hosts
+    if not skip_nvidia and not dry_run:
+        await _ensure_fabric_manager(host)
 
 
 async def _ensure_nvidia_versions(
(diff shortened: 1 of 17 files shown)


emmy/provisioning/remote.py
```diff
@@ -160,6 +165,55 @@ async def _ensure_nvidia_versions(
     return installed
 
 
+async def _ensure_fabric_manager(host: Host) -> None:
```
Slonegg, line 168: Maybe this belongs to rift provisioning

Slonegg (approved):
Not sure about fabric manager changes; the rest seems fine.

emmy/provisioning/remote.py
6erun: It's a fix for one of issues explained in the description:
```
A fresh 16× V100 host could not deploy at all. Every vLLM worker died at cudaGetDeviceCount() with error 802: system not yet initialized while nvidia-smi listed all 16 GPUs. The GPUs sit behind 12 NVSwitches and NVIDIA Fabric Manager was never installed — nothing in provision_remote installed it. Added as the last provisioning step, gated on a non-empty /proc/driver/nvidia-nvswitch/devices/, pinned to the running driver's exact version resolved from apt-cache madison (Ubuntu and NVIDIA publish different revision suffixes).
```

Slonegg: Yes, and that is something that should be installed in the VM image itself by default, instead of customer code patching half baked image.

6erun: But Emmy is not only for our platform, so maybe it will be useful for people launching it elsewhere?

Slonegg: I believe NV FabricManager is installed and running by default on all NVLink-enabled machines.

PR #563, opened 2026-08-20
Give onboarding the same shared-prompt setup as discovery

Slonegg:
## Summary

`onboard-model` now follows the same structure as `discover-models`: tracked prompt files own the run policy, and the
workflow renders only a compact task payload.

- **New shared prompts** — `prompts/onboard-model/qualify.md` is the canonical non-interactive contract (task-field
  table, per-mode recipe policy, caller-owned boundaries, image and bounded-fix investigation policy, artifact/LFS
  rules, summary contract). `prompts/onboard-model/investigate.md` is the bounded `onboard-investigator` prompt,
  mirroring `score-recipes.md` and `size-deployments.md`.
- **Workflow** — the ~60-line Python heredoc that built a prose prompt is replaced by a `jq -n` task object
  (`schema_version`, mode, model, exact target, SSH handles, deadline, publication flag, summary path, expected
  lifecycle). Both prompts are attached from the exact workflow SHA alongside the three skills and `model-fit.md`.
- **`OPENCODE_CONFIG_DIR`** is now pinned to the workflow source, as discovery already does, so the onboarding agent
  definitions come from the commit that started the run rather than whatever sits on the rolling branch. This is the
  only behavior change beyond relocation; the artifact worktree still stays on the rolling branch.
- **Skill and agents** — `SKILL.md` gains a "Step 0 — Choose the run mode" section pointing at `qualify.md`, and its
  delegation paragraph hands the investigator `investigate.md`. `onboard-investigator.md` shrinks to an
  "apply the supplied prompt" role like `discover-scorer`/`discover-fit`.

## Testing

- `make test` — 2947 passed, 578 skipped
- `make lint`
- Two existing tests that asserted the prose inside the workflow script now assert it in the prompt files, plus a new
  `test_onboarding_agent_reads_shared_prompts_from_a_compact_task`.
- Rendered the new `jq` step under representative env values and checked the fragment with `bash -n`.

### DIFF
--- a/.agents/skills/discover-models/SKILL.md
+++ b/.agents/skills/discover-models/SKILL.md
@@ -130,8 +130,11 @@ In automated lifecycle mode, follow `prompts/discover-models/lifecycle.md`: invo
 once and in parallel, then invoke `discover-scorer` once per deterministic recipe batch and in parallel. Each source
 investigator stays read-only, uses at most three public-web calls, and returns a compact source-specific candidate
 list. Each scorer follows `prompts/discover-models/score-recipes.md` without doing more research or choosing lifecycle
-state. The parent alone reconciles identities, compares scores globally, selects the maintained set, and proposes new
-onboarding models.
+state. Invoke `discover-fit` once per onboarding model — every new candidate and every existing onboarding shell —
+in parallel; each sizes exactly one checkpoint under `prompts/model-fit.md` and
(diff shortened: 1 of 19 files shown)


Slonegg:
Superseded by #565, which carries the same change rebased onto main after #562 merged.

PR #587, opened 2026-08-22
Stop recording a split piece as a rival of the whole op it came from

trifonova-n:
The write-side half of the 22 221× regret #586 turned up. Independent of that PR — different files, branched off
`main` — but the same defect seen from the other end.

## The defect

A bench row carries two structural facts that arrive by different routes:

- its **pool key**, `op_sig`, digested from the **pre-descent offer site** (`bench_record.py:25` says so
  explicitly: "NOT the terminal kernel's");
- its **features**, taken from `main.knobs` — the kernel that actually ran.

Nothing made them agree. A fork point realized as **several kernels** is where it breaks. The grouping in
`bench_leaves` exists to sum those into one leaf — but it can only sum kernels that resolved to the same fork point,
and one piece's own `source_chain` can bottom out at the un-split one while its sibling's does not. The piece is
then recorded alone, as a rival of the whole op.

## What it was, concretely

All **nine** affected pools on the RTX 5090 freeze are the same thing: a **fused `rms_norm`→linear megakernel**
beside a row for **one kernel of the same op's unfused realization**.

| row | structure | latency |
|---|---|---|
| piece | `rsqrt=1, reduce_add=1, loads=5` — the RMSNorm kernel | ~6 µs |
| sibling | `rsqrt=0, reduce_add=1, loads=2` — the matmul, under its **own** `op_sig` | ~180 µs |
| whole | `rsqrt=1, reduce_add=2, loads=6` — the fused megakernel | 28–212 ms |

In four of the nine the sibling was recorded too, so the unfused pair's real cost is visible: **24–191 µs** against
the fused row's 28–212 ms. In the other five the sibling was not recorded at all. Every fused row is **scalar tier**
(`WORK: t*`, no `TILE`) and every piece is warp/mma — `S_warp_eligible` is set on the piece and absent on the whole.

A part's latency is not the realization's cost, which is what makes the row wrong regardless of the magnitudes.

## The fix

`bench_leaves` compares the recorded kernel's `S_ext_*` extents against its fork point's and drops the row when they
differ, logging what it dropped and why.

**Extents are the detector, not a work measure.** Tiling and split-K rewrite the loop nest but preserve the extents,
so a kernel realizing the whole fork point runs the fork point's extents. Their *product* is not a work measure for
these kernels — a sweep rides the reduction, so it double-counts, which is exactly why `implausible_value_reason`
abstains from its own `free × reduce` formula on this stamp shape. What a mismatch proves is that the row does not
account for its fork point, and that is enough.

The check is on `main` — the op whose knobs get recorded — so it does not fire on the legitimate summed multi-kernel
group, where `main` is the kernel realizing the fork point and the fragment beside it is what gets summed in.

**Dropped rather than re-keyed.** The key such a row *should* carry is its own fork point's, and if its source chain
could name that we would not be in this situation. A wrong row silently deleted is recoverable by re-tuning; a wrong
row silently re-keyed is not.

`bench_fail` rows are screened the same way — a negative example filed under a pool whose work it never attempted is
just as wrong as a positive one.

## What this does NOT fix

The guard stops a wrong row; it does not produce the right one. The unfused realization still gets **no leaf of its
own**, because its kernels never resolve to a common fork point for the summing to find. So the tune's data still
cannot express the comparison these nine pools were groping at — and on this evidence that comparison matters a
lot: unfused looks two to three orders of magnitude faster on these shapes, and every fused row lands on the scalar
tier. Worth its own investigation; noted in `plans/offline-prior-scale-and-collection-redesign.md`.

## Coverage

`bench_leaves` had **no tests**. It has five now: the ordinary single-kernel row, the split piece being rejected, the
`bench_fail` path, the summed multi-kernel group the check must *not* fire on, and two distinct sites staying two
rows.

## Relationship to #586

#586 keys the *reader* (`group_measured`) on the extents as well, and both are needed: this stops the writer
producing such rows, that stops the reader trusting the ones already written — every existing tune DB and the v3
freeze included. Neither supersedes the other and they touch no common file.

https://claude.ai/code/session_01KXPHwChs61iSChSCddZJoG

### DIFF
--- a/emmy/compiler/pipeline/ARCHITECTURE.md
+++ b/emmy/compiler/pipeline/ARCHITECTURE.md
@@ -1159,6 +1159,16 @@ is what stops measurements from a manual sweep evaporating.
   loop dialect that carries `S_*` features, and falls back to the deepest one in the tile dialect. The mma
   tile-lowering keeps no `LoopOp` in `.source`, so without that fallback every tensor-core kernel was silently
   unrecordable. Both paths digest to the same `op_sig` a tune would write, verified on an RTX 4090.
+- **A row's key and its features must describe the same work**, or the pool it lands in is not a comparison. The two
+  arrive by different routes — the key from the fork point, the features from the kernel that actually ran — and a
+  structural fork can break the correspondence: when it splits a site, one piece's own `source_chain` may still bottom
+  out at the un-split site, so the piece gets filed as a rival of the whole op. The recorder compares the kernel's
+  `S_ext_*` extents against its fork point's and drops the row when they differ. Extents are what settles it because
+  tiling and split-K rewrite the loop nest but preserve them, so a kernel that realizes the whole fork point runs the
+  fork point's extents. Measured on the RTX 5090 freeze written before the check, nine pools held such a pair: a 5.9 µs
+  row scored against a 131 ms one, both honest at about 35 TFLOP/s, a median 8192x apart in work. Consumers cannot
+  repair this after the fact, which is why the reader (`data/group.py`'s `group_measured`) also keys on the extents —
+  the writer stops producing the rows, the reader stops trusting the ones already written.
 - The kernels of one variant (a split-K main kernel plus its combine kernel) are grouped under one fork point and
   recorded as ONE leaf for the whole variant. If every kernel in a graph loses its fork point, the recorder warns
   loudly rather than silently recording nothing. Rows that were flagged (a pin that did not match, a wrong answer, an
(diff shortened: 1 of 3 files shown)


trifonova-n:
Superseded by the grouping-key change in #586, and now the wrong fix.

This PR **drops** a bench row whose extents do not match its offer site's. That was the safe move while a
measured pool was keyed on `op_sig`, because such a row would land beside configs it never competed with.

#586 keys a measured pool on the **kernel's own `S_*` signature** instead — what `Identity.op_sig` computes for
an op, asked of the kernel that ran. Under that key the row this PR discards is a perfectly good measurement of
its own kernel and lands in the right pool by itself. Dropping it would lose real data for no remaining benefit.

The genuinely open item is the opposite of this patch: have `bench_leaves` **key the leaf by the kernel that ran**
rather than by the site, so the writer and the reader agree. That touches the `op_sig` column the tune engine
keys its search tree on, so it needs a real tuning run to validate — its own piece of work, not this one.

Recorded as finding 20 in `plans/offline-prior-scale-and-collection-redesign.md`, along with the unresolved
question behind it: on these shapes the fused `rms_norm`→linear form lands on the scalar tier and looks two to
three orders of magnitude slower than not fusing, which nothing currently records a comparison for.

PR #613, opened 2026-08-24
Compute decomposed SiLU in PyTorch opmath dtype

Slonegg:
## Summary

- evaluate decomposed SiLU in PyTorch's opmath dtype: f16/bf16 widen to f32, while f32/f64 retain their dtype
- keep one explicit widening conversion and one result conversion through Loop IR
- document and test the precision contract, including f64 no-demotion

## Verification

- `make test`: 3,107 passed, 575 skipped, 1 xfailed
- `make lint`
- exact Tesla V100-SXM3-32GB (CC 7.0) strict O3 standalone replay: zero numerical error; Emmy 1.696 us versus
  eager 3.563 us (2.10x)
- rendered CUDA widens once with `__half2float`, evaluates negative/`expf`/denominator/reciprocal/product in f32,
  then narrows once with `__float2half`
- exact Laguna-S-2.1 seeded expert replay: finalizer output now matches the independently f32-computed SiLU/product
  exactly (353 differing elements to 0), its strict intermediate proof passes, and downstream propagation through
  the eager projection passes

## Evidence boundary

The complete Laguna expert replay remains a non-terminal 1/3072 strict failure. The remaining error is isolated to
the independently replayed downstream Emmy projection kernel (one f16 ULP, max absolute error 0.00390625), not the
SiLU/product intermediate fixed here. This PR does not claim complete Laguna qualification or performance parity.


Slonegg:
Combined-overlay follow-up evidence (not attributable to this PR alone):

- Preserved the original exact seeded Laguna replay where this SiLU change fixes the intermediate but the scalar
  downstream row remains 1/3072 strict-negative.
- With this PR plus the unmerged Accum-boundary and Volta m=1 structural overlays, exact current-main inventory exposes
  15,412 legal downstream rows (15,392 MMA; 20 scalar).
- Replaying one exact existing row
  (`WORK=w1x4,TILE=mma_m8n8k4_f16_f32/f1x4/k4,REDUCE=,STAGE=d2/smem,RASTER=`) on the identical seeded inputs is strict
  correct end to end: downstream and whole Laguna expert both pass, max absolute error 0.001953125.
- The downstream kernel improves from about 205.0 us to 79.75 us; the whole three-launch stage improves from about
  308.2 us to 182.2 us (1.69x).

The downstream schedule and its enabling structural work are not part of #613. This comment establishes that the
remaining mismatch is schedule/accumulation-order specific and records a correct proposal for the later cold,
equal-budget hybrid-vs-MCTS comparison; it does not widen #613's implementation or claims.

PR #621, opened 2026-08-24
Return every Torch reference output

Slonegg:
## Summary
- return graph outputs in declared order from the Torch reference evaluator
- preserve the existing tensor return for single-output graphs
- cover a semantic final plus live intermediate output

## Motivation
Laguna exact-source qualification found 52 output-count mismatches: CUDA returned every declared Graph output while the Torch reference returned only the first.

## Verification
- make test: 3129 passed, 575 skipped, 1 xfailed
- make lint: green
- exact affected V100 replay pending before ready/merge

### DIFF
--- a/emmy/compiler/backend/ARCHITECTURE.md
+++ b/emmy/compiler/backend/ARCHITECTURE.md
@@ -36,7 +36,9 @@ FP8 tensors remain exact `uint8` bit carriers; `to_f8*` casts to torch float8 an
 `from_f8*` performs the inverse reinterpretation before widening. `is_runnable(graph)` is `True` only when every
 compute op and every elementwise operation name has a mapping. Data-dependent `GatherOp` / `ScatterOp` remain
 unsupported, so `run --ir` falls back to emmy-only benchmarking for those graphs. `build_callable(graph,
-input_tensors)` returns a pure `fn(*tensors)` (scalar constants read inline) so `torch.compile` can trace it. Symbolic
+input_tensors)` returns a pure `fn(*tensors)` (scalar constants read inline) so `torch.compile` can trace it. A
+single-output graph returns its tensor directly; a multi-output graph returns an ordered tuple following
+`graph.outputs`, matching the backend result boundary. Symbolic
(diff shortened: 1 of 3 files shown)


Slonegg:
Exact 8xV100 transport proof on the rebased head is complete.

- source: `9c080ed23a2b58f56f1f25841b6277c8efa86e31`, tree `e4466674d593860f154d9e0dc12d0d871698327d`, clean task checkout
- hardware: all eight caller-provided `Tesla V100-SXM3-32GB` devices, CC 7.0, with per-shard UUID receipts
- input set: all **52** exact source occurrences (**16** distinct source/program identities) that previously ended with `output count differs`; the manifest is byte-identical to the original selection (`sha256 cb1a94953e93ad977ced3d1d348dee27e720c58ebd0290c8a47c5eb21d941b82`)
- command boundary: `emmy run --ir ... --bench --strict --warmup 2 --iters 5 --bench-backends eager,tcompile,emmy`, default deterministic seed 0
- isolation: 8/8 fresh shards recorded absent `EMMY_CUBIN_CACHE`, `EMMY_TUNE_DB`, and `EMMY_ONLINE_FILE` before launch; 8/8 fail-closed import receipts resolve `emmy`, `commands.run`, `_bench_worker`, and `backend.torch_ref` beneath the exact checkout
- result: **52/52 JSON receipts, 0 output-count recurrences**
- resulting strict categories: 8 passes (2 identities), 26 numerical failures (8 identities), 18 shape failures (6 identities). Those failures are now observable through ordered multi-output transport and remain separate compiler gaps; this PR does not claim to fix their numerics or shapes.
- preserved evidence: `/home/riftuser/.cache/emmy/onb-laguna-agentic.hj9sGn/artifacts/pr621-9c080ed2-output-count-v3.86RInR` (`audit.json` mechanically rechecks every receipt)

Two earlier attempts are explicitly excluded: v1 used a wrong module name in its provenance check; v2 began shard fanout before the newly required one-time explicit import preflight. Only v3 above is promotion evidence.

PR #627, opened 2026-08-24
Fix standalone Loop input boundaries

Slonegg:
## Summary

- register every retained `InputOp` as an external input when slicing one Loop kernel
- remove the narrower two-level workaround that only repaired fragments with no registered inputs
- document and regress the standalone slice contract for partially registered minting fragments

## Why

A heterogeneous placement cut can mint a fragment whose `Graph.inputs` contains one external edge but omits another
retained `InputOp`. The old workaround only ran when the list was empty, so a standalone route-child replay planned the
omitted boundary as scratch and failed because it had no producing launch.

## Verification

- `./venv/bin/pytest tests/compiler/pipeline/search/ -p no:randomly -n auto --dist=loadgroup -q` — 125 passed
- `make test` — 3135 passed, 576 skipped, 1 xfailed
- `make lint`
- `git diff --check`

### DIFF
--- a/emmy/compiler/pipeline/ARCHITECTURE.md
+++ b/emmy/compiler/pipeline/ARCHITECTURE.md
@@ -860,7 +860,8 @@ only (`tile → kernel → cuda`), and returns the Σ once ALL Loop kernels are 
 - The slice keeps the root kernel + its leaf-op closure and turns every other kernel-input into a synthetic `InputOp`.
   The root op is shared **by reference**, so its body — and thus `Op.cache_key` — is byte-for-byte the full-graph op's.
   It filters the graph's canonical topological order rather than iterating its set-backed ancestor closure, so slice
-  inputs and persisted Loop programs stay byte-identical across fresh Python processes.
+  inputs and persisted Loop programs stay byte-identical across fresh Python processes. Every retained `InputOp` is
+  registered as a slice input even when a minting fragment did not list the boundary in its own `Graph.inputs`.
 - Because the inner tree holds one op, MCTS explores only that op's forks with `patience` as the op's own budget —
   `Σ_k n_k` benches total, never the product.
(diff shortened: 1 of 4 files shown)


Slonegg:
Exact post-rebase A100 proof is terminal on head `e966ae29` / main `becdc4cc`.

- Hardware: `NVIDIA A100-SXM4-80GB`, cc 8.0, UUID
  `GPU-b0354a1a-37c2-086d-f6fe-953b6fac5c3e`, nvcc 12.9.
- Raw task: `/home/riftuser/.cache/emmy/a100-kernel-pr627-e966ae29-proof-001`.
- The child wire was regenerated from the exact `PLACE@map=cut` splice through this head's
  `single_node_graph`, SHA-256 `677f13e51f434d2cd2a00646e9b4c90653270c2476a645d7fd907b63d1b6c4d1`.
- Fail-closed planner gate: every declared child `InputOp` is an `input` in the CUDA execution plan; child 2's only
  scratch buffer is its produced `mul_12__partial` with live interval `(0, 2)`.
- The parent route ran once to capture its actual deterministic input/intermediate boundary. Both standalone children
  then replayed those same inputs for five fresh O3 benches through `bench_pinned_async`.
- Child 1: 5/5 strict `same-input-parent-route` passes, one launch, 1.65-1.66 us.
- Child 2: 5/5 strict passes, two launches, about 9.54-9.61 us.
- All ten rows were CUDA-graph captured with max/mean/max-relative error exactly zero. Driver status 0; shared A100
  lock released and independently free.

The prior ffb diagnostic's five child-1 failures were only a harness shape mismatch: the parent debug snapshot stores
the scalar as `(1,)`, while the standalone graph declares `()`. The authoritative harness reshapes that captured
storage to the declared `Graph.buffer(output)` shape, asserts the backend-decoded output has that same declared shape,
and then compares; values were already bit-exact.

Post-rebase CPU gates so far: focused 19 passed / 5 skipped, `make lint`, and `git diff --check` are green. The exact-bec
full suite is running separately.

PR #653, opened 2026-08-25
Check in the measurement freeze and evaluate the prior against it by default

trifonova-n:
> Stacked on #652 (step 8). Branched from it, so review that one first; this rebases onto `main` when it merges.

A reported prior number should be one someone else can reproduce. It wasn't.

## The default was the one thing that can't be checked

`emmy eval prior --dataset nodes` resolved `--db` to `~/.cache/emmy/autotune.db` — machine-local, rewritten by
every tune, and gitignored. The online prior's reservoir is worse for the purpose: a bounded random sample that
Algorithm R churns as tuning continues, so one model evaluated twice **on one machine** need not score the same.

Meanwhile the freeze was already built to be the durable artifact:

```
sha256          26a7e2f3427c1c2ce9a434f9e5262d265c13bd513b6ca9aa0a68d405b67806b0
freeze_ver 3 · feat_ver 3 · knob_ver 3 · encoding_ver 3
counts          3960 rows · 3817 ok · 143 bench_fail
repo_commit, run_ids, created_at, source_db
```

Digest-pinned, version-stamped, identical row-for-row wherever it's read. It just wasn't the one evaluation
reached for — you had to know to pass `--db _tune/autotune-freeze-v3-rtx5090`.

| | persistent | reproducible | portable |
| --- | --- | --- | --- |
| **freeze** | digest-pinned, versioned | yes | yes |
| tune DB | mutable; re-tuning rewrites it | no | machine-local |
| reservoir | reservoir-*sampled*, bounded, churns | no | machine-local |

## What lands

**One freeze checked in** at `emmy/compiler/pipeline/search/freezes/` — RTX 5090, v3, 3960 rows — beside the
`goldens/` it mirrors in kind.

**`config.freeze_path()`**, following the existing `tune_db_path` / `online_path` shape: `EMMY_FREEZE_DIR` →
the checked-in directory. `--db` still points at a tune DB or another freeze when you want one machine's data;
it is now the override rather than the default.

**The freeze's identity in the report header** — `sha256` plus the four version stamps — so two reports *declare*
whether they are comparable instead of leaving it assumed. Absent for a live DB, which has no such identity;
that absence is itself the signal.

```
[prior] nodes dataset — …/search/freezes
  sha256=26a7e2f3…, freeze_ver=3, feat_ver=3, knob_ver=3, encoding_ver=3, rows=3960, groups=336
```

`emmy eval prior --dataset nodes` now runs with no arguments at all.

## Storage

**Payload YAML in git LFS**, manifest in plain git. The YAML is 3.75 MB / 150k lines and a re-freeze would
otherwise bloat every clone; the manifest is 30 KB and staying diffable is the point — a re-freeze should show its
digest, counts and run ids changing in review. The repo already uses LFS for `experiments/**/results*.tar.gz`.

Staged correctly — the tracked YAML is a 3-line pointer:

```
version https://git-lfs.github.com/spec/v1
oid sha256:1d861c5d28ed4097dc2ceee069d6806a3525beb44d4bc2045cc1ee844a346bd5
size 3750871
```

**Deliberately not wheel package-data.** A wheel install never fits or evaluates a prior, and this is several MB
per card. `pyproject.toml` says so at the `goldens/*.yaml` entry so nobody "fixes" the omission.

## Tests

- `test_the_checked_in_freeze_is_the_default_evaluation_corpus` — the shipped freeze loads, its `feat_ver` matches
  the live featurizer, its digest is well-formed, and its rows are leaf-only. Costs ~3 s because loading
  **re-derives every row's features with live code** rather than trusting the file, which is the property that
  lets a freeze outlive a featurizer change. Worth paying once: if this breaks, every reported prior number is
  uncomparable and nothing else in the suite would say so.
- `test_prior_defaults_to_the_checked_in_freeze` — the resolution and the `EMMY_FREEZE_DIR` override, without
  loading, so the CLI suite stays fast.

## Verification

- `make test`: **3211 passed**. The one failure is the pre-existing, environmental
  `test_neptune_emmy_runner_fails_when_one_setup_is_incomplete` (its shell runner needs GNU `timeout`, absent on
  macOS; CI is Ubuntu). Present on `origin/main`.
- `make lint` clean.
- `emmy eval prior --dataset nodes` with no arguments: 336 pools, same figures as the explicit `--db` invocation.

## Note on this particular freeze

Its manifest records `repo_commit: c126c4b4…-dirty`. It is the freeze every measurement in the recent prior
analysis was computed over, and its `sha256` pins the content regardless — but the provenance stamp is honest
about the tree being dirty when it was cut. Worth re-cutting from a clean tree next time a card is swept.

https://claude.ai/code/session_01KXPHwChs61iSChSCddZJoG


trifonova-n:
Superseded by #654 — reopening against `main`.

This PR was stacked on #652's branch. #652 squash-merged into `main` first, after which merging this one applied its commits to the superseded branch rather than to `main`. GitHub marks that MERGED, but the content never reached `main`:

```
$ git ls-tree origin/main --name-only emmy/compiler/pipeline/search/freezes/
(empty)
```

Same commit, cherry-picked onto `main` in #654, verified green there.

PR #662, opened 2026-08-26
Host DeepSeek V4's fork attention in the vLLM plugin and gate it through a real distributed engine

6erun:
Stage 3 of the DeepSeek V4 serving plan: `EmmyGenModel` serves the hyper-connection architecture through the
pinned 1Cat fork's own attention sublayer, verified up through a REAL vLLM engine at TP2×PP2.

## What lands

- **Fork attention hosting** — one engine-owned `DeepseekV4Attention` per local decoder layer (shared indexer
  buffer, shared aux streams, absolute-layer prefixes so every SWA/compressor/indexer cache registers its own
  KV-cache spec), carrier-width pipeline transport, and a loud refusal instead of the silent numpy fallback.
  There is no q/k/v seam in this architecture: the sublayer is handed the normalized hidden states and returns
  the attention output.
- **Config resolution past vLLM's re-registered config class** — inside a vLLM process, `AutoConfig` returns
  vLLM's own rope-only stand-in for this model type (registered process-wide by its config parser), which drops
  every derived field (`layer_types` above all). Both serving lanes now reload with Transformers' same-named
  native class when the resolved class is foreign.
- **The attention ownership table** — every `layers.N.attn.*` key of the rank's interval routes through the fp8
  `.scale` sibling rename, the compressor's `mla_attn` placement, and the two fused stackings (the indexer's
  inner compressor included) into the fork modules' own weight loaders; `head.weight` is the published spelling
  of the head; the runner-owned trunk/expert/embedding families need no claim; the MTP head serves no twin.
  Enforced loudly both ways, because vLLM's strict check waives fp8-quantized parameters.
- **Hash-routed MoE layers get the step's token ids** — the frozen `tid2eid` table selects the experts; the
  runner refuses to route such a layer without the ids. Capture-legal on the fixed-slot decode tier.
- **Tensor-parallel expert sharding in the plugin** — each rank derives its contiguous expert slice from the
  config; the unquantized lane slices the twin's own tables; the group all-reduce completes the partial combine.
- **Two cross-cutting serving fixes the distributed gate forced**: compiled twins may return outputs in their
  accumulation dtype, so the runner normalizes carrier/routed-input/final-norm dtypes at the seam; and the
  cross-process GPU lock is now scoped per physical device — a machine-wide lock deadlocks serving ranks
  (holder waits on a collective whose peer is queued on the same lock).

## Verification

- CPU: the ownership table driven unbound against the published key sets of all three attention layer shapes;
  hash routing contract; MoE serve-probe recognition of the published `n_routed_experts` spelling. Full local
  suite: 3225 passed (the two known `main` failures remain).
- Pinned 1Cat image on V100s: cache registration across sliding/CSA/HCA layers, checkpoint attention weights
  loading into the real fork modules (fused ordering, E8M0 scale decode, head-sharded sink), the compiled seam
  reproducing the eager layer for top-k AND hash layers, and the engine parity gate — the same checkpoint served
  single-rank and TP2×PP2 produces identical greedy token ids modulo numerical ties (both topologies must show a
  noise-sized top-2 margin at a divergent step; exact ids are the real checkpoint's gate, where logits are
  decisive).

Getting the fixtures faithful was most of the debugging: the fork's kernels accept only the published head
geometry (compressor 512 / indexer 128 / 128-aligned fp8 group outputs / 128×128 blocks), and the twin and fork
derive consistent compressor geometry only when `compress_rates` stays unpublished, exactly as the checkpoint
ships it.

## Still ahead in the plan

The real-checkpoint gates ride the next stage with the serving image: the TP8×PP2 target-host boot serving mixed
prefill/decode, then layer-level numerics and greedy token-ID agreement against the eager seam.


6erun:
External review (Codex CLI) ran over the branch; its findings and their resolution, all in `96c775f05`:

1. **High — capture accepted where the fixed-slot tier can't serve it** (hyper-connection combine always
   host-syncs; every GPU gate ran eager and masked it): the boot guard now rejects hyper-connection MoE capture
   loudly, and `emmy serve` defaults such models to `--enforce-eager`.
2. **High — fixed-slot selector vs expert shards** (global ids into rank-local tables): capture is rejected for
   sharded MoE, and the decode dispatch never takes the fixed-slot path when `expert_range` is set.
3. **Medium — parity gate could pass materially different flat distributions**: the gate now also requires the two
   arms' best logprobs to agree numerically at a divergent step, and the fixture gained the published model's
   hash-routed first MoE layer so token-id routing runs through the real engine at both topologies.
4. **Medium — double loading undetected**: a second checkpoint key landing on one destination shard now raises.
5. **Low — stale architecture entry**: the "captured but not yet served" bullet now states what serves, that decode
   capture is unsupported for this architecture, and what remains on-host.

All image gates re-ran green after the fixes (plugin/loading suite + the TP2×PP2 engine parity test).

Slonegg (approved):

PR #676, opened 2026-08-28
Bound splice construction and fix the extraction gate: the fused post block compiles again

6erun:
## The failure

DeepSeek-V4-Flash-0731's post twin (644 nodes) never finished compiling: every rank of the TP8×PP2
serve ground at 100% CPU for an hour on layer 1 of 21, and a solo compile did not finish in
**25 minutes** on the V100. `pre16` (60 nodes) compiled in 8 s — 10.7× the nodes bought >7,500×
the dependency resolutions.

## Three defects, one visible failure

**1. The splicer's demand worklist had no construction bound** (`ir/loop/splicer.py`).
#648 removed the old 8× work ceiling together with the fusion-quality gate it was entangled with.
The quality gate should stay gone — placement owns cuts — but without any bound, a
recurrence-shaped region multiplies distinct bindings instead of deduplicating: each stage is
re-demanded under *compositions* of σs (DeepSeek-V4's 20-iteration Sinkhorn chain is the live
case). Measured: **4.5M distinct bindings from 2,287 input statements**, only 6% of lookups
hitting the dedup table — genuinely distinct work, so no memoization can save it.

Across the whole compiler suite, no successful splice ever takes more than **8** bindings per
statement (2,315 splices measured; aggregate never exceeds ~1 binding per input statement). A
per-stmt cap of 16 refuses the doomed merge in milliseconds and the region stays unfused — a
**termination bound, not a quality gate**, and it has to be cheap because the greedy policy
re-runs fusion on every candidate graph it prices.

**2. `LoopBuilder` was quadratic** (`ir/loop/builder.py`).
Every insert rebuilt the immutable body tuple at every level and found each level's `Loop` by
scanning (404k inserts, 162M isinstance calls in one 110 s profile window). Construction is now a
mutable scope tree — descent is a dict lookup, prepend is an append to a reversed list —
materialized once at `finish()`. **Byte-identical output**: 200 randomized insert sequences match
the old algorithm exactly.

**3. The extraction round-trip gate rejected legitimate spellings** (`ir/tile/ir.py`).
With fusion bounded, a partial merge reached `extract_output_specs` whose sweep bodies were not in
canonical statement order. The gate compared reconstruction against the **raw** stream — but
`ProjectionRegion` stores its body in a `Lambda`, whose construction canonicalizes statement order
and commutative arguments, so any stream the normalization reorders was declared unrepresentable:
a hard *"Loop IR effects cannot be represented as output specifications"* compile failure over
pure spelling. The gate now compares against the stream **as reconstitution will spell it**; an
already-canonical stream is still gated on byte-identity, so nothing that passed before changes.

## Numbers

| | before | after |
|---|---|---|
| `post16` on the V100 (sm_70) | stall, >25 min | **68.8 s** |
| `post16` locally (off-GPU) | stall, >15 min | **18.8 s** |
| `pre16` / `expert16` on the V100 | 8.4 s / 3.2 s | 8.4 s / 3.2 s |

Kernel identity hashes are unchanged (`k_linear_reduce_8beaa5` → the same cubin
`2c9715ec…`), so no golden or corpus churn. The #671 redeclaration fix stays intact (0 nvcc
redeclarations, `compiled 1/1`).

## Verification

- `make test`: **3860 passed** (the two `tests/serving` failures are the known local
  Transformers drift — `LagunaAttention.gate_per_head` — and pass in CI).
- Four new tests, each red without its fix:
  - a recurrence chain built from two non-commuting affine maps fuses at depth 4 (8 bindings —
    the suite maximum) and is declined at depth 12 (2,048);
  - the builder's descend/prepend contract, pinned directly;
  - a two-sibling-sweep stream in non-canonical order still extracts.
- `make lint`: clean.

## What this unblocks

The DeepSeek-V4 V100 serve was hard-blocked on this — the boot could never leave its first
layer's compile. A full TP8×PP2 boot on this branch is the next step.


6erun:
Pushed two follow-up commits after walking DeepSeek-V4's post twin further down the pipeline on the V100. Current state:

**Working:**
- post16: never-finishes → compiles in ~57 s, **54 kernels** (pre-#648 baseline: 92; the cap alone without region-shrink: 433)
- Doomed recurrence chains cost only themselves: the splicer names the offending loop (`UnfusableStmt`) and fusion drops it plus its downstream closure, keeping every sibling merge
- The extraction gate compares both sides under construction normalization, which also makes the fusion lift-preflight predicate correct — an unliftable merge now stays unfused instead of hard-failing at `010_lift`
- `_sweep_start` sees through `Fold` operand edges (`Fold.nested` hides them from `_member_reads`), so a sweep store's wrap encloses a reduce that reads the sweep coordinate
- `make test`: 3862 passed, lint clean

**Known limitation (documented in the last commit):** one post16 kernel (`k_div_36_reduce`) still renders an unbound sweep axis (`a15`) under the schedule the live V100 selects — off-GPU the same graph schedules differently and renders correctly, so the emission bug is schedule-dependent. On-card stage dumps are preserved. DeepSeek V100 serving remains blocked on that one kernel; everything else in this PR stands on its own.

Slonegg (approved):

6erun:
Rebased onto latest main (#674, #675, #678).

**#678 superseded one of my changes, so I dropped it.** An earlier revision widened `_sweep_start` to see through `Fold` operand edges, because `Fold.nested` hides them from `_member_reads`. #678 landed the structural form of that fix — an output sweep read by a nested contraction operand is promoted into the Tile's free-axis placement — which is the right layer. Measured on the V100 both ways: **identical, 53 of 54 post16 kernels compile with or without it**, and it cost compile time (57 s → 31.5 s locally once removed). Dropped rather than kept as a second mechanism doing the same job.

What this branch still contributes, all on top of #678:
- splicer construction bound (per-stmt binding cap) — without it `post16` never finishes
- `LoopBuilder` quadratic → linear, byte-identical output
- fusion region shrink-retry — one doomed chain no longer abandons its sibling merges (433 → 54 kernels)
- extraction round-trip gate compared under construction normalization, plus the lift-preflight it makes correct
- captured-result sweep writes decline cleanly instead of crashing `Lambda` formation

`make test`: **3930 passed**, lint clean.

**Known limitation unchanged:** `k_div_36_reduce` still renders an unbound sweep axis under the schedule the live V100 selects — on #678's main, with or without this branch. Schedule-dependent emission; off-GPU the same graph is correct. Tracked separately.

PR #686, opened 2026-08-29
Five correctness fixes in the placement cut

6erun:
Rebased onto `83130d6d9` (#690). **#679 landed this branch's original fix** — the
evaluation-domain rule (`_carries_iteration` and both refusals) is on main
character-for-character, along with an equivalent producer ordering — so that is
dropped from here. Since then **#688** replaced the scoped-lambda privates with
`Closure`, **#689** moved kernel identity onto the canonical lowered body, and
**#690** reworked reduce binding; this branch merges cleanly over all three, and
every fix below was re-checked against the new base rather than assumed to survive.

What remains is five bugs main still carries. One is the `KeyError` that killed the
TP8xPP2 boot; the other four came from two rounds of codex review on this branch.

## The boot blocker: a piece declared its inputs from the lowered body

DeepSeek-V4's TP8xPP2 boot got through weight load, per-layer compile and KV sizing,
then died in whole-program graph capture:

```
KeyError: 'matmul_1_b_unsq_bc__place_7174ebb040_0'
GraphCaptureError: whole-program capture failed
```

Traced on the live V100: the node's `arg_order` held eight buffers its `node.inputs`
did not; the divergence appeared at `010_materialize`; and the TileOp going in already
read all eight — `Fold.lower()` simply did not surface them.

`Body.loads` walks `Stmt.nested()`, and a Fold's operand **edges** are not nested
statements: `Fold.nested` yields the lift body, and nothing at all for a contraction,
whose algebra is meant to read as edges rather than body deps. That is fine for a fully
flattened stream, but a lowered body still carries Folds as **terms** wherever a region
kept them — `post4096`'s root holds its cones inside `ProjectionRegion`s. `_piece_inputs`
and `_producer_order` asked exactly that walk. Each cut on the node re-derived its inputs
from the lossy view and dropped more; the materializer builds the kernel from the
**stored** tree and emitted them all; the workspace producers, left without a consumer
edge, were pruned as orphans.

`loaded_buffers` answers the question the cut actually has, and the three sites that ask
it use it. **Verified on the V100: `post4096` builds its plan and launches**, where it
previously raised.

## Four from review

- **A shared cone was typed once, not per occurrence.** `edge_dtypes` gained a `scope`,
  and seeded from the root it doubles as the tree's dtype table — but it cached under
  `id(edge)`, so one cone under an f16 and an f32 capture got the first answer for both
  and a cut sized the seam's workspace at the wrong dtype. Keyed now by edge plus its
  captures' dtypes; the table keeps only nodes whose occurrences agree.
- **A fold a plain statement holds had no lexical environment.** A statement binds axes,
  not SSA definitions, so it is not a resolution scope — but it can hold a stored fold,
  and one reached only that way could not resolve its captures, silently dropping its seam.
- **A cone's closure is part of its value.** Clustering compared cone syntax only. Two B
  cones can be alpha-equivalent and still read different sources — a capture is a free
  name, and one host may define `x = sum(a)` where the other defines `x = sum(b)`. Merging
  them materializes one and lets the other read it, handing the second contraction the
  first's value. `_closure_key` makes what a seam closes over part of the comparison.
- **A required producer keeps its own seam.** A dependent reads its producer's workspace
  by the name that producer *binds*, while clustering re-points a value at a representative
  whose result names are its own. Required producers no longer cluster at all.

Plus: requirement injection matched `Load` only, so a producer that cut at a storage
frontier — whose entry is the projection wrapping the raw load and its residue — supplied
nothing and left the name unbound.

## Verification

`make test` 3988 passed / 0 failed; `make lint` clean. Six regression tests, each checked
to fail against `origin/main` and pass here.

Codex's second round re-verified the earlier fixes and separately cleared `loaded_buffers`
(including that it need not follow `step_stmts()` — derived folds reuse stored `Load` edges
and introduce no new buffer, which matches the on-host evidence), `_carries_iteration`, and
both evaluation-domain refusals. On enumeration cost: `post4096` is 50 folds / 50
occurrences / 0.1 ms, attention 23 folds / 38 occurrences / 0.03 ms, max depth 8.

**Line balance (step 18): +133 under `emmy/`.** `loaded_buffers` is new shared capability —
the honest reader for "what does this stored term read", now used by three call sites — and
the rest is the four fixes plus the docstrings this repo keeps beside the failure each one
prevents. The one place the growth was layering, an inline seven-clause alignment test,
became `_closure_key`.

## Still open, and not this PR

`post4096` now launches and then runs for tens of minutes: wide contractions nested under
projection regions lower to serial loops. That is a fusion legality/pricing question and is
what still stands between this branch and a serving boot. Probe on the host at
`~/post4096only.py`; log `~/diag/post4096.log`.

### DIFF
--- a/emmy/compiler/ir/ARCHITECTURE.md
+++ b/emmy/compiler/ir/ARCHITECTURE.md
@@ -84,6 +84,17 @@ These are term operations spelled the same way so one canonicalizer and one deep
 vocabularies; they are not statement behaviour, and `Fold` has no `render`. Impure computation stays
 in Loop IR until total lift; there is no impure `Lambda` construction path.
 
+**`nested()` is the STATEMENT protocol, and it deliberately does not reach a Fold's operand edges**
+— it yields the lift body, and nothing at all for a contraction, whose algebra is meant to read as
+edges rather than body deps. Every walk built on it (`Body.iter`, and so `Body.loads` /
+`Body.writes`) therefore answers for a fully flattened stream only. A lowered body is not always
+one: a term survives lowering wherever a region kept it (`ProjectionRegion` holds its cones as
+terms), and a walk over `Fold.lower()` then silently under-reports everything beneath such an edge.
+Ask `loaded_buffers` instead whenever the answer must cover what a consumer of the STORED tree will
+reach — the kernel materializer walks that tree, so anything deciding a node's graph inputs has to
+see what it sees. Asking the lowered view there is what let a cut declare fewer inputs than the
+kernel it produced went on to read.
+
 ## Invariants by stage
 
 - **Frontend → tensor** (after `decomposition`): `LinearOp`, `MatmulOp`,
--- a/emmy/compiler/ir/pure/fold.py
+++ b/emmy/compiler/ir/pure/fold.py
@@ -875,6 +875,42 @@ def deep_reads(stmts: list[Stmt]) -> set[str]:
     return out
 
 
+def loaded_buffers(node) -> set[str]:
+    """Every graph BUFFER loaded under ``node`` — a term, a stmt, or a body of them.
+
+    ``Body.loads`` walks ``Stmt.nested()``, and a Fold's operand EDGES are not nested statements:
+    :meth:`Fold.nested` yields the lift body, and nothing at all for a contraction, whose algebra
+    is meant to read as edges rather than body deps. That is fine for a fully flattened stream,
+    but a lowered body still carries Folds as TERMS wherever a region kept them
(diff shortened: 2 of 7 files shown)


6erun:
Superseded by #692, which contains this PR's five fixes byte-identically (rebased onto current main as cb061d118) plus the failure-attribution, disqualification, composed-cut compile, and placement-ballot work on top. The pre16 fix developed on this branch landed separately via #682/#688 and is re-verified at main+#692 (run_once 8.182s on the V100).

PR #690, opened 2026-08-29
Bind a sweep-reading reduce inside its output sweep; offer it only the serial fold

6erun:
## What

[#677](https://github.com/cloudrift-ai/deplodock/pull/677) stopped the normalize hoist from *moving* a sweep-reading fold onto an operand edge — the route by which DeepSeek-V4's `k_div_36_reduce` got an unbound sweep axis. This closes the same hole one layer down: a stored term whose operand edge **already** carries such a fold is directly expressible, and the materializer still emits the sweep coordinate as an undefined identifier for it, under every row.

The peel is the reason. `_factorize` binds the root operand *outside* the projection, so no wrap position inside the projection body encloses the reduce, and `apply_output_specs` can never put the sweep loop around it.

## Changes

- **`_factorize`** detects a plain output sweep whose axis the peeled root's cone reads (`edge_refs_axis`). The serial fold realizes the shape whole, so the projection binds **unpeeled** through the zero-axis `_bind` arm and the sweep loop wraps operand and projection together. A cooperative / ILP row (or an output-tiled contraction root) has no such realization — the sweep is distributed across the very lanes the reduce would have to re-run on — so it declines via `UnbindableProjection` (`RuleSkipped(reject=True)`) and the greedy retries the next row.
- **`_reduce_moves`** offers such a fold exactly the serial fold, beside the existing scan rule and with the same pin refusal. Deciding it at the offer keeps the greedy off the binder's decline, which costs a full re-resolve per declined row and retires structural pricing as collateral — the DeepSeek V100 serve compile spent **382 s** on those retries where the narrowed offer resolves once, in **40 s**. The binder decline stays as the backstop.

## Tests

Two at `factorize` (serial emission keeps the sweep axis bound everywhere it is read; a coop row refuses) and one on the narrowed catalog at `_reduce_moves`. All three **fail on main without this change**.

## Verification

- `make test`: **3983 passed**. The two `tests/serving` failures (Laguna gate, GLM4 MoE tail) reproduce unchanged on clean main at 1498d6ac7 — local Transformers drift, not this branch.
- `make lint`: clean.
- Realization corpus: green, no case verdicts changed.
- On-card (16× V100, sm_70): with this change on the branch it was authored against, DeepSeek-V4 `post16` nvcc-compiled 54/54 (was 53/54, `a15` undefined), and `k_div_36_reduce` emitted the sweep loop enclosing the mean reduce.

## Not included

A follow-up commit on the branch this came from generalizes the same idea to an **enclosure gate** — any fold nested inside another fold's reduce axis is emitted serially by `_factor._emit`, which reads no schedule, so the bands offered there are partitions the kernel never realizes. That work predates [#679](https://github.com/cloudrift-ai/deplodock/pull/679) and is **not** in this PR: #679 added `attention/rmsnorm-*` corpus cases whose pinned rows record `REDUCE@<enclosed site>: coop`, and the gate refuses those pins, so 5 cases fail. Deciding whether those stamps are phantom (corpus regeneration) or the gate is over-broad (compiler fix) is its own change with its own evidence.

### DIFF
--- a/emmy/compiler/pipeline/passes/lowering/kernel/ARCHITECTURE.md
+++ b/emmy/compiler/pipeline/passes/lowering/kernel/ARCHITECTURE.md
@@ -104,7 +104,11 @@ kernel-boundary `TileOp.output_specs` are reconstituted at their owning projecti
 (a STREAMED store — one whose values are an observed fold's observer results — rides the recursion down to the leaf
 instead and splices into the reduce loop after the observer stmts), so everything below the peel — the sinks,
 cooperative loop distribution, and split realizers — consumes the identical statement stream that entered total
-lift. The
+lift. An output sweep whose axis the peeled root's cone reads cannot wrap at the peel (the root binds outside the
+projection, so no wrap position encloses it): the serial fold binds the projection UNPEELED so the sweep loop wraps
+operand and projection together, and a cooperative / ILP row — whose lanes the sweep would be distributed across —
+declines via `UnbindableProjection` (`RuleSkipped(reject=True)` at the pass boundary; the greedy retries the next
+row). The
 recursion, the binder, the reduce-axis tiling, and the shared-row staging apply live in `_factor.py`; the four tiling
 levels every tier seals through are `_tiling.py`, which knows a `Side` pair, integer counts and three callables — no
 node kinds, no algebra, no `Ctx`. That is the decide/realize seam: the tile schedule picks the plan, `_tiling` is
--- a/emmy/compiler/pipeline/passes/lowering/kernel/_factor.py
+++ b/emmy/compiler/pipeline/passes/lowering/kernel/_factor.py
@@ -53,13 +53,13 @@ from emmy.compiler.ir.elementwise import ElementwiseImpl
 from emmy.compiler.ir.expr import BinaryExpr, Literal, Var
 from emmy.compiler.ir.kernel import Tile
 from emmy.compiler.ir.kernel.ir import Smem, Sync, TreeHalve, WarpShuffle
-from emmy.compiler.ir.pure.fold import Fold, _operand_result_names, _unique_edges, is_contraction, operand_body
+from emmy.compiler.ir.pure.fold import Fold, _operand_result_names, _unique_edges, edge_refs_axis, is_contraction, operand_body
 from emmy.compiler.ir.schedule import Raster
 from emmy.compiler.ir.sigma import Sigma
 from emmy.compiler.ir.stmt import Accum, Body, Cond, Init, Load, Loop, Select, SelectBranch, Stmt, StridedLoop, Write
 from emmy.compiler.ir.tile import FoldMove, Level, ReducePlan, ReduceStage
 from emmy.compiler.ir.tile.ir import ProjectionRegion, _projection_results, apply_output_specs, observed_result_names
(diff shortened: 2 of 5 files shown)


6erun:
## Codex review — verified, one fix applied, two filed as pre-existing

Ran `codex exec` against this diff. Three findings; I verified each independently rather than taking them at face value, and the picture differs from the report in one important way.

### Fixed here (219b6e417)

**The refusal read its plan off the wrong node.** The decline looked up the `REDUCE` slice on the peeled operand itself, so a chain of zero-axis projections between that operand and the reducing fold hid the partition — the wrapper carries no `REDUCE` site, the lookup found `None`, and a cooperative row bound as the serial fold.

Codex described this as "silently discard a partitioned schedule", which is right, but understates what it replaced: on **clean main the same shape emits `undefined_axis=True`** — broken CUDA that *claims* coop. So the original diff already turned a miscompile into a correct-but-partition-free emission. Still worth closing, since this PR claims the binder declines partitioned rows, so it should actually be total over the shape. The guard now descends to the iterating node, and a new test pins both halves (serial chain emits with every axis bound; coop chain refuses).

### Pre-existing, not introduced here — filed rather than fixed

**1. Cross-CTA `REDUCE=g…` splits still mis-emit for this shape.** `035_split_reduce`'s `_projection_refusal` explicitly permits a head fold stored as an operand edge ([`_split.py:161`](https://github.com/cloudrift-ai/deplodock/blob/main/emmy/compiler/pipeline/passes/lowering/tile/_split.py#L161)), and the deferred partial's workspace is sized over `(ksplit, *free)` — the sweep is dropped. Confirmed: `_projection_refusal` returns `None` for the sweep-reading tile on **clean main at 1498d6ac7 exactly as on this branch**, and this PR does not touch `_split.py`. It is the same family as what this PR fixes (the in-kernel tiers), one tier over. Fixing it means either extending the split's refusal or promoting the sweep into the workspace — its own change, with its own corpus evidence.

**2. The multi-root path bypasses the guard.** `_bind_roots` is dispatched earlier, at [`_factor.py:295`](https://github.com/cloudrift-ai/deplodock/blob/main/emmy/compiler/pipeline/passes/lowering/kernel/_factor.py#L295), before this check. `TileOp.__post_init__` normally makes that safe by promoting a contraction-read sweep to a free axis, but promotion only considers contraction operands, so a forest carrying an independently-owned plain sweep-reading fold reaches `_bind_roots` and fails with `ValueError: output-tiled roots disagree on their physical grid` — not caught by the `UnbindableProjection` boundary, so it aborts instead of declining. `_bind_roots` and `_merge_root_tiles` are untouched by this PR and behave identically on main.

### Confirmed sound

Codex found no serial-equivalence defect in the unpeeled `_bind` path, and I agree with its reasoning: `_emit` still emits every unique operand edge in order, `_emit_body` still attaches `ProjectionRegion`-owned stores, and `_bind` filters region-owned specs and passes `observed_result_names(op)` so streamed observer stores still splice into their reduce loop. It also confirmed the `_reduce_moves` narrowing does not wrongly catch a contraction-promoted sweep, since `__post_init__` clears that sweep before scheduling.

`make test`: 3984 passed. The two `tests/serving` failures reproduce unchanged on clean main. `make lint`: clean.

Slonegg (approved):

PR #695, opened 2026-09-01
Serving compiles and runs the declared W4A4 program for NVFP4 checkpoints

gnull:
## Abstract

Serving an NVFP4-quantized checkpoint previously ran the 16-bit scaffolding: the compile lane spelled the 4-bit activation encode, serving never did, so a W4A4 checkpoint — 4-bit weights and 4-bit activations — served as W4A16. This change makes serving compile the declared 4-bit program end to end. Making those programs sound surfaced a chain of compiler defects, each fixed here with a regression test. The deepest one, a class of fused kernels left without a usable GPU grid, is answered by the output-owning cut: a new form of the existing cut that lets a split-off kernel piece write the outputs its computation owns. Serving booted and answered on this branch's earlier heads. On the final head one remaining slow-kernel family keeps the boot from finishing. This change therefore ships the working serving mechanism with its measurements, and hands the speed work to the follow-up with a designed fix and measured prototypes.

---

All measurements below are from nvidia/Qwen3-8B-NVFP4 (an NVFP4 repack — a re-quantized republication — of Qwen3-8B) on a rented RTX 5090 (sm_120), unless a different card is named.

## What serving does now

`emmy serve nvidia/Qwen3-8B-NVFP4 --generate` compiles every serving split as the declared W4A4 program: the static 4-bit input encode the checkpoint declares runs ahead of each marked linear (`to_f4e2m1`), activations travel as packed `f4e2m1x2` buffers, and the block-scaled tensor-core cell (`mma_m16n8k64_e2m1_f32`) lands on the cut pieces that reach it. The vLLM engine no longer receives the modelopt quantization config when emmy owns the weights (`engine_config_overrides` recognizes NVFP4), and the twin capture gained an NVFP4 branch, so the coded per-layer graphs serving compiles are the ones tuning measures. GPU parity: `tests/serving/generation/test_gen_runner_gpu.py -k w4a4` compiles a serving split and matches the numpy reference executing the same stamped graph, max rel err 5.3e-4 against the 2e-3 bound, green on the RTX 5090. (This verifies the compiled program, not the checkpoint's quantization quality — both sides read the same 4-bit codes.)

## The compiler work

The serving programs exposed defects across the lowering; each fix landed failing-first and carries regression tests. The corpus cases added for them: `fused/nvfp4-gate-up-requant-place-cut.yaml`, `fused/nvfp4-quantize-cut-shared-normalizer.yaml`, `matmul/nvfp4-w4a4-packed-slab-loopify.yaml`.

| fix | defect |
| --- | --- |
| render: a re-rolled fragment resolves its dtype through the family stem | packed-slab kernels crashed at render — and the crash masked a silent mis-render reading one packed byte as one value |
| cut workspaces name their reads after the workspace | two structurally equal cones collided on one SSA name: duplicate declarations nvcc rejects |
| a placement cut re-spells the consumer's boundary stores | stores kept a renamed value's old name, multiplied per cell into hundreds of undefined identifiers |
| the greedy reads the settled terminal to detect un-lowered nodes | a stranded TileOp escaped the decline path and crashed engine init; it is now a clean decline the retry walks past |
| a compile-time guard: the finished kernel body binds every name it reads | turns this bug class into loud Python errors naming the kernel; found a fourth latent instance on landing |
| block-scaled packed-pair operand cones are not cut seams | such cuts destroyed the pair reading the tensor-core cell binds on, and emitted value-corrupting stores; three silent drop-outs became refusal messages |
| the recording codec accepts a kernel with no contraction site | the recording codec rejected a cut's copy-shaped pieces outright |
| the output-owning cut | below |

The output-owning cut is the substantive design contribution (GLOSSARY.md carries the term). A fused NVFP4 kernel writes two outputs over store loops of different widths; main's shared-sweep rule for choosing a kernel's grid axes (adopted unchanged) then gives that kernel no grid axis at all, and the existing cut could not help: it peels contractions, while both stores stay behind. The new form lets a cut seam whose cone solely owns some of the kernel's outputs write those outputs directly — every resulting kernel is then single-output and gets its grid axes from main's rule as-is. No new knob and no new spelling: the same `PLACE` key, offered only where it gains the piece a grid axis. A controlled comparison against the base commit verifies the payoff, over the 64 serving realizations then failing to bench: among the 32 whose kernels offer the seam, the base benched 0 cleanly (a clean bench = the run produces a verified measured row); this branch benched 5 cleanly and brought 9 more from not-running to running-but-over-budget; zero regressed. The flagship pair: the post-attention kernel failed every timing budget at 25.7 s per launch on the base and measures 21.7 ms under the recorded route here, reproduced across three runs.

This branch merged main four times and adopted main's designs wholesale where they superseded ours. Main's evidence replay replaced our parallel replay machinery. Main's shared-sweep rule replaced our shared-axis design after landing stricter. Main's attention rework removed the pass our FlashAttention mask fix repaired, and the fix went with it. Main absorbed pieces of this branch in return — the store-sink decline and the sibling-axis binding — and the merges re-applied what main's copies missed.

## The case that serving delivers once the slow kernels are fixed

The boot on the final head does not reach a healthy server, and this PR deliberately ships without one. The supporting case:

1. The GPU parity test above exercises the serving compile path end to end on the deployment card, with no boot required.
2. Earlier heads of this branch booted to a healthy server and answered chat probes coherently: an RTX 5080 Laptop in 9-14 minutes and the RTX 5090 in 37 minutes, both using canonical golden files recorded on those heads (removed when main's replay retired their format).
3. On the final head the boot ran seven hours and was still in vLLM's warm-up passes when stopped — no crash, no wrong result anywhere in its log. The first 50 minutes are fully attributed: 36 min of layer-0 kernel compiles (one family resolves twice), 12 min of the boot roofline audit timing that same family, 45 s of actual weight binding. The remainder is the warm-up forwards running that family's kernels per layer; a separate open factor is that the long-input post-attention program regressed from 23.2 s to 114.4 s per warm-up run somewhere between an earlier recorded election measurement and this head, which a bisect would localize.
4. The blocking family is the NVFP4 output requant: its packed-code output carries six contraction roots, and the kernel binder requires an output-tiled root to own each output specification independently — six break that requirement, producing near-serial fallback code. The compiler's own source comments name the escape — the same computation is ordinary as a kernel of its own — and the designed fix splits work off through placement cuts, following the same design process that produced the output-owning cut. Judgment, stated as such: the 25.7 s → 21.7 ms result above is one datapoint from a neighboring defect of the same class, and it is why we expect this class to yield; the fix itself is unlanded and is the follow-up's first item.
5. The residual boot costs carry measured prototypes: caching the schedule enumeration's repeated tree walk measured 1.8-5x with byte-identical kernel sources, and the compile pack cache round-trips (it is simply unset outside the serving image today).

The tuning campaign behind these numbers recorded 87 of 128 serving realizations as verified measured rows on the final head; the remaining 41 belong to the requant family and fail their timing budgets until the fix lands. The split's work queue is the eight GEMM kernels blocked at every width.

## What does not ship, and why

The canonical per-card golden and the serving env file: the release-image guard requires a golden that passes the strict release check, which requires all 128 realizations verified. Tuned speed claims: the recorded rows for the re-encode-adjacent kernels are measurements under recorded routes, labeled as such, not search-tuned winners. A stock-vLLM comparison on this checkpoint: vLLM 0.23 cannot load it (its kv-cache-scale loader crashes on the 72 kv-scale tensors this repack carries); the comparison target moves to the 27B checkpoints (nvidia/Qwen3.6-27B-NVFP4, Inferact/Qwen3.8-27B-NVFP4), whose repacks carry none.

The `emmy/` line balance is +819/−219: the growth is the W4A4 serving path and the output-owning cut — new capability, not layering — and the four merges deleted this branch's superseded designs along the way.

## Evidence

`make lint` passes (ruff check clean, 782 files formatted). `make test` ran twice on the RTX 5090 box — once at the merge head, once at the final head — with identical verdicts: 2 failed, 4933 passed, 301 skipped, 41 xfailed. Both failures are external to this PR: one is the known worker casualty of `matmul/imap-transpose-a-tma_xfail_correct.yaml` poisoning its xdist worker (the affected case passes when rescheduled, and all 15 NVFP4 corpus cases pass in isolation); the other, `reduce/rms-norm-cut-sweep-work.yaml-correct`, reproduces identically on origin/main. The realization corpus carries three new NVFP4 cases green through `correct` on the 5090.

## Findings for the team

- `matmul/imap-transpose-a-tma_xfail_correct.yaml` hard-faults the GPU when its `correct` stage runs (Xid 13; 19k compute-sanitizer errors) and poisons the CUDA context for the next test in its worker.
- `_fed_store_dtype` types a contraction-operand workspace by the consuming kernel's store dtype; on a kernel whose store is a bits encode this emits value-corrupting code. Unreachable for block-scaled shapes after this PR's seam rule; reachable elsewhere.
- A corpus case that records a multi-root kernel with a real grid cannot finish `offered` in practical time: the stage materializes the kernel set's complete row set.
- Composed cut routes (a cut of a cut piece) cannot be replayed from recorded evidence; already tracked by an xfail case.
- `emmy tune` logs no knob row for a failing candidate, so a failure seen only under tune cannot be reproduced from its logs.
- `helpers.regenerate` restamps only the first realization's identity in a corpus case; later entries need hand-restamping.
- The cold search retires a whole structural pick when the binder refuses the first row it priced for one piece (improved by main's #730; the multi-kernel-set interaction remains).
- The boot roofline audit times each flagged program with one warmup and three repetitions; its docstring says once. A repetition cap would cut most of 12 minutes from affected boots.
- `tests/durations.json` has drifted from the suite (dozens of 5s+ tests missing on main and here alike); a `make test-durations` re-measure on a GPU box would restore the balance data.
- `test_evidence_row_vouches_reads_a_bare_row_key_like_a_bare_pin` fails under one ad-hoc parallel subset while passing alone, serial, and under `make test`'s own distribution — a cross-test isolation sensitivity, pre-existing.
- `_cut.py` and `_split.py` carry two same-named `_piece_inputs` helpers with different semantics; both predate this PR; consolidation is a standalone refactor.

https://claude.ai/code/session_01MXJRsB9wRXfgbXNX33rvvr


gnull:
@Slonegg this PRs main point is making proof-of-concept serving with NVFP4 work (with simple 8B model). The actual serving time can be dramatically slow because it's hard to predict what kernels serving will pick (so we can at least somewhat tune them with goldens) and make sure it picks some good ones. What's left for follow-ups to make Qwen 3.8 27B work:

- support for DeltaNet layers,
- optimize nvfp4 weights loading in serving (bit packing in python seems to be making is slow),
- tuning of kernels, chasing small issues like lowering failing to generate the right atoms sometimes, which I'm leaving for the last finishing step.

That last step is particularly sensitive to big refactorings and merges with main (the last one raised serving boot time from 15m to 50m, some kernels take seconds to run), so I'm leaving it for later. For now, I just want to land PRs like this one that bring support for things, even if slow atm.

Slonegg (approved):

PR #706, opened 2026-09-03
Nested write-only output sweeps: OutputSpec.sweep as the store's loop-nest path (#699 regression)

6erun:
## A #699 regression: nested write-only output sweeps lost their boundary spelling

#699 gave an output sweep's per-cell projection a term and its stores a single-axis `OutputSpec.sweep`, and
removed the `ProjectionRegion` carrier that sibling and nested output loops rode before. A twin whose output
sweeps NEST — DeepSeek-V4 `post4096`'s serving twin has two such shapes: a sweep whose per-cell projection and
store sit beside nested write-only sweeps (its gate stream: `view_8` beside the `mul_16` and `matmul_1` nests),
and a write-only nest with no outer store (its softmax pair) — could no longer lower at all: the projection
formation scooped the inner sweep `Loop`s into a `Lambda` (`Lambda body must be pure — impure stmt kind(s):
['Loop', 'Loop']`), and the boundary had no spelling for a store that rode two loops.

## The fix

- The lift forms the per-cell projection from the PURE cell members alone and keeps nested write-only sweeps in
  the retained cell at their source position (`_fromloop.lift_body`).
- `OutputSpec.sweep` is the output loop NEST the store rode in the source, outermost first — the one fact
  extraction destroys, since a write's index names its coordinates but not the order its loops nested in; the
  term route (`Fold.lower`) takes only the path's extents and orders its own nest. Reconstitution opens a
  group's loops innermost-out, leaving any prefix a later spec's path also carries to that later group, so one
  source loop is opened exactly once however many specs rode it — prefixes, not names, since sibling nests
  legitimately reuse inner axis names (`apply_output_specs`, `_shares_prefix`); extraction pops nested
  write-only loop nests in stream order (`_sweep_specs`); the byte-identity round-trip gate is unchanged and
  stays the sole arbiter. Single-axis paths reconstitute exactly as before, so no kernel restamps and the
  corpus is untouched.
- The consumers of `sweep` update mechanically (`Fold.lower` coordinates, the promotion rule and `_own_axes`
  in `TileOp.__post_init__`, `classic_projection`, `_split`, `_factor`'s swept-axis refusal, `_dump`, `_cut`'s
  outer-axis list, the lift's axis table).

#699's own recorded debt for this gap — `test_shared_broadcast_chain_no_pure_indexmap_remains`, a strict
xfail — passes and loses its marker. Four nested-shape round-trip tests and a nested-sweep lift test pin the
shapes; the tile ARCHITECTURE's boundary paragraph describes the path vocabulary.

## Verification

`make test` green with no corpus staleness (single-axis paths are byte-identical), `make lint` clean,
`emmy eval prior --dataset nodes` unchanged. The deterministic DeepSeek `post4096` replay (pinned twins + the
pre-tune host-DB copy) lifts and elects again on current main — see the update comments for the election
shape and for #704's golden regime-pin fallback, which is a main-side issue being filed separately. No speed
claims: the host measurement of any route remains owed.

The serving plan's round-three item 4 is rewritten for the review's direction: the path to serving is a
recorded golden carrying every kernel of the route, booted under strict evidence — not pricing floors.

Scope note: this PR was reduced at review to the output-spec fix; the serial-floor pricing work it once carried
was removed (see the update comment), and #702's clamp is untouched here.


Slonegg (changes requested):
The floor should be gone, as mentioned in #702 ; output spec fix makes sense.

6erun:
702 or 704?

Slonegg:
> 702 or 704?

In #708  - apologies. In general, remove the lower bound hack. If the prior kicks in, it means the golden is not replayed or not present for the kernel.

6erun:
Complying: #706 is being reduced to the output-spec fix (the nested-sweep boundary spelling #699 lost), and #708 will be closed.

One scope question so main doesn't end up with half a mechanism: #702's serial-work clamp in `_resolved_price` is the same lower-bound device these two PRs extended — it only matters when the prior decides. Should it be reverted as well, or stay since it's merged?

6erun:
**Reduced to the output-spec fix, per review.** The branch is rebuilt on current `main` as three commits: the nested-sweep boundary fix (`OutputSpec.sweep` as an axis path, the lift keeping nested write-only sweeps in the retained cell, #699's strict-xfail debt test flipped), its review fixes (`_shares_prefix` as a module rule, the path's reason on the spec, the store-then-nest round-trip test), and the serving plan's item 4 rewritten for the recorded-golden direction.

**Removed:** the serial-floor pricing work — `S_ext_serial_cell_issues` and the issue-priced `serial_floor_us` / `SERIAL_ISSUE_FLOOR_US`, the guard-margin re-measure and its test pin, the `SearchDB` schema bump (stays v4), and the corpus restamps (the corpus is back at `main`'s stamps; `make test` reports no staleness). #702's clamp is untouched pending the scope question on the thread.

**Remains:** exactly the #699 regression fix. Gates on the rebuilt branch: `make test` green with no staleness, `make lint` clean, `emmy eval prior --dataset nodes` unchanged. Acceptance for this fix is that the DeepSeek `post4096` twin lifts and elects again on current main — the election shape is reported in the next comment, with no claims about its speed. The earlier merge with `main` (#704, #707) is superseded by the rebase-style rebuild; nothing else changed.

6erun:
**Acceptance for the reduced scope — the twin lifts and elects again on current `main`.** Deterministic replay of the DeepSeek `post4096` twin (pinned twins + the pre-tune DB copy) on the rebuilt branch:

- With #704's golden regime pin (`FAST_MATH: false`) neutralized and nothing else changed: the twin **lifts** (this is what #699 had broken) and elects a composed route — 31 nodes, 10 structural decisions, 0 rejections, **11 kernels**: the FFN contributions are computed once in a producer piece (the `linear_2_wt[a29 * 2048 + …]` walk appears exactly once, inside it) and consumed by a staged mma piece (`linear_4_wt` smem-staged), with the statistics and gate pieces beside them and the residual. Same plan as the branch's earlier post-#699 election; no claim about its speed.
- Under the pin, as already documented on this thread: the residual root's ranked row is refused at `k:010_materialize` ("an output-tiled root forest must cover the complete projection"), the strategy's retry retires the structural picks wholesale, and the compile falls back to the 1-kernel fused monster (1007 s). A main-side issue being filed separately, together with the retry-policy ordering.

Gates: `make test` 4143 passed / 0 failed with no corpus staleness, `make lint` clean, `emmy eval prior --dataset nodes` unchanged (rho +0.51, regret@1 1.00x, 176 pools). The host measurement of any route remains owed.

Slonegg (approved):

6erun:
Updated with `main` (merge `c3e9d778`: #713, #715; no conflicts). Lint clean; 4144 passed. The two failures in the merged suite — `attention/sdpa-hd128-softmax-v-mma.yaml` at `offered` ("STAGE pin 'd1/smem' does not resolve for this contraction") and `realized` (unreproducible WORK/TILE/STAGE/RASTER pins) — reproduce identically on `origin/main` at `7d0a63a4`, where #715 added that case; they are not from this branch.

6erun:
Reverting #702's clamp as well, per the decision: https://github.com/cloudrift-ai/emmy/pull/719

PR #708, opened 2026-09-03
post4096: bound a launch by its blocks — the launch-total serial floor (stacked on #706)

6erun:
Stacked on #706 (`feature/materialize-ffn-matmul-contributions`); base retargets to `main` once #706 merges.

## What this is

The per-thread serial floor (#702, refined per issue in #706) prices what ONE thread must retire. On DeepSeek-V4
`post4096`'s route it left a tie it cannot see: the contribution producer recomputing the mHC statistics once per
(row, a28) block (2^23 blocks) and its statistics cut computing them once per row (4096 blocks) floor at the same
~2.5 ms per thread — the fused arm's 1024-way cta×coop coverage divides the same nest — while they differ 2000×
in what the LAUNCH retires. The greedy kept the tie's fused side.

This adds a launch-total floor beside the per-thread one, under the same 1 ms guard (`_resolved_price` clamps a
summand to the larger of the two floors once either passes it):

- `features.launch_floor_us(knobs, cells)` = `launch_blocks × S_ext_serial_cell_issues × 1e-8 µs` — every
  block evaluates the kernel's worst serial nest at least once, at 1e5 issues/ns, a machine-width rate no GPU
  reaches (a V100 issues ~500/ns). No reduce-partition coverage divide and no cta multiply: partitioning moves a
  nest's trips between threads or blocks, never out of the launch.
- `features.launch_blocks(knobs, cells)` = the launch's output cells ÷ the most cells one block covers: the
  row's codec output tile `tile_m × tile_n` when it spells one, else one cell per thread over the largest block
  a GPU runs. Over-counting a block's cells only under-counts blocks — the safe direction. The codec tile for
  the measured dominant piece's row is 256 × 64 cells, exactly the 1024 blocks its emitted kernel decodes
  (pinned by test).
- `greedy._grid_cells(op)`: the terminal placement's static grid extents (sweeps are loops a thread walks, not
  grid parallelism, so they are excluded by construction; a symbolic extent → no launch bound).

No new stamp, no schema bump, no corpus restamp: the featurizer is unchanged, so every sub-guard election is
untouched by construction, and the freeze prior eval is unchanged (rho +0.51, regret@1 1.00x, 176 pools).

## What was rejected, and why it is written down

Counting per OUTPUT CELL (per-cell issues × cells) was the first design and is not a bound: a serial nest that
is an operand cone of an outer contraction is shared across every cell of a block that does not read its
coordinates, and the tile chooses how many. On `post4096`'s measured dominant piece (13.21 s) it charged
2^24 cells × 2^27 issues → 22.5 s at this rate — above the measurement. Per block the same kernel prices
0.67 µs (its per-thread floor is what catches it). The refutation is a test, and `launch_floor_us`'s docstring
says why, so nobody tightens it back. The tighter honest follow-up — per-nest evaluations projected onto the
axes the nest reads — is recorded in the serving plan with its cost (a structural stamp of the nest's read free
axes plus the codec's m/n slot mapping).

## Acceptance (all GPU-free; full table in the session report)

- Corpus margin, measured from every case's emitted launch grid: largest legitimate launch total 112.7 µs
  (`qwen3emb/gated-mlp-s512`, 1,572,864 blocks × 7168 issues), ~8.9× inside the guard; at 1e4/ns it would
  cross — pinned by test beside the per-thread 26.21 µs pin.
- The deterministic `post4096` replay (pinned twins + host-DB copy) elects **the statistics cut**: 9 kernels;
  the contribution producer's per-cell issues fall from ~2^34.6 to ~2^32.6 (the mHC statistics nest leaves it
  for two per-row `f4x4` pieces; only the carrier chain remains — launch floor 0.53 s, the next seam to elect);
  the contribution consumer stays a pure `w8x1` mma with no serial hidden-dim walk; the two 16384×256 `f4`
  sweep pieces of the measured 23.24 s route (5.3 s + 4.0 s) do not survive — two pieces two orders lighter
  per thread carry their role.
- Unit tests: blocks not cells; coverage does not divide the launch total; the refutation pin; the tie case
  (per-thread floors tie, the Σ elects the cut); the corpus margin pin.
- `make test` 4173 passed / 0 failed; `make lint` clean.

## Not claimed, and one new finding

The route is still not servable: the producer's carrier recompute stands, and — found while building the
per-piece table — **the post-#699 residual is emitted over a 2^56-thread linear grid** (every output sweep the
root's contractions read is promoted into its placement), so it cannot launch. That pathology is in #706's
election too and absent pre-#699; it is independent of both PRs and recorded in the serving plan as its own
item. The host measurement remains owed.

Core line balance (`git diff --stat 5ceb7118 -- emmy/`) is the two floor functions and the cell reader — the
capability this PR adds — plus their docstrings.

### DIFF
--- a/emmy/compiler/pipeline/ARCHITECTURE.md
+++ b/emmy/compiler/pipeline/ARCHITECTURE.md
@@ -409,7 +409,14 @@ reduce-partition coverage, at a per-issue constant conservative for any GPU cloc
 `post4096`'s fused 2^30-trip recomputation nest priced 4.29e-37 µs and beat every recomputation-free
 composed-cut arm until the bound priced it honestly — and its successor dominant piece (2^23 trips, 839 µs on
 trips alone, 16 % inside the guard, 13.21 s measured) rode that escape until the issue count priced its
-~16-statement weight-column-walk trips honestly. The guard is jurisdiction, not tuning —
+~16-statement weight-column-walk trips honestly. The same summand also carries a LAUNCH-TOTAL floor
+(`features.launch_floor_us`): every block evaluates the kernel's worst nest at least once, so a launch retires at
+least `blocks × S_ext_serial_cell_issues` issues — blocks read off the terminal placement's grid over the row's
+per-CTA output tile (or one cell per thread over the largest block a GPU runs) — at an issue rate no GPU reaches
+across all its schedulers (1e5 issues/ns). It is what separates two arms whose per-thread floors tie:
+`post4096`'s contribution producer recomputing the mHC statistics once per (row, a28) block against its
+statistics cut computing them once per row — 2000× apart per launch. Counting per output cell instead of per
+block was rejected as not a bound; `launch_floor_us` carries the record. The guard is jurisdiction, not tuning —
 the bound ignores launch overhead and memory traffic, so at ordinary magnitudes the model's ranking stands
 untouched, while a bound past the guard is un-servable whatever those effects are. A measured µs is never below
 the bound, so the clamp only ever lifts model garbage; the prior's own scoring surfaces are untouched, because
--- a/emmy/compiler/pipeline/search/features.py
+++ b/emmy/compiler/pipeline/search/features.py
@@ -475,6 +475,66 @@ def serial_floor_us(knobs: dict) -> float:
     return _covered_serial(knobs, "S_ext_serial_cell_issues") * SERIAL_ISSUE_FLOOR_US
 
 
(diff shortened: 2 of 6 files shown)


Slonegg (changes requested):
I think it is not needed. If the prior reports are gibberish, we either fix it or we should not rely on it. The golden YAML should contain all kernels; thus, the prior should not even trigger. Now there is a strict mode to catch that.

6erun:
Closed per review — the launch-total floor is not needed under the recorded-golden model (goldens carry every kernel; strict evidence keeps the prior from deciding). Branch left in place.

PR #711, opened 2026-09-04
Stamp the schedule space by what the enumeration reads, not how the regime is spelled

6erun:
## Summary

Since #704, `compile --golden PATH --realization NAME` publishes the records' shared input regime
(`pinned_knobs(shared_regime_pins(...))` in `handle_compile`), which for a standard-lane golden is
`EMMY_FAST_MATH=False`. That pin changes nothing the enumeration reads — `precision_pin(...) is True` is False for OFF
and unset alike — but `knob.schedule_pin_fingerprint` folded the raw env string into the schedule-space stamp, so the
pinned compile minted a different `Fork.pool_id` than the unpinned deploy for the same pool. `pool_id` seeds the
budgeted pool's draw (`greedy._descent_sample`), so every prior-decided kernel above `_POOL_BUDGET` ranked a different
drawn subset. On the DeepSeek-V4 `post4096` twin that put a binder-refused row at the top of the residual root's draw;
with only the pin neutralized the election matched the reviewed 10-kernel plan byte for byte.

Only `handle_compile` derives a regime from golden records. Serve / deploy installs no pin: the regime is the process
environment (`EMMY_FAST_MATH=1` for the `fm` lane, unset otherwise) and `regime_live` filters the golden rows to it.
The release gate and the realization corpus publish the regime through `pinned_knobs` the same way `compile` does.

The regime pin is the compile's regime, and the binder's decline of an offered row is the documented
`RuleSkipped(reject=True)` path, so neither is the defect. The fix: the fingerprint spells the precision gates by
effect — one entry per gate `precision_pin` resolves ON, nothing for OFF or unset — which is exactly what
`classic_forks` reads. Family pins stay raw, since each restricts a domain.

## Test

`test_an_off_precision_gate_stamps_the_same_space` beside the existing gate-stamp test: OFF spellings equal the unset
fingerprint, a real tile's `pool_id` is unchanged under the published regime, and `FAST_MATH=1` equals both gates ON.
Fails on `main` with `(('EMMY_FAST_MATH', 'False'),) != ()`.

## Not touched

- The retry policy behind the fused fallback (owned elsewhere).
- `backend/pack.py` keys the pack on raw `precision_pin` values — the same spelling sensitivity, but warm and boot spell
  the env alike, so it only costs a rebuild.

## Verification

- `make test` merged with main through #722: 4149 passed, 1035 skipped (no CUDA), 21 xfailed, 0 failed.
- `make lint` clean. `knob.py` 17/17 lines; the two ARCHITECTURE files grow by 7 lines to record the invariant.


Slonegg (approved):
I don't have enough context to understand the change, but it is probably ok.

PR #712, opened 2026-09-04
Key measurements on kernel identity, unpack the store's columns

Slonegg:
Measurements are now stored and looked up by what a kernel IS, rather than found by scanning a
table and matching on the feature set the kernel happened to be stamped with. The identities the
store keys on became two small value objects whose fields are the table's columns, and the four
op-inventory tables, the lowering edge table and the perf table collapsed into three: context, op
and measurement. Identity values themselves did not move, so nothing checked in had to be
restamped. This is steps 1 and 2 of `plans/measurement-store-identity-schema.md`; the pick itself
is not rewired yet, so the tune DB's evidence tier is switched off in the meantime and says so.

## Why

The deploy found a measured row by its `S_*` feature set: a full scan of `perf` per process,
memoized on the DB file's mtime, matched by a subset rule that exists only because features drift.
Different kernels never share a signature, so the signature was a stand-in for identity — and
identity is what should be stored and queried. Meanwhile the DB held three notions of what a
kernel is beside the one a deploy matches on, and keys were assembled by hand in about ten places.

## The identity module

`emmy/compiler/identity.py` — two frozen value objects, each carrying its own `digest` as a
computed property over its own columns:

- **`Regime`** — where a measurement holds: `gpu`, `sm_major`, `sm_minor`, `opt_level`,
  `nvcc_flags`, `pins`. The card is new to the key: µs is a per-card fact, and `perf` had no card
  column, so a measurement on one sm_89 card answered a deploy on another.
- **`OpIdentity`** — what a kernel is, knob-free: `dialect`, `body`, `io`.

`Context.structural_key` and `Op.identity_key` route through them. It lives at `emmy/compiler/`
rather than under `search/` because `context.py` and `ir/base.py` sit below `pipeline` and both
produce identities.

The digest is computed, never stored beside the fields and never passed in — a key a caller could
supply is a key that can disagree with the row it sits on. (SQL cannot help: SQLite has no
sha256, a generated column cannot be a `PRIMARY KEY`, and the fold's input is a Python-side
canonical rendering.)

**Identity values are unchanged.** `OpIdentity.io` holds the io fingerprint itself (the column
renders it via `io_json`) and the lattice fold stays flat — `digest(body, io, knobs)`, never a
digest of a digest. Verified against the corpus: the live fork identity equals the checked-in
stamp, so none of the 208 case files needed regenerating.

## Schema

```sql
context (digest PK, gpu, sm_major, sm_minor, opt_level, nvcc_flags, pins)
op      (digest PK, dialect, body, io)
measurement (context, op, decision, status, us_median, us_min, us_max,
             n_samples, measured_at, captured, error,
             PRIMARY KEY (context, op, decision))
CREATE INDEX measurement_fork ON measurement (context, op);
```

`measurement`'s key is also its keep-best key, and it IS the per-fork query: one indexed lookup on
the identity a fork is offered on. What the store no longer keeps is reconstructed on demand — the
`S_*` features by featurizing `op.body`, the `H_*` values from the regime, the kernel source and
geometry by re-lowering the body under the decision, which is what the cubin cache already keys on.

`lowering` goes because every pre-schedule stage of a chain shares one `op` digest, so
`best_per_op_time` reads the terminal's row directly instead of walking parent-to-child links.

**No migration, by necessity.** A `perf` row's `op_key` is a digest of rendered CUDA source, and
nothing re-derives the Loop-IR body it came from. Schema v5 drops the pre-identity generation
whole, with a row count in the log. A tuned machine re-tunes.

## Three findings from verifying the join

The join this design rests on is "the identity a kernel is measured under equals the identity of
the fork that offered it". Checking it on corpus cases moved the design three times:

1. **`OpIdentity`'s digest must not fold `dialect`.** A golden's identity is minted by lifting a
   recorded target to Loop IR; the live compile mints it at a Tile fork. Fold the stage tag in and
   that join dies. It stays a descriptive column.
2. **A `CudaOp` keys on the newest *pre-schedule* op of its chain** — its own `TileOp`, else the
   fused `LoopOp` — body and io alike. Not the nearest predecessor: the intervening `KernelOp`
   body is the schedule already materialized, so it keys apart from the tile term the fork
   decided. Not the oldest: that prices a split's piece as the parent it was cut from. And not
   body-from-chain with io-from-here: the cuda passes rename and re-slab buffers, which splits the
   join at its io half.
3. **`Regime.pins` carries every marker BOOL**, not just the precision gates — it is the exact
   complement of the decision's knob row, which drops them. Otherwise a kernel measured under
   `EMMY_VECTORIZE_LOADS=0` shares a key with the same kernel measured under the default, and
   keep-best silently picks between two different pieces of code.

## What this deliberately breaks

The measured-evidence tier is **off** until the per-fork query lands (step 3):
`_db_measured_index` returns empty with a one-per-process warning. Its rows were findable only by
feature set, and the features are no longer stored beside the measurement. A greedy compile
decides from the recorded goldens and the prior meanwhile — a real loss of evidence on a tuned
machine, which is why it announces itself rather than failing silently into the `except Exception`
that used to wrap the scan.

## Verification

- Realization corpus (`offered` / `realized` / staleness): **409 passed, 25 skipped, 9 xfailed**,
  no case file touched.
- `tests/compiler/pipeline/search` + `test_knob`: **221 passed, 8 failed, 1 collection error** —
  every one of them a test calling a removed API (`lookup_perf`, `_db_measured_index_build`) or
  building a bare `CudaOp` with no rewrite chain, which now correctly has no identity to store a
  measurement under. Those fixtures are the next commit's work.
- Store round-trip: digest equals the digest of its own columns, `S_*` stamps drop out of the arm,
  keep-best holds, `measurements()` and `best_per_op_time` return the arms.
- `make lint` clean. Full `make test` was still running when this was opened; the number above is
  the targeted subset.

## Line balance

`git diff --stat main -- emmy/`: **16 files, +386 / −713** — net −327 under `emmy/`, so the new
capability arrives smaller than what it replaces.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -32,7 +32,8 @@ relevant `ARCHITECTURE.md` before answering.
 - `EMMY_TUNE_DB` environment variable (optional) — overrides the default tuning SQLite cache path
   (`~/.cache/emmy/autotune.db`). `emmy tune` reads from / writes to this path. NOTE: greedy `compile` / `run`
   resolve forks through the deploy evidence hierarchy — the live card's recorded goldens first (the repo-shipped
-  verified tier; consulted, never trained on), then measured reservoir/DB evidence, then the global `Prior` (the
+  verified tier; consulted, never trained on), then measured reservoir evidence (the tune DB's own tier is off until
+  the per-fork identity query lands), then the global `Prior` (the
   online prior with its offline cold-start fallback; the old `_best_fork` DB→fork replay was removed). The online prior
   is a separate JSON checkpoint (`EMMY_ONLINE_FILE` → `~/.cache/emmy/online.json`; legacy `EMMY_PRIOR_FILE` still
   accepted) that `tune` writes and `compile` / `run` read. Use the README architecture index for the prior and
--- a/GLOSSARY.md
+++ b/GLOSSARY.md
@@ -247,11 +247,11 @@ describe how a term is used in Emmy; they are not meant to replace a full textbo
 - **Realize** — A recorded configuration *realizes* when the compiler, at the point where it makes that choice, offers
   a candidate matching the recording. A configuration that realizes nowhere cannot be deployed, however good the
   measurement stored with it.
-- **Regime** — The settings a measurement was taken under, or that a compile is running under: chiefly the compiler
-  optimization level (`-O3`, the deployable one and the only one anything is measured in) plus whether fast-math
-  approximations are enabled. Measurements taken in different regimes are not interchangeable, which is why tuning
-  measures in the regime it deploys into. A lower optimization level remains available as a compile-speed option for
-  the test suite; it is not a measurement lane, and nothing measured under one is read by a deploy.
+- **Regime** — The settings a measurement was taken under, or that a compile is running under: the card it ran on,
+  the compiler optimization level (`-O3`, the deployable one and the only one anything is measured in), and whether
+  fast-math approximations are enabled. Measurements taken in different regimes are not interchangeable, which is
+  why tuning measures in the regime it deploys into. A lower optimization level remains available as a compile-speed
+  option for the test suite; it is not a measurement lane, and nothing measured under one is read by a deploy.
 - **Candidate** — One complete set of choices that the compiler could use.
 - **Greedy selection** — Choosing the candidate that currently appears best without exploring alternatives during
   normal compilation.
--- a/emmy/commands/eval.py
+++ b/emmy/commands/eval.py
@@ -19,7 +19,7 @@ Five subcommands:
   ``(kernel, error)`` with the knob values shared by every failing row.
 
 The ``eval knobs`` regret analysis: for each kernel (grouped by the kernel C
-identifier extracted from ``cuda_op.pretty``), compute per-knob regret:
+identity the rows are keyed on), compute per-knob regret:
 
     regret[K] = max(best_us | K=v) / min(best_us | K=v)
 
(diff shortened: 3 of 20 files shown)


Slonegg:
Not ready — closing while the greedy-pick regression below is understood.

`make test` on the full suite (not the subsets reported in the description) shows 23 failed + 1 collection error. Most are fixtures calling removed APIs, but at least three are real behaviour changes, and the first has been isolated: reverting only `emmy/compiler/context.py` makes `test_raster_default_is_the_flat_order` pass again, so folding the card and the input pins into the regime — and thus into `Context.structural_key` — changes a cold greedy pick (RASTER `gm8` where the conservative `""` is expected). `test_working_shared_placement_pins_select_the_exact_compile_target` and `test_unlowered_terminal_is_bench_fail_despite_cached_residual_kernel` are still unclassified; all three pass on main against the same DB, so they are not data loss.

Branch is kept. Two other things worth recording:

- Two `tests/architecture/test_layering.py` guardrails caught a genuine design error in the first version (a second public `Op.identity` spelling, and `_body_identity` defined outside its home). Fixed by moving the chain walk into `OpIdentity.of(op)`, which let `ir/base.py` and `ir/cuda/ir.py` revert to main entirely.
- The description's line balance was wrong: `identity.py` was untracked when `git diff --stat` was run, so the real figure is 15 files, +499 / −685 (net −186 under `emmy/`), not net −327.

PR #713, opened 2026-09-04
Speed up tests by avoiding cold compilation in gen runner

Slonegg:
The serving tests were spending 35 of the GPU lane's 54 minutes re-deriving compiler decisions
they never assert on. They now compile against a checked-in golden, so the lane costs about two
minutes instead of thirty-five, and compiles the same thing on every machine. No test was deleted
and no assertion was weakened; two that were marked as expected failures now pass.

## Why the lane was slow

All `@cuda` tests are pinned to one worker, so that chain's serial total is the suite's wall clock.
`tests/serving` was 35.3 of its 54.3 minutes, and `test_gen_runner_gpu.py` alone was 23.7.

A profile of one of its tests put 206s of 227s under `_price_graph` / `_descent_sample`:
**979,402 `schedule()` calls and 1,531,703 `_local_support()` calls** for a 3-layer, hidden-64
model. Attributing the profile by module:

| bucket | self-time | share |
|---|---|---|
| schedule enumeration | 72.8s | 38.6% |
| python/stdlib under it (`hash`/`id`/`isinstance`) | 51.5s | 27.3% |
| rest of compiler | 41.5s | 22.0% |
| **cupy (device API)** | **0.0s** | **0.0%** |
| **nvcc** | **0.0s** | **0.0%** |

The device was idle. `make test` scopes no golden (`EMMY_GOLDEN_FILE=`), so every one of these
tests booted fully cold and re-derived every pick by descent sampling — while occupying the single
serialized GPU worker. Cold deploy is the prior's contract and has its own tests; these ask whether
the runner stitches, dispatches and allocates correctly.

## What changed

`helpers.RUNNERS` spells every runner shape the lane builds, in one place. `regen.py` captures each
compiled graph at the plan-cache seam and writes a golden covering exactly those shapes, completed
to one entry per kernel through the realization corpus's own `helpers.complete`. The `built` /
`build_runner` fixtures build inside `sole_evidence`.

**Strict evidence is what makes this safe rather than merely fast.** A fork no row decides raises
`EvidenceError` naming the kernel, so an incomplete or stale golden fails loudly instead of quietly
restoring the search. The tests and the regen read the same table, so the golden cannot drift out
of step with the shapes the tests build; if it does, the next run says which kernel.

| | before | after |
|---|---|---|
| `test_gen_runner_gpu.py` | 23.7 min | **41 s** |
| `tests/serving/generation/` | 9 min 15 s | **1 min 56 s** |
| `host_mapped_embed_table…` | 227 s | 0.85 s |
| `pack_key_separates_quantized_rungs` | 210.9 s | 12.5 s |
| `layers_share_activation_arena` | 121 s | 1.28 s |

Ten shapes intern to 35 distinct graphs — the four Qwen3 ones share six between them — which is the
sharing `PlanTemplateCache` exists for and could not reach while `from_model` built its own per
runner. That is the only `emmy/` change: `from_model(..., plan_cache=None)`, +9/-2 lines.

## Two expected failures closed

Both were marked xfail because "the offline-weights refit steers this tiny fp32 shape onto a
TMA-staged pick with run-to-run last-ulp instability". That is the reproducibility cost of a cold
pick: it resolves through the machine-local tune DB and online prior, so what the test compiled
depended on the box it ran on. The golden decides the pick now, the shape lands on one schedule
every run, and both bit-parity contracts hold. The staging race itself is untouched — it is a
compiler finding, not a parity gap, and no longer what these tests measure.

## Cleanup

`test_gen_runner_gpu.py` loses 280 lines (957 → 677): fifteen copies of a fifteen-line config block
become one table. The pack tests deliberately do **not** take the session plan cache — their claim
is that the first boot compiles and the second loads the pack, which a warm template cache would
decide for them.

Two things I looked at and left alone. `test_matmul_mma_f16acc_coverage` (3 × 25 s) looks like a
corpus duplicate, but its docstring already reasons about that: the static-M column of each
transport *is* a corpus case, and it keeps the symbolic-M run plus a two-config sibling bound the
corpus cannot express. And the realization corpus itself — 412 cases for 5.2 minutes, less than one
old serving test — is exactly the materialization coverage the GPU lane exists for.

## Budget

The GPU chain goes from 54.3 to about 21 minutes. The remainder is 19.0 minutes of genuinely
device-bound work (the corpus plus e2e), so the last mile is a question about the 13.7 minutes of
non-corpus e2e tests, not about deleting coverage.

## Verification

`tests/serving/generation/`: 56 passed, 29 skipped, 1 xfailed — green under strict evidence, which
is also the proof the golden is complete. `make lint` clean. The full suite has not been re-run on
this branch yet.

https://claude.ai/code/session_018LeM7LGM5iUqingymHUEab


Slonegg:
## Full suite verified on this branch

**54:07 → 21:27.** 4853 passed, 276 skipped, 49 xfailed.

| | before | after |
|---|---|---|
| full suite | 54 min 07 s | **21 min 27 s** |
| `@cuda` chain (the wall clock — one worker) | 54.3 min | **21.7 min** |
| └ `tests/serving` | 35.3 min | **4.1 min** |
| └ everything else | 19.0 min | 17.6 min |

The chain measures 21.7 min against the 20-minute target. What is left is genuinely device-bound:
the 412-case realization corpus at 5.0 min, `test_matmul_coverage` at 4.3, `attention_split_gpu` at
2.1.

### One failure, pre-existing

`test_realization[qwen3emb/gated-mlp-s128.yaml-built]@cuda` — `worker 'gw3' crashed`, no traceback.
The identical signature appeared on `main` before any of this work (gw32, then gw1), and it passes
in isolation. It is a hard process death late in a ~700-test CUDA chain, unrelated to this PR, and
worth its own investigation.

### Durations restamped

The converted tests' baseline entries were written when they deployed cold — 90.2 s for the dense
stitch against a lane that now measures 4.65 — so CI would have planned its buckets around work that
is not there. Restamped from this run. Two rows moved *up*, and both are the session fixture working
as intended: the shape built first pays for every build behind it, so `rider_split` goes 6.5 → 33.0
while the lane's total falls.

6erun (approved):

Slonegg:
## Budget met: the GPU chain measures 19.1 min

Converting the last cold serving file (`test_attention_split_gpu.py`) brings the lane in under the
20-minute target.

| | before | after |
|---|---|---|
| full suite | 54 min 07 s | **18 min 53 s** |
| `@cuda` chain (the wall clock — one worker) | 54.3 min | **19.1 min** |
| └ `tests/serving` | 35.3 min | **1.8 min** |
| └ everything else | 19.0 min | 17.4 min |

`tests/serving` is now 9% of the chain rather than 65%. The remaining 17.4 min is genuinely
device-bound work — the 412-case realization corpus at 5.0 min leads it, which is the materialization
coverage the lane exists for.

4853 passed, 276 skipped, 49 xfailed. No durations-gate errors.

### The one failure is still the pre-existing one

`test_realization[qwen3emb/gated-mlp-s128.yaml-built]@cuda` — `worker crashed`, no traceback. Third
occurrence across four full runs, including two on `main` before any of this work (gw32, gw1, now
gw3), and it passes in isolation. Unrelated to this PR; worth its own issue.

### Last file converted

The attention-split carves never reach a runner and trace with a symbolic token axis, so their
programs are their own — they get a `WRAPPERS` table beside `RUNNERS`, which the regen traces
directly.

| test | before | after |
|---|---|---|
| `test_gemma_post_wrapper_compiles_and_runs_dynamic` | 77.2 s | 3.50 s |
| `test_post_wrapper_compiles_with_shared_dim` | 35.7 s | 1.88 s |
| file total | 2.1 min | 12 s |

The lane's helper moved from `serving/generation/` to `serving/` with it, which is what
`tests/ARCHITECTURE.md` asks for — a reusable callable lives at the nearest directory shared by its
callers, and it now has callers in both. The golden is renamed to match (`gen_runner` → `serving`):
it covers the lane, not one module.

PR #719, opened 2026-09-04
Revert the serial-work clamp: the kernel-set Σ prices rows as estimated, never bounded

6erun:
Follows the review decision on #706 / #708 ("remove the lower bound hack — if the prior kicks in, the golden is not replayed or not present for the kernel"): this removes the one that was already merged, #702's clamp.

**Removed.** `features.serial_floor_us` and its per-trip constant, the enforcement guard `_SERIAL_FLOOR_ENFORCE_US`, and the clamp in `policy/greedy._resolved_price` — the kernel-set Σ now sums each row's price as stamped or estimated, with no hand-written bound. Their tests (three) and the doc paragraphs describing the bound go with them; the pipeline ARCHITECTURE now states the rule the other way round: where the prior ends up deciding a production election, the defect is the missing recorded golden or measured row for that kernel, and the fix is to record it.

**Kept from #702**, deliberately: the `S_ext_serial_cell_work` stamp and the `D_serial_cell_work` fit signal (the prior can *learn* serial work — that is the "fix the prior" direction, and dropping the stamp would force another corpus restamp and schema bump for no behavioral change); the disqualification-signature match that survives featurizer vocabulary growth (without it a tune DB's `bench_fail` rows stop binding); and SearchDB schema v4. Say if the stamp should go too.

Core line balance: `emmy/` −71 lines. `make lint` clean. `make test`: 4141 passed; the 2 failures are `attention/sdpa-hd128-softmax-v-mma.yaml` at `offered`/`realized`, which fail identically on `main` since #715 (being handled separately).

The DeepSeek serving plan's round-three item 3 records the revert; item 4 already describes the recorded-golden path.

### DIFF
--- a/emmy/compiler/pipeline/ARCHITECTURE.md
+++ b/emmy/compiler/pipeline/ARCHITECTURE.md
@@ -403,18 +403,11 @@ is reached only by pointing `EMMY_OFFLINE_FILE` (or `--offline-file`) at it. Eac
 `provenance.scope`; read that before drawing conclusions from one, because a scoped artifact has no reason to beat
 the shipped weights outside the slice it was fit on.
 
-The proxy stays uncalibrated, and the one consumer that needs absolute µs enforces a physical bound instead: in
-the kernel-set Σ (`policy/greedy._resolved_price`), a summand whose serial-work lower bound
-(`features.serial_floor_us` — the row's per-thread serial trips, i.e. the nest-aware `S_ext_serial_cell_work`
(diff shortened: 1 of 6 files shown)


Slonegg (approved):

6erun:
CI's two failures are `attention/sdpa-hd128-softmax-v-mma.yaml` at `offered`/`realized` — the case #715 added, which fails identically on `main` (same messages, no other failures). #717 records it as the open gap it is; once that merges I'll merge `main` here and this goes green. Nothing in this branch touches that case.

PR #723, opened 2026-09-05
Promote only a kernel's shared output sweep, not a sibling nest's axis (#699 regression)

6erun:
## The regression

Since #699 the DeepSeek-V4-Flash `post4096` serving twin's residual root cannot launch: every output sweep of the
root is promoted into its placement, and the grid becomes the product of all of them. Replays of the golden
realization `post4096.k_linear_softmax_matmul_mean_reduce_9716a1.f86b6dbe35b7.m4096` (sm_70, the tune-DB copy as
the only evidence, no online prior):

- **before #699** (`d64622a5`): the residual `k_linear_softmax_matmul_mean_reduce_9716a1` launches as
  `int _gid = …; int a0 = _gid;` — 16 blocks × 256 threads, every output sweep (`a10`, `a14`, `a17`, `a20`, `a27`,
  `a35`, …) a serial loop inside the kernel.
- **on `origin/main` (`30480ae5`)**: the same residual decodes a 2^56-thread linear grid —
  `long long _gid …; int a0 = _gid / 17592186044416; int a10 = (_gid / 4398046511104) % 4; … int a34 = (_gid) % 4096;`
  — a launch of **281,474,976,710,656 CTAs × 256 threads** (`a0×a10×a16×a19×a19_1×a26×a26_1×a34 =
  4096·4·4·4096·4·4096·4·4096 ≈ 7.2e16` cells). `gridDim.x` tops out at 2^31−1: the kernel cannot launch.

## Root cause

`TileOp.__post_init__` promotes an output sweep into `place.free` when any nested contraction operand reads it
(#648, widened to nested sites in #675). What that rule could SEE was bounded by extraction: before #699,
`extract_output_specs` set `OutputSpec.sweep` only for the "legacy single output sweep" — the one non-reduce loop at
the tail of the stream holding every store (rms/softmax's normalize sweep), and only when no flat tail store stood
beside it. Every other output loop — a sibling nest, a nested nest — became a `ProjectionRegion` whose specs carried
`sweep=None`. The promotion rule therefore only ever met the kernel's one shared output sweep.

#699 retired `ProjectionRegion`: every output loop lifts to a zero-axis term evaluated over its sweep coordinate and
its stores reach the boundary as sweep specs; #706 then spelled `OutputSpec.sweep` as the store's whole loop-nest
path and the promotion block began iterating every axis of every store's path. On the post4096 root that domain is
four sibling output nests — `(a10,a16)`, `(a19)`, `(a19_1)`, `(a19_1,a26)`, `(a19_1,a26_1,a34)` — and each axis is
the genuine output coordinate of the contraction hoisted out of that nest (`linear_4_wt[a27, a26]` reads `a26`, the
`acc49` slab reads `a26_1` and `a34`, the score matmul reads `a10` and `a16`). Nothing is captured-but-unread: the
operand edges' `free_axes` are exact. The rule itself over-applies — promoting a sibling nest's axis evaluates every
OTHER nest once per cell of it, so promoting all seven made the placement their product. The cut consumer inherits
`place=tile.place` verbatim, so after every contraction was cut out into producer pieces (which size their own grids
from what the seam reads — `_cut.py` already anticipated "an output sweep the grid carries for a sibling's sake"),
the residual — a pointwise store kernel — kept the 2^56-cell grid.

## The fix, and why it is the semantic one

Promotion is sound exactly when the axis is on EVERY store's sweep path — the kernel's one shared output sweep, the
only thing the pre-#699 `sweep` field could name: hoisting it replicates nothing but the statistics evaluated ahead
of the sweep (the per-cell replication the invariant describes). A sibling nest's axis is not such an axis, whatever
reads it. `TileOp.__post_init__` now takes the promotion candidates as the intersection of the stores' sweep paths
and promotes a shared axis exactly as before when a contraction operand reads it. A contraction under a sibling nest
stays under its sweep loop, where `Fold.lower` already places any term reading the axis, and the placement fork cuts
it out at its own free coordinates. No cap on grid size, no pricing change; `emmy/` Python is net −1 line.

`tests/compiler/passes/test_total_lift.py::test_sibling_q_and_kv_regions_total_lift_with_separate_outputs`, written
by #699, pinned the over-promotion for the fused q/k/v shape (`free == [m, q, kv]`, both sweeps stripped): that is
the same pathology at `4·2` instead of `4096·4·4·4096·…` — the q contraction evaluated once per `kv` cell and the
k/v pair once per `q` cell. Its placement expectations now read `free == [m]` with both sweeps kept; the one-kernel
claim, the sibling-loop `loop_body` and the `{m,q}` / `{m,kv}` seams it asserts all still hold.

## Replay after the fix (same golden, same evidence, deterministic)

| piece | launch | launchable |
| --- | --- | --- |
| `…__place_10cf5b720b` (`acc49` contribution; was `843da639e9`) | 67,108,864 × 128 | yes (unchanged) |
| `…__place_f5f8bc94ac` (was `05983f700d`) | 67,108,864 × 128 | yes (unchanged) |
| `…__place_895f05e860` (was `bed580ec13`) | 4 × 256 | yes |
| `…__place_947deab1a4` (was `ee3ad8be8b`) | 16,384 × 128 | yes |
| `…__place_1962fa992f` (was `ab50fc9ead`) | 1,024 × 128 | yes |
| `…__place_a97bb1167e__place_8a807213cf__place_ee37df43ab` | 16,384 × 128 | yes |
| `…__place_921364b0af` (was `b2c9ead08b`) | 65,536 × 256 | yes |
| `…__place_a97bb1167e__place_8a807213cf` (was `ab7fe4b606`) | 512 × 256 | yes |
| `…__place_a97bb1167e` (new: the `mul_16` gate cone, cut at `a26 × a0 × a19_1`) | 40,960 × 256 | yes |
| **residual `k_linear_softmax_matmul_mean_reduce_9716a1`** | **16 × 256**, `int a0 = _gid;` | **yes** |

The residual is back to the pre-#699 shape (grid `a0` alone, every sibling nest a serial loop, no contraction inside).
No piece of the route is un-launchable; the two 67,108,864-CTA contribution pieces are identical before and after and
belong to the other known blocker (the producer recomputing the carrier chain per cell), out of scope here.

## Not a corpus case

`tests/compiler/realization/ARCHITECTURE.md` reserves a case for a schedule that should be realizable and is not.
This residual `offered` and `realized` — the row decodes and lowers — and fails only at launch on the exact card, a
stage the corpus cannot observe GPU-free; the placement invariant is pinned by the synthetic Tile IR test beside the
existing promotion tests instead.

## Checks

- `make test` recipe (`EMMY_NVCC_FLAGS="-Xcicc -O1" EMMY_GOLDEN_FILE= pytest tests/ -n auto --dist=loadgroup`):
  **4150 passed, 1035 skipped, 21 xfailed** (no CUDA on this machine); corpus staleness clean, no regen needed.
- `make lint` recipe: `ruff check` and `ruff format --check` clean.
- `make test-goldens` recipe: **11 failed** — every checked-in golden file. The identical gate on `origin/main`'s own
  `ir.py` (`30480ae5`) fails the same 11 files with byte-identical failing-row sets, per-file counts
  (`8/8, 9/9, 10/10, 10/13, 24/44, 48/73, 186/194, 38/63, 153/240, 226/352, 267/279`) and reasons — rows such as
  `pointwise.n4096`, `rms_norm.k2048`, `reduce.k2048` and `matmul.square.512` (no output sweep, no contraction) are
  among them, which this change cannot reach. Pre-existing on `main`, not re-recorded here.
- `git diff --stat origin/main -- emmy/`: `ir.py` +5/−6 (net −1), `ir/tile/ARCHITECTURE.md` +18/−9 (the promotion
  paragraph restated).
- `plans/deepseek-v4-emmy-serving-v100.md`: the Stage 0 round-three blocker clause now records the fix; `plans/`
  holds 10 files, none executed by this change.

### DIFF
--- a/emmy/compiler/ir/tile/ARCHITECTURE.md
+++ b/emmy/compiler/ir/tile/ARCHITECTURE.md
@@ -101,11 +101,18 @@ the placement fork; a Loop IR fusion or emission workaround sees one kernel at a
 construction. Copies that differ in captured axis names cannot share an object and stay with value clustering; copies
 that differ in exposed result names stay distinct because unifying them would rename their consumers.
 
-An output sweep used by any nested contraction operand is promoted into the Tile's free-axis placement. Promotion
-expands the enclosing-axis context, so construction normalizes the Fold tree once more under that final scope; one
-construction and a reconstruction therefore expose the same closed operand edges and placement seams.
-The invariant also applies when a schedule row constructs or reloads an already-mapped Tile: promotion extends the
-grid in lockstep with the free axes, so per-cell replication never mistakes the swept coordinate for an SSA name.
+The kernel's SHARED output sweep — an axis on every store's sweep path — is promoted into the Tile's free-axis
+placement when any nested contraction operand reads it: the coordinate is that contraction's output coordinate, and
+promoting an axis every store rides replicates nothing but the statistics evaluated ahead of the sweep. A sibling
+output nest's axis is never promoted, however many contractions read it: the other nests do not ride it, so promoting
+it would evaluate them once per cell — DeepSeek-V4 post4096's residual root holds four sibling nests, and promoting
+all seven of their axes made its grid their product, 2^56 cells, which no launch can cover. A contraction under a
+sibling nest stays under its sweep loop, where `Fold.lower` places it like any term reading the axis, and the
+placement fork cuts it out at its own free coordinates. Promotion expands the enclosing-axis context, so construction
+normalizes the Fold tree once more under that final scope; one construction and a reconstruction therefore expose the
+same closed operand edges and placement seams. The invariant also applies when a schedule row constructs or reloads
+an already-mapped Tile: promotion extends the grid in lockstep with the free axes, so per-cell replication never
+mistakes the swept coordinate for an SSA name.
 
 **Storage-decode factors hoist to the epilogue.** A product argument whose cone is a STORAGE DECODE
 (`ElementwiseImpl.decodes` — the trait, never an op-name list) times factors constant along the fold
@@ -134,8 +141,9 @@ as a supported slow path.
 
 A node's own canonical forms are formation's (`Fold.__post_init__` orients a bilinear term A-first, `Lambda` orders
 its body); `TileOp.__post_init__` applies the tree-wide ones and the legacy output-sweep-to-free-axis adjustment
-whenever a contraction operand reads the sweep axis. The contraction may be the root compute node or a later site in
-the Fold tree; in either case the coordinate belongs in kernel placement rather than a post-compute output loop.
+whenever a contraction operand reads the kernel's shared output sweep. The contraction may be the root compute node or
+a later site in the Fold tree; in either case the coordinate belongs in kernel placement rather than a post-compute
+output loop.
 A specification's `sweep` is the output loop NEST the store rode in the source, as an outermost-first axis path — the
 one fact extraction destroys, since a write's index names its coordinates but not the order its loops nested in.
 Consecutive output specifications over one path reconstitute one loop nest, sibling sweeps sibling loops, and a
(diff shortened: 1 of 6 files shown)


6erun:
Merged `main` (#710) — clean, no overlap with the tile IR; local suite 4152 passed, lint clean.

The CI failure on the merge commit was the durations gate, not a test: `test_an_off_precision_gate_stamps_the_same_space` (#711) has no `tests/durations.json` entry and ran at exactly 5.0 s on CI, so the session-end staleness gate failed the run (pytest itself reported 4152 passed). This branch's first run passed because the same test landed under 5 s that time. Added the entry by hand in sorted position at the CI-measured value (7893bc7f), the way #717 did — `make test-durations` on a GPU-less box would replace the whole baseline.

Slonegg (approved):

PR #726, opened 2026-09-05
FlashAttention-2 out of a blocked twisted carrier

Slonegg:
## What this is

FlashAttention-2, derived. A twisted carrier's stream is split into blocks and the fold
re-associated over the two levels, which separates the rescaling ⊕ from the base monoid and leaves
the expectation channel reading as a contraction — a site the tensor-core tier can take. The
emitter then realizes it as one kernel: `Q·K` into C fragments, a fragment row reduce for the block
pivot, the weight applied to those fragments, `P·V` against the staged V slab.

Nothing is recognized. The rescale coefficient is read out of the stored combine and the value it
multiplies is the fold's own lift result, so no recipe is consulted and no operation family is
matched; Welford declines by the same reading, because its M2 is a square.

## Measured (RTX 5090, sm_120, f16 `F.scaled_dot_product_attention`)

| shape | main (cold greedy) | this branch (cold greedy) | this branch (pinned best) |
| --- | --- | --- | --- |
| `(1,4,128,32)` | 197 µs | **5.5 µs** | **3.3 µs** — 1.23× eager's 4.0 µs |
| `(1,8,512,128)` | 1490 µs | **150 µs** | not swept |

`grep -c mma.sync` on the emitted kernel is 4 where main's is 0. Accuracy `max_diff=4.9e-4`.

## The shape of the change

**Blocking is a normalization, not a decision.** `035_block` runs the rewrite; it is parameter-free
and idempotent (every axis it installs carries a `Window` receipt), so it opens no fork, adds no
schedule family, and puts no key on any row. The width is the FORM's — `block_width` is the largest
power-of-two fraction of the extent within `MAX_BLOCK` — and the row chunks the block with its own
`bk` exactly as it chunks any other K.

An earlier draft on this branch left the width a SYMBOL for the row to bind. That is what the memo
described, and it does not work: everything downstream then sizes itself against `Var(__blk_a2)`.
No warp tile was offered at the score (its N was symbolic), no `smem` compute fill at `P·V`, no
partition could count trips, the prior priced a strided axis it could not read (a cold pick went to
1.4 s/iter at 512×128), and an unmapped `TileOp` leaked the symbol into the kernel SIGNATURE as a
runtime `int`.

**Only a carrier a block pays for is blocked.** A contraction's block is already spelled by `bk`
and a plain reduction's partition by `REDUCE`. A plain online softmax is twisted and still
declines: its channels are sums of the weight itself, so nothing in it comes out bilinear and the
block would buy a second pass over the stream for nothing. Blocking reads its own result and leaves
such a carrier alone.

**Two passes over a block is the floor, and the channels are the second of them.** They share one
binder and one weight instance, and `Fold.lower` places sibling reduce terms over one axis in one
loop (`_scope`), so what they share is emitted once — three passes over the block down to two. The
fuse is guarded exactly as the Loop-IR merge is: a loop reading what the loop above it defines is
reading a FINISHED accumulator.

**One pass, in the emitter.** Inside a block the pivot's copy of the score and the weight's are the
same value under two coordinates, so `_fold_staged` rewrites the block into one spelling (binders to
the stream, the outer coordinate to zero, every lambda re-closed so a rename cannot reorder its
trailing params) and the score is emitted once and read twice. That is what makes this FA2 rather
than a two-pass softmax.

**The block's two cross-site equations.** The channel's K-step consumes the block in ONE trip and
the score's fragment grid covers it in one cover. Both are the kernel binder's own — a fragment grid
that does not cover the chunk it stores has no seam — and the projection offers only the tiles that
satisfy them, so a row the binder must refuse is never offered. The scalar tier is offered inside a
block only when no warp tile fits at all.

**`035_block` runs after `030_cut`.** A piece the cut mints is a different carrier from the one it
came out of: attention cut at its value channel leaves a statistics piece whose channels are sums
of the weight, and blocking ahead of the fork handed that piece a block it paid a second pass over
the stream for — 18.8 ms on a 512-key head where the fused kernel is 128 µs. A block is still a
working set inside ONE kernel, so no seam evaluated over a block coordinate is cuttable either.

## Defects this exposed on the way

- **`Fold.bindings` is per result COMPONENT, not per operand.** Three sites in `block.py` indexed
  the operand tuple with a binding position, which is the same number only while every edge exposes
  one result. A fused cone exposing three crashed blocking outright.
- **`scheduled_fold_contraction` and the whole fragment-Fold path were unreachable.** It scanned
  `fold.lift.body` for a nested `Fold` and `_ScheduledContraction.b_trans` asked
  `isinstance(self.b, Load)` — neither has been possible since the node collapse — and
  `fold_store_tail` called `.exposes` on an adapter that has none. The adapter is deleted and the
  emitter reads the carrier off its operands.
- **`_derive_mn` read the IMMEDIATE parent site** for a nested contraction's n axis, so a cone's
  score with a zero-axis projection between it and the fold it feeds was offered no tile at all. It
  reads the nearest ENCLOSING fold now; a projection binds no coordinate.
- **`_projection_finalize`** built its lambda with the bare former and evaluated it with no cell
  bases, so a projection reading gmem at the output cell emitted the raw grid coordinates the
  tiling had already replaced. **`_strided_fold`** dereferenced `lane` before testing it.
- **`structural.form` rendered every dataclass field**, so `Axis` and `Window` growing the
  blocked-stream vocabulary moved the identity of every kernel in the tree for a pair of fields
  almost none of them set. It skips TRAILING fields left at their default now: a field a value does
  not use says nothing about what that value IS.

## Removed

`_ScheduledContraction`, `bind_widths`, `_block_widths`, the `__blk_` namespace and
`block_width_var`, `Window.trip`, the public `simplify_extent`, `BLOCK` in `PATH_FAMILIES`,
`BLOCK_STEPS`, the `BLOCK` knob spellings in `knob.py`, `is_blocked`, `block`, and
`plans/blockify-memo.md` (its reasoning is in `passes/ARCHITECTURE.md`, `ir/tile/ARCHITECTURE.md`,
`ir/schedule/ARCHITECTURE.md` and the glossary).

`git diff --stat main -- emmy/` is net positive: this adds a kernel form the compiler could not
emit, and the emitter half of it is a rewrite of a path that could not run.

## Testing

- **CI is green.** Locally `make test` is 4890 passed / 1 failed, and the one failure is an xdist
  worker crash on a `@cuda` realization case that passes in isolation and is skipped on the CI lane.
- `make test-corpus-regen`: the corpus is restamped (identity churn, above) and nine set cases gain
  the entry `--complete` fills in for a piece no row named. `rmsnorm-gqa-sdpa-stat-fill` is
  RE-KEYED rather than restamped: its row spelled a bare `TILE` / `STAGE`, and a blocked carrier has
  several sites, so the same schedule is pinned under site-keyed names.
- `make test-goldens`: red, and red on `main` too — 48/73 recorded rows on `rtx4090_sm89.yaml` equal no
  enumerated leaf at `main`'s HEAD. Not this branch's, and not re-recorded here.
- `tests/serving/goldens` re-authored through `tests.serving.regen` (the axis vocabulary moved the
  identities those rows are filed under).
- New: `tests/compiler/passes/test_flash_block.py` (the emission — both halves on the tensor cores,
  one block loop, ONE `Q·K` per block, the pivot as a fragment row reduce, the weight through the A
  slab) and a rewritten `test_block_forks.py` (which carriers block, the shared binder and weight,
  the width, and the tiles the created sites offer).

## What is left

- **The score still runs twice per block in the SCALAR tier.** On the mma tier it runs once (the
  fragments are shared); a row that takes the scalar tier inside a block pays the second pass. That
  is why the scalar tier is not offered there when a warp tile fits.
- **The cold prior does not know a blocked carrier.** The greedy reaches FA2 because the block's
  equations narrow the domain to rows that work, not because the prior prices them. A tuning round
  on this shape has not been run.
- **No golden rows are seeded for the blocked form.** Every number above is a cold or hand-pinned
  compile.

https://claude.ai/code/session_017cxTAnHwqm2D82k6Wiygm4


Slonegg:
## Performance audit (RTX 5090, cold: empty tune DB and empty online prior)

| case | eager | main | this branch | vs main | hand-tuned |
| --- | --- | --- | --- | --- | --- |
| `sdpa (1,8,512,128)` | 19 µs | 1492 µs | **177 µs** | 8.4× | 42 µs |
| `sdpa (1,32,1024,128)` | 127 µs | 21732 µs | **1603 µs** | 13.6× | — |
| `sdpa (1,4,128,32)` | 4 µs | 174 µs | **18 µs** | 9.7× | 3.3 µs |
| `matmul 512x4096x4096` | 110 µs | 1068 µs | **1074 µs** | 1.0× | — |
| `matmul 4096x4096x4096` | 745 µs | 700 µs | **700 µs** | 1.0× | — |
| `rmsnorm 512x4096` | 6 µs | 7 µs | **7 µs** | 1.0× | — |
| `softmax 512x4096` | 10 µs | 7 µs | **7 µs** | 1.0× | — |

**No regressions.** matmul, rmsnorm and softmax are within noise of `main` — the change touches
only what a blocked carrier reaches.

**SDPA is 8–14× faster than `main` cold**, and 35× on the 512-key head once tuned. Manual sweep of
36 pinned schedules there (`WORK` × the channel tile × the score tile × stage depth) bottoms out at
**42 µs** against eager's 19 µs, accuracy `max_diff=4.9e-4`. The compiler can generate an efficient
FA2 schedule; the greedy does not find it cold.

**Still behind torch on SDPA** — 2.2× on the tuned 512 head, 8–13× on the cold picks. Nothing is
tuned and no golden rows exist for the blocked form.

**`matmul 512x4096x4096` at 0.10× of eager is pre-existing** — `main` measures 1068 µs on the same
shape. Not this branch's, not addressed here.

## Where the cold picks go wrong

`plans/schedule-space-inflation.md` records it. Blocking turns one contraction site into three and
the schedule space grows 35 million-fold (6.5 × 10⁹ → 2.3 × 10¹⁷); `emmy compile --ir tile` on the
512 head takes 55 s. Two of the three sites are the SAME VALUE — the pivot's copy of `Q·K` and the
weight's — and the emitter already merges them, so one 456-choice domain is enumerated for a tile no
kernel reads.

Collapsing it was implemented and measured: the space drops to 5.0 × 10¹⁴ and the cold pick on the
512 head goes from always ~176 µs to 40 µs on half the runs (it alternates, because the greedy
benches candidates and every second run has a warm cubin cache). It is **reverted** in this PR:
`test_flash_block` stops realizing its pinned row under pytest with it in place, while the same row
compiles to four `mma.sync` through `emmy compile` and through a standalone script running the
identical passes. That is not understood, and it does not belong in this PR.

## Testing

`make test` at this HEAD (rebased on `28a9f4f48`): **2 failed, 4889 passed, 290 skipped, 49 xfailed.**

- `attention/rmsnorm-gqa-sdpa-stat-fill.yaml-realized` — the known red described above.
- `qwen3emb/gated-mlp-s128.yaml-built` — passes in isolation; an xdist ordering effect, not a
  compiler failure.

PR #730, opened 2026-09-06
Keep the elected route when one piece's row is refused, and stop offering rows the kernel binder refuses

6erun:
## Abstract

Two compiler defects found while a host evidence loop was electing a route for the DeepSeek-V4-Flash post4096 serving twin. The greedy retry keyed its blocklist on the un-lowered node's final knob row, which a later pass stamps with a policy knob no schedule leaf spells, so a refused row was never re-ranked; the retry then withdrew every splice and resolved onto the one fused kernel the measurements had already disqualified, silently. And the schedule offer enumerated rows tiling two contraction roots of a projection whose one output reads both, the one shape the kernel binder cannot bind, so that refused row was the top-ranked one. The retry now blocklists the pick it made and, once a piece has no row left, retires only the cut that minted it, logging the retirement; the offer applies the binder's ownership rule so such rows are never enumerated. On the 12-row tune-DB copy the deterministic compile elects a 14-kernel route instead of the single fused kernel.

## Election on the 12-row DB (twins golden, sm_70, no online checkpoint)

| | before | after |
| --- | --- | --- |
| kernels | 1 | 14 |
| attempts | 3, the third with every splice withdrawn; nothing logged | 1, nothing refused |
| placement | 8.5 min | 7.6 min |

Before: `k_linear_softmax_matmul_mean_reduce_9716a1` alone, 256 threads over N=4096, every fold serial per thread — the un-partitioned fused root that two `bench_fail` signatures already disqualified, which its own placement fork had priced `inf` before the retry discarded the cut it took. After (`k…` = that name; block = `__launch_bounds__`, N = the `--ir kernel` grid total):

| kernel | block | grid N |
| --- | --- | --- |
| `k…__place_c044485311__place_0853af173e` | 128 | 2,097,152 |
| `k…__place_eab46d15d0` | 128 | 8,589,934,592 |
| `k…__place_fea87a6870` | 128 | 524,288 |
| `k…__place_726953bcde` | 256 | 1,024 |
| `k…__place_335c827d1b__place_5e3efe0aaa__place_2d10103927` | 128 | 524,288 |
| `k…__place_335c827d1b__place_ce3dda12e1__place_914b1ae82d` | 128 | 2,097,152 |
| `k…__place_635d99748d` | 256 | 16,384 |
| `k…__place_c0842397d4` | 128 | 524,288 |
| `k…__place_f77140bb99` | 128 | 131,072 |
| `k…__place_335c827d1b__place_5e3efe0aaa` | 128 | 2,147,483,648 |
| `k…__place_c044485311` | 256 | 16,777,216 |
| `k…__place_335c827d1b__place_ce3dda12e1` | 512 | 66,560 |
| `k…__place_335c827d1b` (the consumer's residual) | 256 | 10,485,760 |
| `k…` (the root's residual) | 256 | 4,096 |

Against the host's iteration-6 route (12 kernels) the nine kernels outside the consumer are the same set; the consumer's two iteration-6 children (one of them disqualified by its own `bench_fail` row) are replaced by two seams with a child each, and its residual binds at 256 threads. With the both-tiled rows gone from the consumer's offer, its schedule prices differently and its placement ballot elects two seams rather than the single next arm the host loop expected.

---

## Defect 1: the retry surrendered the route

An instrumented replay printed, at every fork the decide saw on the consumer piece `matmul_1_b_unsq_bc__place_335c827d1b_0`, the identity of the picked leaf beside the node's blocklist. The blocklist entry was the terminal node's knob row and carried `LOOPIFY: 0`, a kernel-stage policy knob stamped on the node after its schedule fork; no schedule leaf spells it, so `pick in blocked?` was `False` at every fork of the retry, the piece re-picked the identical row, and the loop read the re-pick as "no row of the piece binds" and set `price_structural=False`, resolving onto the keep-fused root.

The blocklist now takes the identity of the pick the resolve made at the un-lowered node, read off the trace (the last `Decision` at that node id), which is by construction the same dict `greedy_decide` compares a leaf's row or a splice's decision knobs against. A refused row re-ranks its piece across as many retries as it has rows. Once the piece's last pick is already blocked, the strategy retires only the cut that minted it: `Decision.minted` (new; `Candidate.apply` returns the splice receipt's post-promotion ids) names the pieces a graph pick spliced in, and the minting decision's identity is blocklisted at its own node, where `greedy_decide` withdraws that one splice and re-prices the fork over the remaining arms with the same evidence. A disqualified fused side therefore keeps losing to a finite arm, and the fused root returns only when every cut above the piece has been retired in turn. The retirement is logged at WARNING with the pass and the rejection reason. Wholesale retirement is gone; `price_structural=False` now belongs to the nested price probes alone, and the prior-off validity fallback keeps structural pricing so measured arms still decide the forks they spell.

One limitation to name: the blocklist is keyed by node id and pick identity, and a node id is reused by the kernels that successively occupy it (a residual inherits its parent's id). Retiring a cut drops its pieces' blocklists, which covers the rows; a retired residual cut whose seam spells identically to one of its parent's seams would also withdraw that parent arm at the same node, a conservative over-retirement that only arises two retirements deep.

## Defect 2: the offer enumerated a row the binder refuses

The refused row, pickled at the rejection and inspected in-process: the piece is a projection over two contraction roots with one output specification, `v216 = v142 * acc47`, reading both, so `projection_regions` refuses ("an output-tiled root must own each output specification independently"). The binder takes that multi-root path only when the row tiles more than one root, and the row did (`TILE@map.1/map.1/inner: f26`, `TILE@map.2/inner: f2`). The classic offer projects every root's tile domain independently and the schedule context composed only worker, physical-axis and fragment agreements, so the row was enumerated, ranked first, and refused at materialize.

The rule is a term fact and a relation between root sites, so it composes in `ClassicScheduleContext.extend` beside those agreements: `_shared_roots` is the set of `kernel_roots` when `projection_regions` refuses for the term, and `_support_refusal` refuses a tiled pick on one of them while another is already tiled in the prefix. A pinned row tiling both is refused at decode through the same path. Where the projection partitions its outputs by root nothing changes.

The corpus already recorded this shape as a known gap, `fused/gate-up-distinct-a_xfail_realized.yaml` (offered, then refused at lowering). The pinned row is no longer enumerated, so the case moves to `_xfail_offered` with its evidence comment kept; its recorded gap (the tier the shared-A twin realizes) is unchanged, only the stage at which it shows. No case regenerated. `tests/compiler/ir/schedule/test_classic.py::test_independent_nodes_compose_only_at_matching_physical_axis_geometry` built a two-root projection with no stores at all, a shape the binder refuses too; it now carries one store per root, the partitioning form, and keeps its assertion.

## Tests

`tests/compiler/pipeline/test_lowering_error_guardrail.py`: the composed-route fixture stamps a policy knob on every node still un-lowered after materialize (the production shape). `test_refused_piece_rows_rerank_within_the_composed_route` (first two rows refused, third binds; red before, electing the fused root), `test_exhausted_piece_retires_only_its_own_cut` (the residual cut again, every row of the inner piece refused: the inner cut is retired, the root's cut kept, a WARNING carries the reason; red before), and the all-rows-refused fallback kept. `tests/compiler/passes/test_classic_schedule_domains.py`: a two-root projection whose one output multiplies both accumulators enumerates no row tiling both roots while each root still reaches the tile tier alone (red before); its partitioning twin still offers both tiled.

## Gates

- `make test`: 4190 passed, 1036 skipped, 21 xfailed. No corpus case came back stale. The renamed corpus case keeps its measured durations under its new id (its `realized` entry dropped, that stage now being skipped).
- `make lint`: clean.
- `make test-goldens`: 11 files fail on the branch, and the same 11 files fail with identical failing-row counts on a pristine detached `origin/main` (`b285dbfe`) in the same worktree — Laguna-S-2.1-exl3/rtx5090 8/8, rtx4080 9/9, rtxpro6000 10/10, OLMoE/rtx5090 10/13, Qwen3.5/v100 24/44, rtx4090 48/73, gemma-4-12B/rtx5090 186/194, rtx5090 38/63, gemma-4-12B-it/rtx4090 153/240, DeepSeek-V4-Flash/v100 267/279, gemma-4-12B-it/rtx5090 226/352. Pre-existing; this change moves no row.
- Core line balance: `git diff --stat origin/main -- emmy/` is +178 −111 over nine files, +152 −103 (net +49) over the five Python modules. This is growth for capability the old design could not express: a compatibility relation the offer did not carry (+25 in the schedule context), and a retirement that withdraws one splice instead of all of them, which needs the trace to know what a splice minted (+7) and the decide to withdraw a blocklisted splice (+6). The strategy's net +18 is the `_retire` / `_block` pair replacing the wholesale-retirement branch; its two readers of "which rejected nodes are still un-lowered" fold into one.


Slonegg (approved):

6erun:
Merged `main` (#731) — clean; the two shared files auto-merged. On the merged head: lint clean, `make test` 4193 passed, `make test-goldens` 11 failed with per-file counts identical to the pristine `origin/main` baseline (pre-existing; neither #731 nor this branch moves a row).

PR #734, opened 2026-09-07
Blame the kernel nvcc names when a bench's compile fails, quote escaped or not

6erun:
## Abstract

A bench that fails because nvcc refuses one kernel's source names that kernel, but the failure was recorded against nobody, so the next compile elected the same route and failed the same way — the evidence loop for the DeepSeek-V4-Flash post4096 twin stopped on exactly this after the parallel tune had filled its DB. The blamed kernel is read off the failure text because the exception class does not survive the bench worker's pipe; the worker wraps the child's exception through `repr`, and an nvcc message quotes identifiers with `"`, so `repr` escapes the kernel name's own quote and the blame pattern missed it. A hang's message holds only single quotes, which is why every hang was blamed and this was not. The pattern now accepts the escaped quote, so a compiler-refused kernel earns its `bench_fail` row like a hung one and the election moves on.

## What it took to see

`run --bench --record-greedy` on the 623-`ok`-row DB elected a 15-kernel route whose second kernel (`…__place_0fee4f4677`, a Volta `mma` tile over `a26_1 × a34`) does not compile — its B-operand address uses the unsplit axis `a26_1` while the kernel's decode only defines `a26_1_b` / `a26_1_u` (a codegen defect, reported separately, not touched here). The log said `pinning bench_fail … for 0 of 2 kernel(s)` shape-wise: nothing filed, `perf` unchanged, and the concurrent CPU dump proved the election deterministic. Reproduced GPU-free: `_NAMED_KERNEL.search("bench worker error: " + repr(RuntimeError("nvcc compile failed for kernel 'k_x': … \"_lane\" …")))` is `None`; with the escaped quote allowed it is `k_x`, and the hung-kernel text still matches.

## Tests

`tests/compiler/pipeline/search/policy/test_terminal_bench.py::test_a_kernel_nvcc_refuses_is_blamed_alone`: a two-kernel terminal whose worker-wrapped nvcc failure names the culprit; red before (`0 of 2 kernel(s)` blamed), green after, the innocent kernel rowless either way. The hung-kernel and unattributable-failure cases beside it are unchanged.

## Gates

- `make test`: 4194 passed, 1039 skipped, 21 xfailed.
- `make lint`: clean.
- `make test-goldens`: the change is a pattern in the failure writer, outside the schedule codec; the gate's state is reported in the PR thread once the run completes (it is red on `main` today for all files).
- Core line balance: `emmy/` +7 −4 — the pattern's comment grew by the reason the escaped quote exists; no new mechanism.

### DIFF
--- a/emmy/commands/ARCHITECTURE.md
+++ b/emmy/commands/ARCHITECTURE.md
@@ -267,9 +267,10 @@ the deploy evidence the next `compile` / `run` / `serve` picks from, which is ho
 embedded golden lowers in-process, knobs and all, so it records like a traced model; only the `--ir` JSON path, whose
 serialization drops the knobs, stays unrecorded. A greedy row that fails to bench is recorded the same way the
 tuner records a hung terminal (`bench_record.record_bench_failure`, the tuner's `persist_bench_failure`): the kernel
-the watchdog named — or a one-kernel graph's only kernel — earns a `bench_fail` perf row at the run budget's fail
-sentinel and the innocent kernels earn none, so the next compile disqualifies that arm instead of electing the same
-route and hanging again; a failure that names no kernel records nothing, and a compile-budget overrun measured
+the failure names — the one the watchdog saw hang, or the one nvcc refused to compile — or a one-kernel graph's only
+kernel earns a `bench_fail` perf row at the run budget's fail sentinel and the innocent kernels earn none, so the
+next compile disqualifies that arm instead of electing the same route and failing the same way again; a failure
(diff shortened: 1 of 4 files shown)


6erun:
`make test-goldens` on this branch: 11 files fail with per-file failing-row counts 8/8, 9/9, 10/10, 10/13, 24/44, 48/73, 186/194, 38/63, 153/240, 267/279, 226/352 — the same eleven counts #730 measured on pristine `main`. Pre-existing; this change moves no row (it touches only the failure writer's blame pattern, outside the schedule codec).

PR #739, opened 2026-09-07
Offer the composed arm a measured route row names, so a recorded composed cut replays and deploys under strict evidence

6erun:
## Abstract

A pinned compile takes every placement pin that resolves on one kernel as a single composed decision, and recording the greedy pick writes that decision as one measured routing row naming every seam. Reading such a row back — at deploy, and in the replay that keys a golden's rows — took only the first seam the row marks, because the placement fork offers single seams and nothing else, so the recorded kernels were never minted again, every per-kernel receipt matched nothing, and a strict-evidence compile of the file refused the very route it had just measured. The placement fork now offers, beside its single seams, the composed arm a measured route row of the kernel names, and a row that marks several offered seams spells that arm.

## Mechanism

`030_cut._placement_restriction` consumes the scoped `PLACE@seam=cut` pins that resolve on a kernel as one realization with `placement_decided=True`; `KernelSetDecisions` records that as one decision whose knobs carry every seam, and `record_greedy_pick` writes it as one routing row. The unpinned fork, however, is `[fuse] + one DeferredFork per seam`, and `pins.spelled_arm` returned the first offered option marking any seam the row marks `cut`. So `golden._replay` of the routing row took one seam, left the rest `unresolved`, resolved a program whose kernels matched none of the receipts (`holders` empty on every one), and at deploy `_route_candidates` handed the root's fork that one seam — after which the residual had no measured row and strict evidence raised `EvidenceError`. Reproduced on the DeepSeek-V4-Flash `post4096` serving twin on the V100 host: the route benched completely under its pins, `--record-greedy` wrote 1 routing row (11 seams), 2 split rows and 12 receipts, and `compile --golden … --strict-evidence` failed at the root's `030_cut` fork with "no measured row spells a kernel-set arm".

## Fix

`pins.composed_routes` is a scoped registry of measured composed routes — `(kernel signature | None, cut keys)` — and `composed_cuts_for` answers it with the same subset-and-equal rule the evidence index matches rows by. `030_cut._placement_forks` offers one composed arm per registered route whose keys resolve to at least two seams of the kernel, realized with `placement_decided=True` like the pinned cut that measured it. `GreedyStrategy.run` registers the index's multi-seam route rows (keyed by the recorded kernel's signature, less the schedule-fork stamps) around its resolves; `golden._replay` registers the cut sets of the record, the set's lead and every identity-named sibling around its own resolve, since any of them may decide a fork. `spelled_arm` prefers the option whose cut set equals the offered seams the row marks when it marks more than one, and reads a single-seam row exactly as before. The passes stay evidence-blind: the pass reads a registry the way it reads the environment, and a route row never becomes a pin.

## Tests

`tests/compiler/pipeline/search/test_pins.py::test_a_row_marking_several_offered_seams_spells_the_composed_arm` (red: the first single seam was returned) and `tests/compiler/cli/test_golden_file_replay.py::test_recorded_composed_pick_is_picked_again_under_strict_evidence` — a composed pinned pick recorded with `record_greedy_pick`, then re-picked under `sole_evidence` and strict evidence with the same kernels and rows (red: `EvidenceError` at the first piece, the host's failure shape). Both watched failing against pristine `emmy/` for those reasons before the change. The realization corpus case `attention/rmsnorm-qk-sdpa-composed-cut` recorded exactly this gap as `_xfail_realized` ("as evidence the route is read fork by fork … the case needs one receipt per piece"); it realizes now and is renamed to drop the suffix per the corpus rules.

## Gates

`make test` (Mac, correctness lane): 4219 passed, 1045 skipped, 19 xfailed, plus the corpus case that moved from xfail to pass. `make lint`: clean. `make test-goldens`: red in the same 11 files as pristine `origin/main` with identical per-file counts (listed in the last comment). `tests/durations.json` gained the new test's entry at its measured 7.5 s.

## Line balance

`git diff --stat origin/main -- emmy/`: +143/−33 across four modules. The growth is the capability — a measured composed decision is now an arm the deploy and the replay can take — and not a special case layered onto the evidence pick: the registry is the one seam between the strategy, the replay and the pass, and `spelled_arm` keeps a single reading for both consumers.

The registry fills only for a pipeline that lowers to CUDA: a golden record's derivations lower its target through the loop passes — 623 resolves for the DeepSeek working file — and none reaches the cut pass, so the first form of this branch had each of them build the evidence index (about two minutes apiece on the host with a tune DB and a card golden in scope); guarded, the same strict compile takes 46 s.

On the host, with these four modules synced, `compile --golden working.yaml --realization post4096.… --strict-evidence --target sm_70` follows the recorded route from the file alone (plus the tune DB copy's per-kernel rows): the emitted program is byte-identical to the one the route bench measured at 2.10 s per forward, 12 kernels, no pin in the environment.

## Plan

The DeepSeek V4 serving plan's Stage 0 item 4 ended on the host measurement of a route still being owed. It now records the measured 2.10 s route and how it was recorded, the three findings that redirect the item (the prior's pricing of unmeasured arms rules out resuming the loop or the tune as they stand; the mixed stream's RMS statistic is the one recompute behind every hang; two naive kernels hold 93 % of the time), and that the route has no correctness verdict yet; the consequence paragraph and the effort estimate stop quoting 23.2 s.


6erun:
`make test-goldens` on this branch: red in the same 11 files as pristine `origin/main` (`c59212fa`), with identical per-file counts of recorded rows that equal no enumerated leaf — Laguna-S-2.1-exl3 rtx5090 8/8, rtx4080 9/9, rtxpro6000 10/10, OLMoE-1B-7B rtx5090 10/13, Qwen3.5-122B-A10B v100 24/44, rtx4090 48/73, gemma-4-12B rtx5090 186/194, rtx5090 38/63, DeepSeek-V4-Flash-0731 v100 267/279, gemma-4-12B-it rtx4090 153/240, gemma-4-12B-it rtx5090 226/352 — so no row is attributable to this change.

Slonegg (approved):

PR #791, opened 2026-09-11
Source the classic schedule enumeration from the row

Slonegg:
## Abstract

The classic schedule enumeration is now sourced from the knob row instead of filtered by it. A schedule problem is factored into sites, one per node and the kernel site last, and a site the row names offers the row's value alone, parsed and checked with the same rules its catalog passes through; the context keeps only the compatibility join, and the restriction subsystem that filtered projected domains after the fact is gone. The pipeline offers the schedule fork as one unexpanded root, which a row-directed descent narrows to the row before expanding, so replaying a recorded row is one path. On the realization corpus's heaviest case the pinned enumeration drops from 84 s to 32 s and the evidence replay from 106 s to 2 s.

| corpus node (RTX 5090 dev box, CPU only) | before | after |
| --- | --- | --- |
| `nvfp4-gate-up-requant-full-projection-cut` offered | 84 s | 32 s |
| `nvfp4-gate-up-requant-full-projection-cut` realized | 106 s | 2 s |
| every case, `offered` + `realized`, 8 workers | — | 44 s |

---

## The interface

`ir/schedule/base.py` names the two terms of an enumeration. A `ScheduleProblem` is the tile, the target and the row, factored into `Site`s; `Site.options` is the source of every candidate. A `ScheduleContext` is the prefix and owns compatibility only. `narrowed(row)` is the one way a row enters after construction: an empty prefix over the same problem with the row installed.

`ClassicProblem`, `ClassicNodeSite` and `ClassicKernelSite` implement it. A node site's `nodes` are the row's parsed value where the row names the site's codec key, else the catalog; `edges` likewise for the transport; the kernel site derives its inventories from what the node sites offer, so it is last. The per-choice refusals the projection already had (atom, chunk, plan, k-step, wgmma, catalog grid and budgets, reduction and transport membership) are applied to a parsed value exactly as to a catalog entry, so a row selects a value the catalog would have offered and cannot manufacture one.

There is no product object any more: `ClassicDomains` held a copy of what the sites offer, and once the sites became the source nothing in a compile read it, so it folded into the test helpers along with `project_classic` and the `domains_literal` injection. A bounded test hands a site hand-written factors by subclassing it (`tests/compiler/helpers.literal_classic_context`), which is the only way to offer a site values it did not project.

## What the context lost

`restrict`, the pin tables, the per-site restriction predicates, the bare-pin readings and the precision gates left `ClassicScheduleContext`. The bare-pin disjunction (one site of an ambiguous family carries the value, the others are OFF) and the precision policy (f16-accumulate and FP8 atoms offered unpinned only where allowed; an authored tile bypasses it) moved to the problem, where they shape what a site offers. The transposed rasters are no longer the catalog's own offer; a row naming one takes it.

The pinned enumeration's cost was the restriction itself: checking each catalog choice re-spelled the whole domain through `values(key)`, quadratic in the domain. The evidence replay's cost was the schedule fork's root level, expanded eagerly at offer time for every kernel before any descent could narrow it.

## Layout

`ir/schedule/classic/` replaces `classic.py` and `classic_projection.py`, one role per module: `assignment` (the choice types, the sites' wire spellings and keys, `ClassicDomains`), `refusals` (every per-choice legality rule, asked of a catalog value and a parsed row value alike), `sites` (the source), `context` (the join), `codec` (the wire boundary) and `materialize` (the lowering boundary). Imports flow in that order. The five reads of `ir/tile/ops` sit inside the functions that need them, so nothing in the package imports the tile package at module level, which is what lets `ir/tile/ops` keep reading the assignment names through the package without a cycle. `domain.py` folds into `catalog.py`.

## The pipeline side

`classic_forks` folds the environment's pins into one row (`pin_row`) and offers one lazy root. `leaf_for` narrows a root to the row it descends with, and `_find_decided_leaf` is that descent held to an exact match. The pool bounds the greedy reads before walking come off the problem lazily.

## Behavior changes

- A hand-pinned transport no support resolves, a warp-group tile its grid cannot feed, and a transport the card cannot run still raise with the rule's message under a hand pin (`loud_pins`); a row a descent follows is answered by an empty site and the caller re-decides.
- Under `validate_pins=False` an unrealized bare pin on an ambiguous family is refused only when some site could have carried it; before, it refused every completion even when no site could.
- `project_classic` no longer lists `gn4` / `gn8` in an unpinned raster domain; two domain tests' expectations changed accordingly.
- The precision-policy test now states the policy as a catalog property instead of an `extend` refusal over literal domains.
- Two tests promised that a pin changes no factor, which is the opposite of what this branch does; they now assert that a row selects among the values a site offers and adds none.

## Line balance

Python under `emmy/` grows by 178 lines net (2379 added, 2201 removed, most of it the module split moving lines). The removal is the context's restriction subsystem; the growth is the site interface itself, the row parsing that replaces filtering, the lazy root, and six module headers where there were two. The change adds a capability the old design could not express: an enumeration that offers one value where a row names a site.

## Verification

`make test` on the RTX 5090 dev box: 5378 passed, 15 failed, and the 15 are exactly this box's known set on `main` (the 12 serving GPU tests missing cut-fork evidence, the `rms-norm-cut-sweep-work` correctness node, the block-scaled cell tolerance test, and one xdist worker crash on the packed-slab build, which passes alone). `make test-goldens`: the same three files red as on `main` (Laguna EXL3 on the 5090, gemma-4-12B-it on the 4090 and 5090), four green; no row went red here. CI's durations gate named a serving test at the 5 s bar; it and its neighbour at 4.6 s are recorded.

## Not done here

The `offered` oracle still walks the unpinned pieces of a cut target in full, which is the remaining 32 s on the heaviest case. The per-site `options` tuple builds one `Schedule` per node × transport pair eagerly; about a fifth of that walk is those constructions.

https://claude.ai/code/session_018i6CpBdUZmbeWw954KcGYb


gnull (commented):
This seems to ignore a knobs pin passed through an environment variable in favor of goldens. Reproduction example from AI:

```sh
EMMY_TILE=mma_m16n8k64_e2m1_f32/f1x2/k8 emmy compile --golden tests/compiler/realization/cases/matmul/  nvfp4-w4a4-packed-slab-loopify.yaml --realization k_linear_936f0e.0cbc6cafc0d2 --target sm_120 --ir tile --no-readable
```

It does not use what you pass in `EMMY_TILE` and prefers what it finds in the golden.

gnull (approved):
Nevermind my earlier comment, it wasn't actually true.

PR #795, opened 2026-09-12
Record a working M=1 decode schedule, and stop deploying the tier when it is slower

6erun:
## Abstract

DeepSeek V4 Flash could not answer a single generation request on 16x V100: every one died on vLLM's 300 s
`sample_tokens` deadline. The cause was one kernel. Its M=1 schedule in the golden was the only measured candidate
for that shape and it takes 29.7 seconds, so strict evidence had no alternative to elect, and single-token decode ran
at roughly 61 seconds per layer. This records a schedule that runs it in 419 us instead, stops the runner deploying
the M=1 tier when it is not actually faster than the twins it replaces, and fixes the boot audit that should have
reported the mispick and did not. Time per output token goes from 5.565 s to 0.899 s.

---

## The kernel

`pre1.k_linear_mean_reduce_7fce9f` at M=1 had one measured candidate, at 29,693,246 us. Both of its candidates
bench-fail outright:

```
HungKernelError: kernel 'k_linear_mean_reduce_7fce9f (iter 1)' did not complete within 2000 ms
```

Its Loop IR shows why, and it is the same defect already recorded for `pre4096` in the experiment's RESULTS.md — the
same `k_linear_mean_reduce` kernel at a different width. A 16,384-long dot product that depends only on the stream
index sits inside a 4,096-wide output-channel sweep, so it is recomputed 4,096 times, and the structure appears twice
in the program:

```
for a1 in 0..4096:              # output channel
    for a2 in 0..4:             # stream
        for a3 in 0..16384:     # depends on a2 only, recomputed for every a1
            acc1 <- add(acc1, v8)
```

The same four placement cuts that fixed `pre4096` fix this shape. Only the m4096 shape had ever been recorded.

## What changed

**The golden.** `emmy run --golden ... --realization ... --bench --record-greedy` under those cuts adds five
realizations: the four kernels the cut program splits into and the routing row that composes them, priced at 411 us
against the old 29.7 s row for the same identity. Nothing else in the file changes and no realization is removed.

**The runner.** The M=1 tier is an optimization — the bucket twins already cover `T=1` by padding up, and the M=1
twins exist only because the contractions demote to faster planar forms at one row. Nothing checked that they were
faster. Here they were 500x slower. The boot audit already times both tiers under the GPU lock, so it now returns its
measurements and the runner drops the tier when the M=1 twins do not beat the bucket twins.

**The audit.** It exempted any program whose roofline floor sat under `MIN_FLOOR_US`, which is why this went unseen
for three rounds — this program's floor is under it. The reasoning in the docstring was that "a mispick there costs
little in absolute terms", and that is the error: a small floor bounds what a healthy program costs, never what a
mispicked one does. Programs with no usable floor are now judged on absolute cost, the measurement is bounded so the
audit does not pay four times a mispick's cost to learn what one run showed, and every warning carries the measured
time.

## Measured

Serving boot on the 16x V100 host, `--strict-evidence`, TP8 x PP2, M=1 tier enabled, single stream, greedy:

| | before | after |
| --- | ---: | ---: |
| Time per output token | 5.565 s | **0.899 s** |
| Output tok/s | 0.180 | **1.11** |
| `pre1` @ M=1, whole program | bench_fail (>60 s) | **419 us** |

Thirty-one steady decode steps span 1.3% (0.894 - 0.906 s). The first decode step of a request costs 1.85 s: each
layer's M=1 programs are CUDA-graph captured on first use. Completions are coherent English, which is the correctness
check that matters here — there is no eager twin for a golden Loop IR target on this model, so no CLI-level numerical
verdict was obtainable for the cut schedule alone. The cuts are code motion, hoisting a loop-invariant dot out of a
sweep, which is why the equivalent `pre4096` cuts were bit-exact.

Against the pinned 1Cat fork on the same host, the fork's lead per output token falls from 38x to 6.1x. Time to first
token is unchanged at 23x: prefill runs the m4096 twins, which this does not touch.

## What this does not do

Two programs now dominate and neither has been recorded: `post.decode.m1` at 293x its roofline floor is 17 ms per
layer and most of the remaining 0.9 s, and `post.chunk.m4096` at 317x is 619 ms per layer and most of the time to
first token.

The tier guard reuses the boot audit's measurements rather than timing separately, so it inherits the audit's
sampling: one layer per attention class, not every layer. A model whose M=1 tier is fast in the audited layer and slow
elsewhere would not be caught by it, though the absolute-cost rule now reports the slow program either way.

## Line balance

`git diff --stat main -- emmy/` is positive. The growth is the two guards described above, both new behaviour rather
than layering: the audit gained an absolute-cost rule for programs it previously skipped, and the runner gained a
tier choice it previously did not make.

## Gates

`make test` passes: 4667 passed, 1017 skipped, 12 xfailed. `make lint` clean.

`make test-goldens` fails on `DeepSeek-V4-Flash-0731/v100_sm70.yaml`, and it fails identically on pristine `main`:

```
pre1.k_linear_mean_reduce_7fce9f.7c4629931db7.m1.0733f2fad083:
  no enumerated row equals the recording (551 candidate rows)
```

That is the stale 29.7 s row, which `emmy run` independently reports as `unreproducible pin:
TILE@map.1/map.1/reduce.1/inner=f1x10 realized TILE=f1`. It is not touched here and not re-recorded to go green —
the five rows this adds are new names alongside it. Deleting it is worth doing separately, on its own argument.
`Laguna-S-2.1-exl3/rtx5090_sm120.yaml` and both `gemma-4-12B-it` goldens also fail on main and are untouched by this.

### DIFF
--- a/emmy/serving/gen_runner.py
+++ b/emmy/serving/gen_runner.py
@@ -1749,7 +1749,27 @@ class EmmyGenRunner:
             for key, tag, m in (("bucket", f"bucket.m{decode_bucket}", decode_bucket), ("one", "one.m1", 1), ("m256", "m256", 256))
             if tiers[key] is not None
         ]
-        audit_boot_programs(named, dtype_bytes=np_dtype.itemsize)
+        measured = audit_boot_programs(named, dtype_bytes=np_dtype.itemsize)
+        # The M=1 tier is an OPTIMIZATION: the bucket twins already cover T=1 by padding up, and
+        # this tier exists only because the contractions demote to faster planar forms at one row.
+        # When its programs do not measure faster, keeping it is strictly worse. On 2026-09-12 it
+        # was 485x worse — a DeepSeek-V4 V100 boot elected an M=1 pre program at 29.7 s per forward
+        # and every request died on the engine's RPC deadline, with the roofline audit blind to it
+        # because that program's floor was too small to form a ratio. The audit already timed both
+        # tiers under the GPU lock, so this reads its numbers rather than measuring again.
+        if runner._pre_m1 is not None:
+            m1_us = sum(us for label, us in measured.items() if label.endswith(".decode.m1"))
+            bucket_us = sum(us for label, us in measured.items() if label.endswith(f".decode.m{decode_bucket}"))
+            if m1_us and bucket_us and m1_us >= bucket_us:
+                logger.warning(
+                    "[gen_runner] the static M=1 decode twins measure %.3f ms against %.3f ms for the bucket-%d "
+                    "twins they replace — dropping the M=1 tier; T=1 decode rides the bucket twins. Tune the M=1 "
+                    "twins (`emmy tune`; see emmy/serving/ARCHITECTURE.md → 'Tuning what serving actually runs').",
+                    m1_us / 1e3,
+                    bucket_us / 1e3,
+                    decode_bucket,
+                )
+                runner._pre_m1 = runner._post_m1 = None
         return runner
 
     def embed(self, input_ids):
--- a/emmy/serving/roofline.py
+++ b/emmy/serving/roofline.py
@@ -163,23 +163,27 @@ def flag_ratio(
     return floor_us, ratio
 
 
-def audit_boot_programs(named_programs, dtype_bytes: int = 2) -> None:
+def audit_boot_programs(named_programs, dtype_bytes: int = 2) -> dict[str, float]:
     """Time each ``(label, _Program, m_tokens)`` against ``max(weight-streaming, compute)`` floor and
     warn on outliers. ``_Program`` here is the serving wrapper (``gen_runner._Program``): ``.program``
     is the ``CompiledProgram`` and ``.weight_bytes`` the per-forward weight footprint — bound constants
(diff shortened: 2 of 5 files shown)


6erun:
Receipts for the `test-goldens` claim in the body, so the red gate does not have to be taken on trust. Same filter (`-k DeepSeek`), same selection count, run twice — once with this branch's golden, once with `origin/main`'s restored in place:

**This branch**

```
pre1.k_linear_mean_reduce_7fce9f.7c4629931db7.m1.0733f2fad083:
  no enumerated row equals the recording (551 candidate rows)
1 failed, 296 deselected in 1067.87s
```

**Pristine `origin/main`**

```
pre1.k_linear_mean_reduce_7fce9f.7c4629931db7.m1.0733f2fad083:
  no enumerated row equals the recording (551 candidate rows)
1 failed, 296 deselected in 3.43s
```

Identical row, identical message, identical candidate count. The 83 lines this adds introduce no new decode failure.

The wall-clock difference is contention on my machine, not the test — the slow run shared cores with other pytest processes.

Slonegg (approved):

PR #802, opened 2026-09-14
Release emmy-ml 0.3.4: ship the Qwen3.8-27B-FP8 recipe

Daksh-10:
## Problem

We want to serve `Qwen3.8-27B-FP8` on CloudRift inference. The relay reads its model
catalog from the installed `emmy-ml` wheel (`emmy recipe list --json`), pinned at
`EMMY_VERSION=0.3.3` in `rift-relay/Dockerfile`. The recipe exists on `main` (#733,
merged 2026-09-07) but not in `v0.3.3`, so `GET /v1/admin/recipes` cannot return it on
any environment and `POST /v1/admin/offerings` 404s.

`main` is 251 commits / 300 files ahead of `v0.3.3` and is mid-refactor, so tagging it
to ship one recipe is not an option. This branch is cut from the `v0.3.3` tag and carries
only what the release needs.

## What's here

| Commit | Change |
| --- | --- |
| `137d7e26` | `recipes/Qwen3.8-27B-FP8/{recipe.yaml,RESULTS.md}`, cherry-picked from #733. Data only, no code from `main`. |
| `3ce27c90` | Cherry-pick of #568 — adds `NVIDIA Tesla V100 SXM2 16GB` to `GPU_INSTANCE_TYPES`. |
| `7e79c534` | `pyproject.toml` version → `0.3.4`. |

**#568 is required, not incidental.** The recipe's only qualified hardware is
4x `NVIDIA Tesla V100 SXM2 16GB`, and `v0.3.3`'s hardware table has no entry for that GPU
— only SXM3 32GB. Without it the recipe loads and publishes fine but the deployment dies
at provisioning:

```
ValueError: Unknown GPU 'NVIDIA Tesla V100 SXM2 16GB' — not in hardware table
```

With it, `iter_candidates(...)` resolves to `v100-6-52-400-generic.4`, which the live
CloudRift catalog stocks (provider Cato, `usny01_a01` / `ustx1a_a01`, 4 nodes, 3 free).
#568 also brings its own regression test asserting every GPU a bundled recipe declares is
in the table.

## Compatibility with the v0.3.3 release line

The recipe uses fields added after `v0.3.3` was cut, so I checked each against that code:

| Concern | Result |
| --- | --- |
| `tags: [best-effort]` | Valid lifecycle tag; `DISABLED_TAGS` is `{obsolete, onboarding}`, so `recipe_is_runnable` is `True` and relay classifies it `RecipeKind::Llm` |
| `model.rationale` / `revision` | Present on `ModelConfig` |
| `model.heat` (main-only) | `Recipe.from_dict` reads model keys explicitly, so unknown keys are ignored, not an error |
| `matrices:` as a mapping | Same form `recipes/DeepSeek-V4-Flash-0731` already uses at `v0.3.3` |
| `extra_args` | `validate_extra_args(..., engine="vllm")` passes |
| vLLM image | `cloudriftai/1cat-vllm-deepseek-v4-flash-0731:1.2.3-d76126608` is already referenced by a `v0.3.3` recipe, so it is published |
| Wheel bundling | `prepare_dist.py --recipes` stages 27 recipes (was 26) — it filters on `recipe_is_runnable`, which `best-effort` passes |

## Verification

```
tests/test_hardware.py tests/recipe/       121 passed
tests/provisioning/ tests/commands/        191 passed
ruff check                                 All checks passed
ruff format --check                        627 files already formatted
scripts/prepare_dist.py --recipes          staged 27 recipes
```

Catalog document `emmy recipe list --json` produces, which is the relay's contract
(`RECIPE_CATALOG_SCHEMA_VERSION = 1`):

```json
{
  "name": "Qwen3.8-27B-FP8",
  "model_id": "Qwen/Qwen3.8-27B-FP8",
  "tags": ["best-effort"],
  "task": "generate",
  "runnable": true,
  "deployments": [
    {"gpu": "NVIDIA Tesla V100 SXM2 16GB", "gpu_count": 4, "context_length": 262144}
  ]
}
```

## Notes for review

- **Base is `main` only so the diff renders.** The merge base is `v0.3.3`, so Files Changed
  shows just these six files. The branch is meant to be **released from**, not merged —
  merging would move `main` to `0.3.4` (it currently reads `0.3.3`) and no-op the two
  recipe files, which already exist there identically. Say the word if you'd rather this
  land a different way.
- `best-effort`, and the recipe's own header notes RTX PRO 6000 Max-Q and H200 are
  *unqualified* rather than unsuitable — only the Volta path was measured.
- The offering will be **text-only**: `--language-model-only` drops the vision tower.
- Release path once approved: `workflow_dispatch` on `publish.yml` against this branch —
  it uploads to PyPI first, then mints `v0.3.4` and the GitHub Release, so a failed upload
  leaves nothing behind.


Daksh-10:
Don't approve this PR in github as we don't need to merge with main. If it looks fine, then I'll deploy it using `workflow_dispatch`.

Slonegg (approved):

PR #806, opened 2026-09-14
Plan reusable Rust runtime and native LLM serving

6erun:
## Abstract

Plan a reusable Rust runtime for Emmy's compiled programs, shared by benchmarking and a later native API server. Start with measured baselines and an independently usable executor; keep vLLM as the serving default.

---

## Migration

Measure dispatch, padding, and memory costs first. Extend the existing execution-plan export and qualify Rust execution for benchmarking and cached Qwen3 generation. Evaluate reusable OpenAI-compatible frontends before implementing the API server.

## Review decisions

Keep emmy-runtime independent of emmy-server and remove the proposed Python generation bridge. Replace Python dispatch only after parity, timing, and fault recovery are qualified. Existing CUDA graphs and persistent workers remain the performance baseline.

## Validation

Documentation only; whitespace checks pass. Runtime tests and GPU benchmarks were not run. Three experimental reports are planned; no Rust speedup has been measured.


plans/native-llm-serving.md
```diff
@@ -0,0 +1,351 @@
...
+the inference runtime.
+
+The first milestone serves dense Qwen3 models in FP16 on one GPU, with cached generation and a streaming API.
+It runs without vLLM installed. Success means correct behavior and ownership of execution; outperforming vLLM
+comes later.
+
+## Approach
+
+- Reuse Emmy's compiled model runner and device execution paths.
+- Use Rust with Axum for HTTP/SSE, Tokio for asynchronous I/O, and Serde for request and response serialization.
```
Slonegg: Is there an existing library for an OpenAI-style API that we can use? What functionality would we need to implement ourselves if we don't use a library?

plans/native-llm-serving.md
```diff
@@ -0,0 +1,351 @@
...
+
+The strongest opportunity is designing compilation and serving together. Rust can lower CPU overhead, but changing
+the server language alone does not establish an inference speedup. These are hypotheses to measure:
+
+- **Unified GPU memory planning:** budget weights, activation/scratch buffers, and KV cache together. Reuse storage
+  and size buffers to scheduled work so more memory remains for concurrent requests. Existing serving evidence
+  identifies non-KV footprint as a concurrency limit; replacing an allocator alone does not shrink that footprint.
+- **Scheduling matched to compiled kernels:** choose efficient batch sizes and prefill chunks, reducing padding,
+  unnecessary shape changes, and gaps between useful GPU work.
+- **Lower CPU dispatch overhead:** remove Python from step preparation and submission. This matters most for short
```
Slonegg: We're using CUDA graphs. Where is the overhead coming from? What are the optimization opportunities, and how much can we save? Create separate MD files with supporting experimental data.

plans/native-llm-serving.md
```diff
@@ -0,0 +1,351 @@
...
+
+## Potential gains
+
+The strongest opportunity is designing compilation and serving together. Rust can lower CPU overhead, but changing
+the server language alone does not establish an inference speedup. These are hypotheses to measure:
+
+- **Unified GPU memory planning:** budget weights, activation/scratch buffers, and KV cache together. Reuse storage
+  and size buffers to scheduled work so more memory remains for concurrent requests. Existing serving evidence
+  identifies non-KV footprint as a concurrency limit; replacing an allocator alone does not shrink that footprint.
+- **Scheduling matched to compiled kernels:** choose efficient batch sizes and prefill chunks, reducing padding,
```
Slonegg: Compare it with the current serving path. Give an example of inefficiency.

plans/native-llm-serving.md
```diff
@@ -0,0 +1,351 @@
...
+thread blocks. Tokio schedules CPU I/O tasks, not GPU kernels. There is no automatic CPU/GPU placement policy:
+text/control work is CPU work, and model computation is GPU work.
+
+### Milestone 2: Rust inference runtime
+
+Replace the Python worker with a dedicated Rust execution thread in the server process. Keep blocking GPU waits
+and tokenization off Tokio's I/O threads. Use bounded channels between HTTP tasks, CPU text processing, and GPU
+execution. One execution thread owns the CUDA context, allocations, streams, and request cache state for the GPU.
+
+Extend Emmy's existing execution-plan and compiled-binary export into a versioned, self-contained serving artifact.
```
Slonegg: This can be a standalone contribution: define the runtime format. Create a basic Rust execution engine to execute the program.

Would it make sense to drop our current Python CUDA dispatch engine? Can we fully rely on the Rust runtime? Can it be persistent, so we avoid restarting the process/thread and reloading CUDA binaries? Will it speed up the current benchmarking pipeline or slow it down?

plans/native-llm-serving.md
```diff
@@ -0,0 +1,351 @@
...
+[cudarc](https://docs.rs/cudarc/latest/cudarc/), [Tokenizers](https://docs.rs/tokenizers/latest/tokenizers/),
+and [MiniJinja](https://docs.rs/crate/minijinja/latest).
+
+## Repository structure
+
+Keep the server in this repository as one Rust crate under `server/`. Do not introduce a separate repository
+or a multi-crate workspace initially. The intended layout is:
+
+```text
+server/
```
Slonegg: "server" sounds very generic.

plans/native-llm-serving.md
```diff
@@ -0,0 +1,351 @@
...
+├── Cargo.toml
+├── Cargo.lock
+├── ARCHITECTURE.md
+├── src/
+│   ├── main.rs              # Startup, configuration, shutdown
+│   ├── api.rs               # HTTP endpoints and streaming
+│   ├── protocol.rs          # Messages exchanged with Python worker
+│   ├── worker.rs            # Worker process, admission, cancellation
+│   └── runtime/             # Added in milestone 2
```
Slonegg: Should we use runtime independently for benchmarking, or only through the server crate?

PR #812, opened 2026-09-16
Refresh CloudRift NVIDIA VM images

Slonegg:
## Summary

- update the default R580 open-driver Ubuntu 24.04 image to the current CloudRift recipe image
- update the proprietary R580 CUDA 12.9 image to the newly promoted NVLink-fixed image
- leave the ROCm image unchanged because it already matches the public recipe catalog

## Validation

- `./venv/bin/pytest -q tests/provisioning/test_cloudrift.py` (84 passed)
- confirmed both image objects return HTTP 200


trifonova-n:
The proprietary bump looks right in direction, but the open-driver image here predates the NVLink fix, and this overlaps with #810.

**The open-driver image still disables NVLink.** `ubuntu-noble-server-gpu-580-129-20260716-134310.img` was built on 16 July 2026. CloudRift removed `NvLinkDisable=1`, `NvSwitchBlacklist=all` and `RmDisableFabricManager=1` from the image build on 8 August 2026 (cloudrift `0029a1d6` "FC-1379 Enable NVLink in GPU VM images"). So this build still ships the compute-only modprobe config that turns NVLink off, the same as the 2025-10-15 image it replaces. Every RTX, A100, H200 and B200 rental would keep booting without NVLink.

That matters because the catalog image is what caused the problem in the first place: on a four-card V100 rented with the April image, `nvidia-smi topo -m` reports `PHB` between every pair, `nvidia-smi topo -p2p r` reports no peer-to-peer, and the driver shows `NvLinkDisable: 1`. Tensor-parallel all-reduce is then staged through host memory, and cold prefill drops to **138 tok/s** (a 4,038-token prompt takes 29.0 s), with the GPUs pegged at 100% utilisation while drawing 50 W of a 300 W budget.

**Suggested replacement:** `ubuntu-noble-server-gpu-580-129-20260810-232754.img`, the same-day twin of the proprietary build and the first open-driver build after the fix.

**The proprietary image you picked (2026-09-16) is newer than the one I measured**, so it is probably fine, but neither of us has booted it. #810 pins `gpup-580-129-20260810-232733`, verified end to end on a rented four-card V100:

| Check | April image | 2026-08-10 image |
| --- | --- | --- |
| Driver / `NvLinkDisable` | 580.126.20 / `1` | 580.173.02 / `0` |
| Topology / peer-to-peer | `PHB` / `NS` | `NV1`,`NV2` at 25.8 GB/s / `OK` |
| Cold 4K prefill | 29.0 s (138 tok/s) | 2.0 s (2,036 tok/s) |
| Cold 6K prefill | 43.3 s (139 tok/s) | 2.1 s (2,789 tok/s) |

**Overlap:** #810 changes these same two constants, adds tests that pin both builds so a revert to an NVLink-disabled image fails, and documents why in `emmy/provisioning/ARCHITECTURE.md`. The two branches conflict. Simplest path is to close this one in favour of #810, or reduce it to the proprietary line and decide which proprietary build we want — happy to boot the 2026-09-16 one and compare before switching to it.

Note on validation: an HTTP 200 on the image URL and the provisioning tests both pass with an NVLink-disabled image, so neither catches this. `nvidia-smi topo -m` on a rented host is what shows it.

Slonegg:
Thanks — good catch on the open-driver image. I rechecked the live `/recipes/list` response: the open R580/CUDA 12.9 catalog entry still points to `ubuntu-noble-server-gpu-580-129-20260716-134310.img`, so matching the catalog is not sufficient here. That build predates the 8 August NVLink fix. The open constant in this PR should use `ubuntu-noble-server-gpu-580-129-20260810-232754.img` instead.

The proprietary September image is verified, not just assumed. After workflow run 35055689221 promoted `ubuntu-noble-server-gpup-580-129-20260916-043401.img`, I rented a fresh 4x V100 SXM2 VM on `cato-ubuntu-66-172-10-183` and waited for cloud-init. The guest reported driver 580.178.04, `NvLinkDisable: 0`, `NV1`/`NV2` topology, active 25.781 GB/s links, `OK` in `nvidia-smi topo -p2p r` for every linked pair, and `cudaDeviceCanAccessPeer=1` for those pairs. Fabric Manager was installed and conditionally inactive because this host has direct NVLink rather than an exposed NVSwitch. The VM was terminated after the test.

One correction on overlap: #810 is merged, but its GitHub file list does not include `emmy/provisioning/cloudrift.py` or `tests/provisioning/test_cloudrift.py`. It documents and uses an explicit `--image-url` override; it did not update Emmy’s default constants. Closing #812 in favor of #810 would therefore leave the stale defaults in place. The appropriate follow-up is to amend this PR to use the 10 August open-driver image, retain the end-to-end-tested 16 September proprietary image, and add exact regression assertions for both pins.

Slonegg:
Clarification after confirming the intended source of truth: this PR will use the latest images selected by the live CloudRift recipe catalog for every image family. Those are currently:

- open NVIDIA CUDA 12.9: `ubuntu-noble-server-gpu-580-129-20260716-134310.img`
- proprietary NVIDIA CUDA 12.9: `ubuntu-noble-server-gpup-580-129-20260916-043401.img`
- AMD ROCm 7.2: `ubuntu-noble-server-rocm-64-20260220-025112.img`

Accordingly, no August bucket-only override will be used. The V100/P100 path selects the September proprietary recipe image, which was verified end to end with NVLink. The reviewer concern about the July open-driver recipe remains a CloudRift catalog issue, but Emmy should mirror the promoted recipe rather than independently select an unpromoted object. This supersedes my previous suggestion to change the open constant to the August object.

PR #814, opened 2026-09-16
Rent CloudRift hosts on VM images that leave NVLink enabled

trifonova-n:
## Abstract

CloudRift's VM images disabled NVLink in the NVIDIA driver until 2026-08-08, and Emmy still rents the builds from before that fix, so every host it provisions comes up without NVLink and without peer-to-peer between GPUs. On a four-card V100 that costs most of the prefill speed: a 4,038-token prompt takes 29.0 s instead of 2.0 s. These two commits were written on the branch of #810 but pushed after it was squashed, so they never reached main; this PR carries them over unchanged. Relay provisions through the `emmy` CLI and pins no image of its own, which is why its deployments picked the old image too.

| Check | Image on main | Image here |
| --- | --- | --- |
| Driver / `NvLinkDisable` | 580.126.20 / `1` | 580.173.02 / `0` |
| Topology / peer-to-peer | `PHB` everywhere / `NS` | `NV1`,`NV2` at 25.8 GB/s / `OK` |
| Cold 4K prefill | 29.0 s (138 tok/s) | 2.0 s (2,036 tok/s) |
| Cold 6K prefill | 43.3 s (139 tok/s) | 2.1 s (2,789 tok/s) |

---

## What changes

- **Proprietary image (V100, P100)** moves from the 2026-04-30 build to `gpup-580-129-20260810-232733`. Verified end to end on a four-card V100 rented through `emmy vm create gpu` with no override: NVLink links train at 25.8 GB/s, peer-to-peer works on every linked pair, and the same recipe prefills 15-20x faster. During the slow runs the GPUs read 100% busy while drawing 50 W of a 300 W budget — small kernels waiting on host-staged all-reduce, not real work.
- **Open-driver image (RTX, A100, H200, B200)** moves from the 2025-10-15 build to `gpu-580-129-20260810-232754`, the same-day twin built from the same image configuration. Boot-checked on a rented RTX 4090: the driver loads, CUDA sees the card, `NvLinkDisable` is 0, and the image keeps none of `NvLinkDisable`, `NvSwitchBlacklist`, `RmDisableFabricManager`. NVLink itself is unverified there — no A100 or H200 capacity was rentable, and RTX parts have no links. The code comment and `emmy/provisioning/ARCHITECTURE.md` say so.
- **ROCm image** unchanged: it already pins the newest build in the bucket.
- **Tests pin both builds**, so a revert to an image from before the fix fails rather than silently costing speed. An HTTP 200 on the image URL and the existing provisioning tests both pass on an NVLink-disabled image, which is why the pin checks the build.

## Deploying this

Merging is not enough for Relay. It runs the installed `emmy` binary (`emmy deploy cloud`, `emmy vm delete`), so its host needs the new version, and existing deployments keep the image they booted with until recreated.

## Overlap with #812

That PR changes the same two constants. Its proprietary pick (2026-09-16) is newer than the build verified here and may well be fine, but neither has been booted; its open-driver pick (2026-07-16) predates the 2026-08-08 fix and still disables NVLink. Happy to test the newer proprietary build and follow up if we prefer it.

## Verification

`pytest tests/provisioning` 202 passed. `ruff check` and `ruff format --check` clean. Measurements above come from rented hosts, one per image, with the raw commands in the commit messages.


trifonova-n:
Superseded by #812, which landed the proprietary image with its own end-to-end NVLink verification. Closing to keep one source of truth.

The open-driver constant is the one thing this carried that #812 does not: it still points at `ubuntu-noble-server-gpu-580-129-20260716-134310.img`, built before the 2026-08-08 NVLink fix, so RTX, A100, H200 and B200 rentals keep booting with NVLink disabled. That now rests on CloudRift promoting the August build (or newer) into the recipe catalog, per the policy agreed in #812 that Emmy mirrors promoted images.

PR #817, opened 2026-09-16
Find, bound, and stop poisoning the fused attention target the kernel corpus blocks on

Daksh-10:
## Abstract

Three defects that made the Qwen3-0.6B layer-0 kernel corpus unusable on RTX 4090 and 5090 in the verify-3 cycle, all of them converging on one triple-fused attention target and the two targets queued behind it. The golden selector could not find that target at all, because the kernel the compiler builds for it carries one origin more than the recording. The tuner could not bound a search over it, because a terminal that fails to lower spends no budget. And the splitter kept minting the malformed pieces that made it fail to lower in the first place.

Verified against the verify-3 working files on both cards, base and FP8: 54 of 56 targets resolved before, 56 of 56 after, and the two recovered are the blocking target on each card.

---

## The budget accounting

The terminal bench already knew which of three things had happened to a terminal. It only had a bool to say it with, and that bool was read as both "spent GPU time" and "was fresh work". Those are not the same question, and the un-lowered terminal is the case where they disagree: it is fresh, and it measured nothing.

Replacing `measured: bool` with an `origin` of `live`, `replay` or `dead_end` lets the origin alone decide what is spent, and drops a flag that carried three meanings rather than adding a second one beside it.

| origin | what happened | measurement budget | patience |
| --- | --- | --- | --- |
| `live` | the backend ran, `ok` or failed | spent | spent |
| `replay` | cache hit, stub backend, or no `CudaOp` | — | — (a better cached result still resets it) |
| `dead_end` | finished holding an un-lowered kernel node | — | spent |

The verify-3 4090 and 5090 legs logged that warning 26 times per leg, beside searches that ran two hours and produced no admissible row.

## The split closure gate

Cherry-picked from `exp/verify-3` (`834ac4c7`) and rebased: `add_output_piece` lost its underscore on main, and the pricing helpers gained a `deadline` argument.

`030_cut` enforces closure on the seams it offers. Nothing re-checked the region `realize_split` slices for a placed piece, so a slice that lost a `Load` the rebound region still reads minted a kernel that every later body computation refused. `_unclosed_piece` applies the same rule at all three mint points.

Pricing then has to treat that refusal as data, because a deferred structural leaf is a thunk and expanding it runs the rewrite. `_materialized_graph` drops a refused splice from the priced set instead of letting the raise escape a fork that had a healthy alternative in hand. Both structural pricing sites go through it now, so `_leaf_graph` has no other caller.

This closes the greedy path only. The MCTS offer-time refusal for the same class is still open and needs a card to reproduce — the one attempt to find its raise site without one failed.

## The target selector

A frontend target names its kernel by the origins the trace gave it, and the selector demanded the lowered kernel carry exactly that set. Two contexts are tried: the whole persisted program, then the target's own frontend slice.

Exact equality is right in the program, where a node carrying more origins is a genuinely larger fused kernel and selecting it would measure something the record does not describe. It is too strict in the slice: the slice *is* the recorded cone, so every origin a node there carries came from inside it, and a surplus is what lowering now materializes for the cone rather than work fused in from outside. The blocking target misses by exactly one such origin — a broadcast of a normalization weight the recording predates — on every card.

The slice now also accepts the smallest kernel that covers the recording, and stays silent when two cover it equally. An ambiguous target is still a target that no longer resolves.

Worth recording because it sent the last investigation the wrong way: the verify-3 notes diagnosed this as the recording naming an origin the compiler no longer produces. It is the opposite. The recording is a strict subset, which is why the slice fallback #797 added rescues every other target and not this one.

## Documentation

Two claims in `pipeline/ARCHITECTURE.md` no longer matched the code and are corrected: patience and failed lowering, and `--max-candidates` being a per-kernel budget, which is true of the inner search only — the outer structural search is constructed with no measurement cap at all.

## Line balance

`git diff --stat origin/main -- emmy/` is +97 / -38, net +59 lines of code. One reduction the change exposed is taken: both structural pricing sites now go through `_materialized_graph`, so `_leaf_graph` had a single caller and existed only to be wrapped in a `try` — inlined.

The rest is not shaved, because it is not layering. Roughly half is comment explaining why each rule is where it is, and the executable half is behaviour that did not exist: a closure check at the three mint points, patience on an outcome that spent nothing, and a selector rule that resolves targets the old one could not. The `origin` change is a concept removed rather than added — one three-valued value replacing one boolean that three different questions were being asked of.

## Testing

`tests/compiler/pipeline/search/policy/`, `tests/compiler/pipeline/search/test_golden.py`, `tests/compiler/passes/test_cut_forks.py`, `test_split_fresh_kernels.py`, `test_schedule_walk.py`. The existing parametrized stopping test already covered evaluation origin; it gains three `dead_end` cases, including one proving a level of nothing but dead ends stops on patience rather than running to the visit limit.

No checked-in program reproduces the selector's surplus origin — six broadcast shapes were tried and the compiler folds every one of them at that size — so that test stubs which graph the slice hands back and uses a real lowered one, pinning both halves of the rule. The real evidence is the verify-3 replay above, which is a check that was run rather than one that can be committed.

`make test`: **1 failed, 4807 passed, 1022 skipped, 9 xfailed**. The one failure is `test_the_checked_in_freeze_is_the_default_evaluation_corpus`, which fails on this machine because the measurement freeze payload is a git-LFS pointer — the checkout has `skip-smudge` set deliberately. The test's own error message names that condition. It is unrelated to this diff. `make lint` clean.

`make test-goldens` goes from **6 failing model goldens to 4**. Two Qwen3.8-27B V100 files turn green, and that was checked rather than assumed: on `origin/main` both fail with *the persisted target selects no kernel after lowering* — one row in the AWQ file, two in the GPTQ one, all on the same `k_linear_pointwise` target. The rows were never wrong; the target could not be found. Making it findable is a fix, not a relaxed check.

The remaining four failures are pre-existing and untouched by this PR: `gemma-4-12B-it` on sm_89 and sm_120, `Qwen3.8-27B-EXL3` and `DeepSeek-V4-Flash-0731` on sm_70. They fail with *every key and value is offered, no one candidate carries them together* — an enumeration change, not a selector one. The two V100 files among them were recorded by #797 on 2026-09-14 and already did not decode two days later, which suggests #804. Worth its own issue; deliberately not re-recorded here.

Neither compiler fix is verified on a GPU. Both problems were found on hardware and the one that remains — the MCTS offer-time refusal — cannot be closed without a card; the re-tune that would do it is the next step, not this PR.

---

## RTX 4090 evidence (added 2026-09-18)

The three fixes above were taken to a driver-verified RTX 4090 and the layer-0 corpus re-tuned at deployable `-O3`. Seventeen targets recorded with paired eager / Inductor / Emmy timings, committed as working-level goldens beside the existing A100 and H100 files. Full write-up in `experiments/golden-bench-2026/kernels/RESULTS.md`.

Emmy beats Inductor on every recorded target except the base-checkpoint transpose (0.88), which also loses to eager — consistent with the previous cycle.

### What the hardware said about the three fixes

**The selector fix is confirmed useful.** It is what made the deferred fused targets resolvable at all, and what let the corrected recording pass attribute measurements to the right realization.

**The budget fix fires, at the wrong altitude.** Its case was observed eight times on the card (`N node(s) left un-lowered — bench_fail without benching`), each now spending patience where it previously spent nothing. But the fused target still ran to an external 30-minute cap rather than stopping on its own: patience bounds one level, and the two-level strategy enrolls each minted kernel as a fresh level. A per-target wall clock is what that case needs; a counter cannot bound work whose unit cost is a 60-second watchdog. Not addressed here.

**The split closure fix cannot be attributed.** Zero candidates were dropped this run — the `KeyError` class that discarded 2,603 candidates in the previous cycle did not fire once. But main's placement-cut fix (#827) and the identity re-key (#804) landed in the same window, so no claim is made about which change killed it.

### What the hardware said that nobody expected

The base checkpoint has lost most of its corpus to maximal fusion — four distinct targets per sequence length against the nine the A100 and H100 sections report. The fused attention kernels that absorbed them cannot be tuned: they build cleanly and then hang, 24 watchdog kills in thirty minutes, with two variants completing at 8.81 s and 7.14 s per launch. The search failure recorded in the previous cycle was a symptom; the cause is a schedule space with no usable answer for that shape. That is the open performance work, and it is untouched by this PR.

The name-based deferral heuristic everyone uses — match on both `sdpa` and `reduce` — is also wrong, broken from both sides by three further targets.

### Not done

No 5090 leg; none was supplied. The recipe's 4090 rows are left on the search lane rather than switched to golden replay, because the corpus no longer matches what a replay would need to claim.


Slonegg (approved):

Slonegg:
## The fused attention target is reachable with placement cuts. Choosing that route is what fails.

The description says no knob can express the hand-written study's three fixes. The compiler already offers them as one pair of placement cuts. Nothing selects that route. The bigger fused kernels on the RTX 4090 also hit real compiler gaps. Everything below is compile-only on the rebased branch (after #864). None of it has been timed on a GPU.

### The study's target: two cuts

`k_sdpa_mean_reduce_29d3df` (A100, sequence 512: q/k RMSNorm, RoPE and SDPA with the online softmax) offers 8 cut seams. Two of them are the operand cones of the score contraction, the normalized and rotated Q and K. Pinning both gives three kernels: a Q norm+RoPE kernel, a K norm+RoPE kernel, and an attention kernel whose default schedule uses `mma`, `ldmatrix`, `cp.async` and shared memory. That covers all three rungs of the study: hoisting, tiling with online row statistics, and tensor cores. Without the pins, greedy selection keeps one scalar fused kernel.

```
EMMY_KNOBS='PLACE@map.1/twist.1/inner.1/map=cut,PLACE@map.1/twist.1/inner.2/map=cut' \
  emmy compile --golden experiments/golden-bench-2026/kernels/golden/qwen3-06b-s512_a100.golden.yaml \
  --realization k_sdpa_mean_reduce_29d3df --target sm_80 --dump-dir DIR
```

The study compared against only one cut, the one that writes out the score matrix (0.04x and 0.06x on A100). The operand cuts are different ones.

### Why nothing finds it

Greedy selection prices the structural fork with the prior, which keeps the kernel fused. The tuner spends its budget timing schedules of the fused kernel, which take seconds per launch or hit the watchdog, so it never reaches the cut route. So this is a search problem, not a missing transformation. In the paper's terms the budgeted-search assumption fails here, not the cut or enumeration assumptions.

### Real gaps: the two RTX 4090 fused kernels

On the base checkpoint, maximal fusion puts the projections, SDPA and MLP into `k_linear_sdpa_mean_reduce` (14 seams) and `k_sdpa_linear_mean_reduce` (19 seams). Any attention or linear point for that card has to come from cutting these apart. Trying routes on the post-attention kernel turned up:

- **Duplicate work.** Attention and o_proj each appear twice as separate seams, because fusion inlines a shared value at each use. Every route I tried computes both of them twice.
- **A crash.** A six-cut route (both attention occurrences, both o_proj occurrences, the normalized MLP input, the MLP activation) fails in `100_loopify`: `Lambda body reads ['_r0'] it does not bind`. With `LOOPIFY=0` it compiles, into 7 kernels.
- **No tensor cores for attention there.** The attention pieces of that kernel get no `mma` on their default schedule, unlike the A100 target. I did not check whether the tier is offered at all.
- **Missing schedule options.** Converting the goldens to Loop IR showed that a kernel stored on its own drops options that it has inside the program: three RTX 4090 FP8 rows no longer decode. That is in the conversion commit's message.

### Suggested next steps

1. Bench the two-cut route on a rented A100 or RTX 4090 with `emmy run --golden ... --realization k_sdpa_mean_reduce_29d3df --bench` under the pins above. It takes about a minute, and it decides whether the study's numbers are reachable today.
2. If it is fast, record it as a route row so the deploy compile picks it over the fused kernel. The FP8 decode mean in `RESULTS.md` shows a structural winner needs this.
3. Fix the loopify crash on multi-cut routes.
4. Make a value shared by several uses one seam in the post-attention kernel, so a route computes attention and o_proj once. Then check whether its attention pieces reach a tensor-core tier.
5. Correct the "not expressible as a knob" claim in this PR's description before it lands.

PR #859, opened 2026-09-19
Add native cached generation for dense Qwen3

6erun:
## Abstract

Prepared dense FP16 Qwen3 artifacts can now run cached, single-request generation in Rust. Keys, values, and the next token stay on the GPU, and CUDA graph capture advances state exactly once. This adds the generation foundation for a later native server; vLLM remains the serving default.

---

## Execution

Python prepares the existing compiled pre/post-attention programs and output head. Packaged CUDA source supplies embedding, rotary embedding, contiguous cache writes, attention, and greedy sampling. The existing supervised worker owns the complete generation loop and retires on failure. The existing exporter, plans, compiler cache, and supervisor are reused.

Prefill is sequential and sampling is greedy. Context capacity is limited to 4,096. Weight copies and per-layer scratch remain unoptimized. HTTP serving, concurrency, Python-dispatch replacement, and performance qualification remain separate work.

## Numerical qualification

Explicit FP16 rounding repairs a fused rotary multiply/add that changed a checkpoint token; an independent regression protects the operation boundaries.

Qualification compares against eager FP16 and FP32 using the same FP16-rounded weights and token prefixes. Native error must stay within 2% relative L2 and 0.02 probability total variation at every position, with per-prompt RMS error at most twice the FP16 control. Native argmax must match one reference; when references agree, it must match both. These are experimental acceptance budgets, not guarantees for arbitrary prompts.

The original pointwise tolerance failed, including for Hugging Face's own cached/full-prefix comparison. A first comparative trial also failed six of eight cases because it compared individual positions against near-zero reference error and required the control to pass native's budget. The revised RMS comparison was committed before four fresh prompts ran. The failed trials remain recorded; absolute limits are unchanged.

All four fresh cases pass (265 positions). Across calibration and held-out runs, all 819 argmax choices match both references; maximum native relative L2 error is 1.28% and total variation is 0.0114. Complete-checkpoint coverage reaches 256 positions. An independent attention/cache test reaches 4,096; this does not qualify a complete checkpoint at that length. Tiny-model logits retain the strict `1e-3` gate, and the identical Python/Rust artifact must produce bit-identical logits.

The [report and sanitized evidence](https://github.com/cloudrift-ai/emmy/blob/codex/native-cached-generation/experiments/Qwen3-0.6B/native_generation/RESULTS.md) record the protocols, failed trials, layer isolation, and coverage limits. No compiler arithmetic or schedule changes are included.

## Validation

The full suite reports 5,191 passed, 748 skipped, 21 xfailed, and 27 failed. Twenty-six failures reproduce unchanged on main (`22a5253b3`): TMA hardware requirements, a block-scaled kernel expectation, and existing serving strict-evidence gaps. The other test timed out during overlapping GPU checks and passes alone on both branches. All native-generation tests pass. Python lint, Rustfmt, Clippy, and seven Rust unit tests pass. After merging main at `7e6024f4e`, rerunning the 27 previously failing tests gives 26 failures and one pass: the same baseline failures remain, and the benchmark-worker check passes. The merge changes DeepSeek documentation and goldens, not native-generation code. The GPU full-suite gate remains non-green; the PR has been marked ready for review. The complete generation client reproduced all 15 checked tokens from the diagnostic step path with CUDA graphs and its default deadline. The CUDA source is packaged in both wheel and sdist, with byte-identical resource loading verified outside the checkout.


Slonegg (approved):
Very cool!

